PR #16, opened 2022-04-21
12 add mock and tests barebones for spacewalk pallet

### LINKED ISSUE #12: Add mock and tests barebones for spacewalk pallet
We need to configure `mock.rs` and `tests.rs` with the pallets that we want to include in our mocked runtime, in order to start writing tests for the pallet.

ashneverdawn:
Closes #12


pallets/spacewalk/Cargo.toml
```diff
@@ -41,6 +41,25 @@ orml-utilities = {default-features = false, git = "https://github.com/open-web3-
 pallet-transaction-payment = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.17"}
 substrate-stellar-sdk = {git = "https://github.com/pendulum-chain/substrate-stellar-sdk", branch = "polkadot-v0.9.17", default-features = false, features = ['offchain']}
 
+# for mock runtime
```
gonzamontiel, line 44: Are all of these pallets required for mocking a simple runtime?

pallets/spacewalk/src/tests.rs
```diff
@@ -1,20 +1,74 @@
...
+const VAULT: [u8; 32] = [
+	102, 78, 11, 2, 53, 135, 217, 14, 19, 204, 166, 232, 115, 151, 132, 5, 222, 65, 16, 180, 11,
+	100, 184, 124, 145, 235, 247, 247, 92, 133, 186, 199,
+];
+const USDC: [u8; 4] = [b'U', b'S', b'D', b'C'];
+const AMOUNT: u128 = 1_000_000_000_000;
 
 #[test]
-fn it_works_for_default_value() {
+fn report_stellar_transaction_mints_asset() {
```
gonzamontiel, line 31: cool that you already added some tests! it wasn't required but thanks!

gonzamontiel (approved):
good to me! Still looks like a pretty big and complex mock. I'm specially concerned that we added quite a few dependencies to the pallet just for mocking the runtime, but I guess is this the way everyone does it? If that's the case I'd consider this approved so we can move on and maybe iterate over this to make it more efficient.

pallets/spacewalk/Cargo.toml
ashneverdawn: Yeah I needed to include them just to get the thing to compile. I based it off the testchain. I created tickets to go through and minimize both the testchain and the mock runtime and remove as many features as possible. I'm sure it could be simplified it to be at least 10x smaller and remove most dependencies. Especially since spacewalk is just one pallet, not all of pendulum.

PR #27, opened 2022-04-27
add ci actions for pull requests

xiuxiu62:
Clippy checks are set to produce warnings, but are recommended

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,49 @@
+on: [pull_request]
+
+name: continuous-integration
+
+jobs:
+  ci:
+    strategy:
+      fail-test: false
+      matrix:
+        platform: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


xiuxiu62:
resolves [#20](https://github.com/pendulum-chain/spacewalk/issues/20)

xiuxiu62:
just merged the same workflow into pendulum-launch -- i'll test against that and let you know if config is sound

xiuxiu62:
just tested this against my fork with errors.  I'll work on the config and let you know when I have something functional.  Let me know if you have suggested changes.  Clippy flags are set to warnings, which I find quite helpful (for example).  here's the original repo I grabbed this [from:[](](https://github.com/nushell/nushell/blob/main/.github/workflows/ci.yml) https://github.com/nushell/nushell/blob/main/.github/workflows/ci.yml)

PR #35, opened 2022-05-03
remove unnecessary orml dependencies

### LINKED ISSUE #9: Make Spacewalk independent of orml for Currency specification
Ideally, we would want that Spacewalk is independent of how users of the pallet want to implement multiple currencies in their chains.
That could be achieved by implementing our own set of traits as a layer of abstraction of those mechanisms.

As an effort in that direction, we want to minimize the dependency with orml pallets in general. So far, the spacewalk pallet is depending on `orml_tokens`, `orml_currencies` and `orml_traits`.

We want to remove the `orml_tokens` and `orml_currencies` dependencies, moving part of the related code to the `testchain`, and only depend on `orml_traits`, which is a fair compromise for the time being.

EDIT: even though `orml_traits` provides an interface for MultiCurrency that we wanna use, we don't need to depend on orml only for that spec. We need to come up with the minimum trait specification for MultiCurrency that our pallet needs and implement it. It may turn out that is really similar to https://github.com/open-web3-stack/open-runtime-module-library/blob/d42fdd6b64fbe4d234accd5d6bd862f27b97e767/traits/src/currency.rs#L18.

ashneverdawn:
Closes #9


ashneverdawn:
See #9 about why `orml_traits` is still needed.

gonzamontiel (approved):
As per discussion in #9, I think it is enough for now to depend only on `orml_traits`.

PR #61, opened 2022-05-06
Dev/38 resolve parity util mem compile error

ashneverdawn:


ebma:
What exactly fixed this error? I noticed that the vault compilation fails for the current main branch now so we would need it on the main branch as well.

gonzamontiel:
HJey @ebma , It was bumping the polkadot version to v0.9.18, and it was actually also merged to the main branch in https://github.com/pendulum-chain/spacewalk/pull/60

Could be _that_ the cause of the problem in the main branch?

ebma:
Yes, it seems like #60 introduced this error on the main branch. But I thought this PR contains a fix for it? I'm a bit confused

gonzamontiel:
@ebma can you create an issue on main describing the compilation errors?

ebma:
I now created an issue ticket (#69) for this.

PR #67, opened 2022-05-16
65 create stellar crate and move conversions logic

b-yap:
I encountered this error again:
```
error: the feature `const_fn_trait_bound` has been stable since 1.61.0 and no longer requires an attribute to enable
 --> crates/currency/src/lib.rs:6:12
  |
6 | #![feature(const_fn_trait_bound)]
  |            ^^^^^^^^^^^^^^^^^^^^
  |
note: the lint level is defined here
 --> crates/currency/src/lib.rs:3:9
  |
3 | #![deny(warnings)]
  |         ^^^^^^^^
  = note: `#[deny(stable_features)]` implied by `#[deny(warnings)]`

error: could not compile `currency` due to previous error
```

Now if I remove `#![feature(const_fn_trait_bound)]`, a new error will appear:
```
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
  --> crates/currency/src/amount.rs:26:6
   |
26 | impl<T: Config> Amount<T> {
   |      ^
27 |     pub const fn new(amount: BalanceOf<T>, currency_id: CurrencyId<T>) -> Self {
   |     -------------------------------------------------------------------------- function declared as const here
   |
   = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
   = help: add `#![feature(const_fn_trait_bound)]` to the crate attributes to enable
```

~So to fix all of this I upgrade to version `1.62.0-nightly`.~ reverted it back


crates/spacewalk/Cargo.toml
```diff
@@ -36,7 +36,7 @@ sp-std = {git = "https://github.com/paritytech/substrate", default-features = fa
 orml-traits = {git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "2b5d4ce1d08fb54c0007c2055653892d2c93a92e", default-features = false}
 
 pallet-transaction-payment = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.18"}
-substrate-stellar-sdk = {git = "https://github.com/pendulum-chain/substrate-stellar-sdk", branch = "polkadot-v0.9.18", default-features = false, features = ['offchain']}
+stellar-support = { path = "../stellar", default-features = false}
```
gonzamontiel, line 39: what is `stellar-support` ?

gonzamontiel: do we need to change the name for some reason?

gonzamontiel (changes requested):
Just added a small comment that I think we need to change. The rest is fine!

crates/stellar/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub mod address_conv;
+pub mod balance_conv;
+pub mod currency;
+pub mod currency_conv;
+pub mod horizon;
+
+pub use substrate_stellar_sdk as substrate_sdk;
```
b-yap, line 9: @gonzamontiel  here I made it public, so that other crates which uses the stellar pallet, won't need to add
`substrate_stellar_sdk` as a dependency in their `cargo.toml`.

crates/spacewalk/Cargo.toml
```diff
@@ -60,7 +61,7 @@ std = [
   "sp-std/std",
   'orml-traits/std',
   'base64/std',
-  'substrate-stellar-sdk/std',
+  'stellar-support/std',
```
b-yap, line 64: there are 2 `substrate-stellar-sdk/std` defined. The first one is found at like 53.

gonzamontiel (approved):
To merge!

PR #72, opened 2022-06-15
Fix: proper extrinsic format and change sr25519 to ed25519

TorstenStueber:
This fixes problems with the Spacewalk integration in Pendulum.

### DIFF
--- a/clients/runtime/src/cli.rs
+++ b/clients/runtime/src/cli.rs
@@ -1,17 +1,18 @@
 use crate::{
     error::{Error, KeyLoadingError},
-    SpacewalkParachain, SpacewalkSigner, rpc::ShutdownSender,
+    rpc::ShutdownSender,
+    SpacewalkParachain, SpacewalkSigner,
 };
 use clap::Parser;
-use sp_keyring::AccountKeyring;
-use std::{collections::HashMap, num::ParseIntError, str::FromStr, time::Duration};
-use subxt::sp_core::{sr25519::Pair, Pair as _};
+use sp_keyring::Ed25519Keyring;
+use std::{collections::HashMap, num::ParseIntError, time::Duration};
+use subxt::sp_core::{ed25519::Pair, Pair as _};
 
 #[derive(Parser, Debug, Clone)]
 pub struct ProviderUserOpts {
     /// Keyring to use, mutually exclusive with keyfile.
     #[clap(long, required_unless_present = "keyfile", parse(try_from_str = parse_account_keyring))]
-    pub keyring: Option<AccountKeyring>,
+    pub keyring: Option<Ed25519Keyring>,
 
     /// Path to the json file containing key pairs in a map.
     /// Valid content of this file is e.g.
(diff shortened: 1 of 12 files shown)


clients/runtime/src/cli.rs
```diff
@@ -57,8 +58,18 @@ fn get_credentials_from_file(file_path: &str, keyname: &str) -> Result<Pair, Key
     Ok(pair)
 }
 
-pub fn parse_account_keyring(src: &str) -> Result<AccountKeyring, Error> {
-    AccountKeyring::from_str(src).map_err(|_| Error::KeyringAccountParsingError)
+pub fn parse_account_keyring(src: &str) -> Result<Ed25519Keyring, Error> {
```
TorstenStueber, line 61: Unfortunately, `Ed5519Keyring` does not have a `FromStr` implementation (but `Sr25519Keyring` does – strange). I just copied a version here.

clients/runtime/src/lib.rs
```diff
@@ -79,7 +77,7 @@ impl Config for SpacewalkRuntime {
     type Hash = H256;
     type Hashing = BlakeTwo256;
     type AccountId = AccountId;
-    type Address = Self::AccountId;
+    type Address = Address;
```
TorstenStueber, line 80: This is the elementary fix that ensures that subxt uses the actual address format used in the (testchain) runtime and not the account id format.

clients/runtime/src/rpc.rs
```diff
@@ -1,425 +1,410 @@
...
-		const DEFAULT_SPEC_VERSION: RangeInclusive<u32> = 1..=1;
-		const DEFAULT_SPEC_NAME: &str = "standalone";
-	} else if #[cfg(feature = "parachain-metadata")] {
-		const DEFAULT_SPEC_VERSION: RangeInclusive<u32> = 3..=3;
-		const DEFAULT_SPEC_NAME: &str = "parachain";
-		pub const SS58_PREFIX: u16 = 2032;
-	}
+    if #[cfg(feature = "standalone-metadata")] {
+        const DEFAULT_SPEC_VERSION: RangeInclusive<u32> = 1..=1;
+        const DEFAULT_SPEC_NAME: &str = "pendulum-parachain";
```
TorstenStueber, line 22: Maybe we need to revert this change in order to make the vault usable with the testchain. In this case we should document how to overwrite the spec name on the command line.

clients/runtime/src/rpc.rs
```diff
@@ -1,425 +1,410 @@
...
+    ) -> Result<(), Error> {
+        self.with_unique_signer(|signer| async move {
+            self.api
+                .tx()
+                .spacewalk() // assume that spacewalk pallet is registered in connected chain
+                .redeem(
+                    asset_code.to_vec(),
+                    asset_issuer.to_vec(),
+                    amount,
+                    stellar_vault_pubkey.clone().to_vec(),
```
TorstenStueber, line 402: We realized in our bug fixing session yesterday that this was missing.

clients/runtime/src/rpc.rs
```diff
@@ -1,425 +1,410 @@
...
+        const DEFAULT_SPEC_VERSION: RangeInclusive<u32> = 3..=3;
+        const DEFAULT_SPEC_NAME: &str = "parachain";
+        pub const SS58_PREFIX: u16 = 2032;
+    }
 }
 const TRANSACTION_TIMEOUT: Duration = Duration::from_secs(300); // 5 minutes
 
-type RuntimeApi =
-	metadata::RuntimeApi<SpacewalkRuntime, SubstrateExtrinsicParams<SpacewalkRuntime>>;
+type RuntimeApi = metadata::RuntimeApi<SpacewalkRuntime, PolkadotExtrinsicParams<SpacewalkRuntime>>;
```
TorstenStueber, line 31: Changing this from `SubstrateExtrinsicParam` to `PolkadotExtrinsicParam` will ensure that subxt uses an extrinsic format where tips are only value and not pairs of values and asset code.

Quite surprising considering these names.

gonzamontiel: wow, curious and weird, nice catch

gonzamontiel (commented):
I left one small suggestion, but fine for me otherwise!

ashneverdawn (approved):
thanks for your comments, really helpful. 
lgtm 👍

PR #73, opened 2022-06-15
Change parachain metadata feature

ebma:
This PR changes the cargo features. Now `standalone-metadata` and `parachain-metadata` are available. Chosing one of these two for running and building the vault will have an impact on the expected spec name and metadata of the chain the vault is connecting to. Having these two features enables us to use the vault with both the testchain and pendulum by only changing the supplied feature.

I didn't test if this works yet. I added the `parachain-metadata.scale` that works for pendulum but the `standalone-metadata.scale` is still a bit out of date (the signature of the `redeem()` function of the spacewalk pallet changed but that change is not yet included in the scale file) so this would also have to be updated.

Based on #72.


gonzamontiel (approved):
Looks good to me, but didn't test it neither. @ashneverdawn could you test this?

TorstenStueber:
It is rather annoying that GitHub does not bother to close this [6 year old issue](https://github.com/isaacs/github/issues/750). The diff in this ticket still show the changes that we are already merged into the main branch with a [previous PR](https://github.com/pendulum-chain/spacewalk/pull/72).

PR #74, opened 2022-06-23
Make clients independent of test chain

### LINKED ISSUE #71: Make clients independent of test chain
Currently the clients refer to the testchain (they have a dependency to `spacewalk_runtime`). This is meant to use the definition of `AccountId`, `Address` and `Balance` of the testchain. However, this makes the client only usable for Substrate chains that use the same configuration parameters as the testchain – which is not true.

Another parameter that is hardcoded is what kind of extra information the pallet-transaction-payment requires in the extrinsic of the chain. This information is usually just the tip but the type of the tip also differs between chains: it can be a tip in the native currency and will just require the tip amount or it can be a tip expressed as a multicurrency that requires the tip amount and tip asset.

We should understand what typical configuration parameters are (research Polkadot and Kusama parachains) and come up with a list of settings for each of the parameters. I guess that this is
* `AccountId`: seems to be `sp_core::crypto::AccountId32` for most chains
* `Address`: seems to be either `AccountId` or `sp_runtime::MultiAddress<AccountId, ()>`
* `Balance`: seems to be `u128` for most chains
* tip format: either `PlainTip` (native asset) or `AssetTip` (specifying an asset)

Once we know what kind of parameters we need, we should find a way for users of the vault to specify the proper settings for their own chain. This should most likely happen at compile time. Maybe it can happen through cargo features.

### Tasks
  
* [ ] research typical settings for the 4 parameters
* [ ] find a way that allows to specify these parameters at compile time without requiring to touch the code
* [ ] implement
* [ ] add to README

ebma:
Removes the usages of `spacewalk-runtime` in the vault client and adds a new `multi-address` cargo feature which can be supplied (and should be used when compiling spacewalk for pendulum). 

Although this might still not cover all possible type configurations (since there are only two options for now), I would opt for keeping it simple for now and extending it only once we need to. The different types for account Id and address are based on what interbtc has configured for their chain, and what we need for pendulum.

Closes #71.

### DIFF
--- a/clients/Cargo.lock
+++ b/clients/Cargo.lock
@@ -5406,7 +5406,6 @@ dependencies = [
  "sp-runtime",
  "sp-std",
  "sp-version",
- "spacewalk-runtime-standalone",
  "spacewalk-standalone",
  "substrate-stellar-sdk",
  "subxt",
--- a/clients/README.md
+++ b/clients/README.md
@@ -24,6 +24,14 @@ To run the vault with a parachain (e.g. Pendulum) you need to specify the URL, s
 cargo run --bin vault --features parachain-metadata -- --keyring alice --spacewalk-parachain-url ws://localhost:8844 --stellar-vault-secret-key SB6WHKIU2HGVBRNKNOEOQUY4GFC4ZLG5XPGWLEAHTIZXBXXYACC76VSQ
 ```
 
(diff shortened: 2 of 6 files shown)


ashneverdawn (approved):
lgtm. 👍

PR #97, opened 2022-09-23
94 create security pallet

gonzamontiel:
This is a copy of [this pallet](https://github.com/interlay/interbtc/tree/master/crates/security/src), compilation and tests work properly. The pallet is by default included in the Spacewalk pallet under `dependencies`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4027,6 +4027,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "pallet-security"
+version = "1.2.0"
+dependencies = [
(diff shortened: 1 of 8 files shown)


ebma (approved):
LGTM. Although we might want to replace mentions of 'BTC' in some code comments with spacewalk.

gonzamontiel:
> LGTM. Although we might want to replace mentions of 'BTC' in some code comments with spacewalk.

quickly performing a search I see there are other references, maybe we should perform a cleanup of these wordings all around

Issue #95, opened 2022-09-20
Implement smart pruning mechanism for stored SCP messages in oracle agent


ebma:
We should have an intelligent mechanism to prune the SCP messages from the oracle agent. This way, we save storage space and increase the speed of lookups.
This is not a trivial task, because it is not easy to identify whether SCP messages can already be dropped.

Some metrics we can use to identify entries that can be dropped:
- If all transactions in the transaction set belonging to the SCP messages of a certain slot do not have any transaction memo, (or it is not of the MEMO_HASH type with the correct length), we can drop the SCP messages for that slot since these transactions cannot be related to spacewalk (or if they are, they are faulty). 
- ~Assuming the oracle agent has the list of open issue/redeem requests, it can check whether the ID of one of these requests is included in any of the transactions in a transaction set. If this is not the case, the SCP messages for that transaction set can be dropped. **BUT** it could happen that due to high latency, a transaction is first externalized on Stellar, whereas the issue/redeem request was not created yet. So here we would need some kind of delayed check to prevent this race condition.~ This is too risky on race conditions/bad timing probably.
- SCP messages for transactions that are 'older' than the issue/redeem periods of the respective pallets allow can also be dropped. Assuming that the parachain status is 'running' and is not stalled due to an error

vadaynujra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/95?planning-poker) with Zenhub @b-yap @ebma @TorstenStueber

TorstenStueber:
@ebma is this ticket urgent or merely a nice to have right now? Do we already see too much memory usage or slow down?

Isn't there a definite way that any SCP message from a slot a certain number away from the current slot does definitely have no use anymore (a few days probably)? This your last point, correct?

In that case I think that this would be optimization enough and there would always just be fairly a limited number of SCP messages in memory/under consideration.

ebma:
It's really not urgent. As you can see I originally created the ticket half a year ago already, and at the time it was relevant because it was obvious that we could strip that down a little.
But now that we solved #309 it's probably not really necessary to have this smart pruning, I assume the memory we save doing so is not that significant.

TorstenStueber:
Okay, maybe we close this ticket for now? We can record it at another place (e.g., Future Proposals) and reopen any time when necessary. This way we keep the backlog clean.

[...]

Issue #86, opened 2022-09-01
Listen to and execute issue events in vault client


ebma:
Requires #85. 

Make the vault client listen to events related to issue requests, emitted by the connected parachain and execute them if they are directed at them.

### Handling Issue events
- On a new `RequestIssueEvent` check if it targets this vault and if it does, store it as a new `Issue` request locally in a set
- On a new `CancelIssueEvent` or `ExecuteIssueEvent` remove the related `Issue` if it is contained in the local set
- In a separate task, sequentially try to execute all the issues contained in the issue set (described in more detail below)

### Storing Stellar transactions directed at the vault
- When issuing, the user has to send a Stellar transaction with a payment to the vault
- The vault continuously fetches the transactions for its Stellar account from the network and stores them locally
    - It would be okay to only store the new transactions after going online
    - It would also be okay to discard all transactions that don't contain a transaction memo, because the memo is required for fulfilling issue requests

### Executing Issue requests
- First, look through the stored (vault-related) Stellar transactions and try to find one that contains the `issue_id` as a transaction memo
    - If no such transaction can be found, skip the issue for now (or remove it from the set? to be defined) 
- Check if the `amount` and `asset` contained in the `Issue` match `amount` and `asset` contained in the Stellar transaction
    - If they don't match, print an error and remove the issue from the set 
- If they match, build the proof
    - Look at the Stellar transactions `ledger` field and retrieve the `externalizedMessages[]` and `transactionSet[]` from the map (built in #85)
    - (Optional: Reduce the externalized messages to a minimal amount that builds a valid quorum set (11-12 nodes) and check all the messages for validity, i.e. check their signatures, if they all contain the same txSetHash, etc.)
    - Call the `executeIssue()` extrinsic with the Stellar transaction, the externalized messages, and the transaction set

b-yap:
@ebma 
> On a new CancelIssueEvent or ExecuteIssueEvent remove the related Issue if it is contained in the local set
In a separate task, sequentially try to execute all the issues contained in the issue set (described in more detail below)

Uhm, when do I actually  "execute"? 
After receiving an `ExecuteIssueEvent`, or a spawned task that will go over the `RequestIssueEvent` set?

b-yap:
done by https://github.com/pendulum-chain/spacewalk/pull/146

PR #98, opened 2022-09-27
85 collect and store scp messages in vault client

b-yap:
(I moved the substrate-stellar-sdk to [stellar-relay](https://github.com/pendulum-chain/stellar-experiments/blob/85-collect-and-store-scp-messages-in-vault-client/src/lib.rs#L5))

Why draft:
1. To show what I've been doing.
2. I'm not sure how we want to access the maps? Via channels, or to be returned from a function? They're inside a structure called [ScpMessageCollector](https://github.com/pendulum-chain/spacewalk/compare/85-collect-and-store-scp-messages-in-vault-client?expand=1#diff-bd0b0ed44f1fb2b39149cd43c3c230f53d492cd9a8d4c3a159e4f32918bfc5aaR308-R316). 
3. Testing the run; that it should keep going beyond an hour.

To get it started, ~call the [collect_scp_messages](https://github.com/pendulum-chain/spacewalk/compare/85-collect-and-store-scp-messages-in-vault-client?expand=1#diff-bd0b0ed44f1fb2b39149cd43c3c230f53d492cd9a8d4c3a159e4f32918bfc5aaR476) function~. run the example.

For file saving, there's [EnvelopesMap, TxSetMap, ~TxHashMap~](https://github.com/pendulum-chain/spacewalk/compare/85-collect-and-store-scp-messages-in-vault-client?expand=1#diff-bd0b0ed44f1fb2b39149cd43c3c230f53d492cd9a8d4c3a159e4f32918bfc5aaR47-R52)
We're not saving the `TxHashMap` for now, since this requires discussion on how to save it to file and how to name it.

For old scpenvelope messages, these are transferred to a file depending on the constants `MAX_SLOTS_PER_FILE` and `MIN_EXTERNALIZED_MESSAGES`.  

For old transactionsets ~and the transaction hashes~,  saving to file depends on the constant ~`MAX_TXS_PER_FILE`~ `MAX_TXSETS_PER_FILE`

How I named the files: `<starting_slot_number>_<ending_slot_number>`

These names would make it easier to find tx or scpenvelope based on the slot number.

----

## Filter
Since we're not listening to StellarTransactions individually, we need to be able to filter and choose in the transactionSet what we want to process. 

Say if we're looking for a transaction with a specific issue-id in the memo, A filter should be created for it, and added to the list of filters. See the readme on how to do it.
And if we want to remove such filter, the method `remove_filter` is available.

Basically the `ScpMessageHandler` holds the task of running the `StellarOverlayConnection`.  We can pass it around, and easily use its methods to collect data from the `ScpMessageCollector`.  
This way, `ScpMessageHandler` solves:
> stellar-relay keeps running
> listen to outside events by adding functions we need in the `ScpMessageHandler`.


b-yap:
so @gonzamontiel the script you suggested before, we could use a script and go over the old files and just delete them. 
That's my idea.

b-yap:
by the way I tested this from the [stellar-experiment](https://github.com/pendulum-chain/stellar-experiments/blob/85-collect-and-store-scp-messages-in-vault-client/src/bin/example.rs):
branch `85-collect-and-store-scp-messages-in-vault-client`

just run `RUST_LOG=info cargo run`

and it will show something like this:
```
[2022-09-28T04:07:00Z INFO  stellar_relay::connection::services] Starting Handshake with Hello.
[2022-09-28T04:07:00Z INFO  stellar_relay::connection::connector::message_handler] Hello message processed successfully
[2022-09-28T04:07:00Z INFO  stellar_relay::connection::connector::message_handler] Handshake completed
[2022-09-28T04:07:04Z INFO  example] slot: 42867088 add to envelopes map
[2022-09-28T04:07:05Z INFO  example] ---- PID: 317 Handle TxSet----
...
[2022-09-28T04:08:31Z INFO  example] slot: 42867103 add to envelopes map
[2022-09-28T04:08:32Z INFO  example] ---- PID: 7821 Handle TxSet----
[2022-09-28T04:08:32Z INFO  example] saving old transactions to file: [42867088, 42867089, 42867090, 42867091, 42867092, 42867093, 42867094, 42867095, 42867096, 42867097, 42867098, 42867099, 42867100, 42867101, 42867102]
[2022-09-28T04:08:32Z INFO  example] slot: 42867103 inserting transacion set
[2022-09-28T04:08:38Z INFO  example] The map is getting big. Let's write to file:
[2022-09-28T04:08:38Z INFO  example] start slot is now: 42867103
...
```
sample files are found in their respective folders: 
[Map<Slot,ScpEnvelopes>](https://github.com/pendulum-chain/stellar-experiments/tree/85-collect-and-store-scp-messages-in-vault-client/scp_envelopes)

[Map<Slot,TransactionSet>](https://github.com/pendulum-chain/stellar-experiments/tree/85-collect-and-store-scp-messages-in-vault-client/tx_sets)

[Map<TransactionHash,Slot>](https://github.com/pendulum-chain/stellar-experiments/tree/85-collect-and-store-scp-messages-in-vault-client/tx_hashes)

Files are in bincode, so to read them, the function [read_file](https://github.com/pendulum-chain/stellar-experiments/blob/85-collect-and-store-scp-messages-in-vault-client/src/bin/example.rs#L160) will do.

clients/vault/src/stellar_oracle.rs
```diff
@@ -0,0 +1,507 @@
...
+pub type SlotEncodedMap = BTreeMap<Slot, SerializedData>;
+
+/// The slot is not found in the `StellarMessage::TxSet(...)`, therefore this map
+/// serves as a holder of the slot when we hash the txset.
+pub type TxSetCheckerMap = HashMap<TxSetHash, Slot>;
+
+/// Todo: these maps should be merged into one; but there was a complication (which differs in every run):
+/// Sometimes not enough `StellarMessage::ScpMessage(...)` are sent per slot;
+/// or that the `Stellar:message::TxSet(...)` took too long to arrive (may not even arrive at all)
+/// So I've kept both of them separate.
```
ebma, line 45: I would remove this 'TODO' comment. Yes, it was originally intended to have one map but it is not important, so maybe change it to something like 'There are separate maps for TX sets and SCP envelopes because...'

ebma (changes requested):
I noticed that the number of data files grow quite fast. So maybe we could tweak the parameters that more data is put into these files and we have less of them? Or do you think the parameters are good the way they are @b-yap? This is not really a blocker, I was just wondering. By the way, I think it is confusing that the files have the extension '.json' when they are not json. Maybe wholly remove the file extension.

Also, I noticed that you are not including the tests that are on the similar branch in the 'stellar-experiments' repo. You could also include them here maybe.

Furthermore, I would suggest extending the `get()` functions of the custom maps we have (EnvelopesMap, TxSetMap, TxHashMap) to actually do the file lookup if the requested key/slot is not found in the map. If I understand correctly, if the data in the map gets too large it is stored to file **but** retrieving it back from the file is not automatically handled by the map (which is very inconvenient for the one trying to get data from the maps).

clients/vault/Cargo.toml
```diff
@@ -48,9 +48,12 @@ sp-runtime = {git = "https://github.com/paritytech/substrate", branch = "polkado
 sp-std = {git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.24"}
 
 # Substrate Stellar Dependencies
-stellar-relay = {git = "https://github.com/pendulum-chain/stellar-experiments.git", branch = "85-collect-and-store-scp-messages-in-vault-client" }
+stellar-relay = {path = "../../stellar-relay" }
```
b-yap, line 51: so this is going to be in another PR, where I add the `stellar-relay` (aka the stellar-experiment) to spacewalk.

clients/vault/src/stellar_oracle/actor.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    async fn handle_message(&mut self, msg: Message, collector: &ScpMessageCollector) {
+        match msg {
+            Message::GetCurrentMap { sender } => {
+                let _ = sender.send(ScpMessageMaps::new(collector));
+            }
+        };
+    }
+
+    async fn run(&mut self, mut user: UserControls, mut collector: ScpMessageCollector) -> Result<(), Error> {
```
b-yap, line 56: This is the new bit I added, where it's either going to handle messages coming from the stellar-relay, or a message coming from the event listeners, wanting to get the map.

clients/vault/src/stellar_oracle/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+///
+/// 		let (sender,receiver) = tokio::sync::oneshot::channel();
+/// 		handler.get_map(sender).await?;
+///
+/// 		if let Ok(res) = receiver.await {
+/// 			log::info!("the map {:?}", res.envelopes_map().len());
+/// 		}
+/// 	}
+/// ```
+pub async fn handler(node_info: NodeInfo, connection_cfg: ConnConfig) -> Result<ActorHandler, Error> {
```
b-yap, line 117: This right here will run similarly with the demo, and returns a handler that  we can send a message to, to retrieve the map.

b-yap:
my todo:  move the stellar-relay (a.k.a. the stellar-experiments) to spacewalk.

@ebma... I get a little confused about how the scpenvelopes and externalmessages are being stored/used.
So following the event 
```
RequestIssue {
			issue_id: H256,
			requester: T::AccountId,
			amount: BalanceOf<T>,
			fee: BalanceOf<T>,
			griefing_collateral: BalanceOf<T>,
			vault_id: DefaultVaultId<T>,
			vault_stellar_public_key: StellarPublicKeyRaw,
			public_network: bool,
		},
``` 
I  traverse through all the transactions with a memo containing the `issue_id`? 
Since the oracle will already be running, whether or not it receives a `RequestIssue`.

Unless as stated in #86 
> When issuing, the user has to send a Stellar transaction with a payment to the vault
The vault continuously fetches the transactions for its Stellar account from the network and stores them locally
It would be okay to only store the new transactions after going online

Meaning the oracle will only run once an issue or redeem event happens? 

Or.. ^ the setup above is a different one? And like the demo, the oracle will only have the transaction (and then derive the `tx hash`) as an argument. Once a match is found, return a proof

clients/vault/src/stellar_oracle/mod.rs
b-yap: but i'm thinking, that it doesn't have to be the map. It could just be proof.

clients/vault/src/stellar_oracle.rs
b-yap: I forgot to reply to this :scream:
I did not want to extend the `get()` function, because the map shouldn't be allowed to access the files. Instead, I have FileHandlers that can be used to search for  transactions not found in the map.


For the tests... it's a todo. :bow:

clients/vault/Cargo.toml
b-yap: so this draft PR will fail.

ebma:
It depends on whether the vault has to handle an issue or redeem event.

When it sees a new RequestIssueEvent happening on the chain, it will expect the user to send a payment to the vault's Stellar address. This means, it is very similar to the demo. The vault listens to the stellar transactions that are targeting its account and then checks all of them if they contain a memo matching one of the `issue_id`s of a RequestIssueEvent.
But the oracle (ie SCP message and transaction set collection part) has to run all the time because we don't know when the user sends a relevant transaction. Thus, when the vault client starts, the oracle agent starts, and it will keep running until the vault client is shut down.

Does this answer your question?

b-yap:
Ah, one more thing @ebma .

The `RequestIssueEvent` for a certain _`issue_id`_ only happens **_once_** . In our demo, we're actually filtering based on [is_tx_relevant](https://github.com/pendulum-chain/stellar-experiments/blob/demo/src/bin/demo.rs#L739) which is not enough, I guess? 
So the additional work to be done is checking the `issue_id` against all `pending_transactions`?

Is it possible to [save](https://github.com/pendulum-chain/stellar-experiments/blob/demo/src/bin/demo.rs#L616-L618) only those dedicated to the vault's account?

ebma:
@b-yap yes you're right, `is_tx_relevant` would also have to check if the memo of the transactions contains a memo with one of the issue_id's that we know (i.e. one of the issues that is in the local list of the vault client, stored after a `RequestIssueEvent` occurred).

Not sure if we can only save the transactions dedicated to the vault's account though, I have to think about that again.

gonzamontiel:
> @b-yap yes you're right, `is_tx_relevant` would also have to check if the memo of the transactions contains a memo with one of the issue_id's that we know (i.e. one of the issues that is in the local list of the vault client, stored after a `RequestIssueEvent` occurred).
> 
> Not sure if we can only save the transactions dedicated to the vault's account, though, I have to think about that again. Update: Yes, it should be enough to only store transactions that are targeting the vault account in the transaction hash set. To be precise, we only have to keep Stellar transactions that contain a 'Payment' Operation with the vault's stellar address as the destination and contain a memo of 'MEMO_HASH' type.

The `is_tx_relevant()` function was only meant for demo, and of course some extra filtering. In this case, Marcel's right about adding the memo type to the filter and save only those txs.
Also, I remember we discussed about a) extracting the transactions from the tx set vs b) listen to single transactions. The implementation here is still based on a), do we have plans to change that?

b-yap:
I still prefer **a)**. We need the slot anyway, and it's in the `TransactionSet`. And you are right, filtering is involved in both options.  

 But I added a [message with a dynamic trait as a param](https://github.com/pendulum-chain/spacewalk/pull/98/files#diff-2a7c934f750ad5ea9d31e17003803dcc14418b5643e2d7a3c3b5f7d04d4293ffR29), where it can update the filter depending on the event: The `RequestIssueEvent`, `RedeemEvent`, etc..
My idea is... the user can conveniently pass the filter it wants: like the issue event,  it needs to check the `issueid` from the memo. 
```rust
pub trait TxHandler<T> {
    fn process_tx(&self, param: &T) -> bool;
}
```

So the `process_tx()` will be called in the [update_tx_hash_map](https://github.com/pendulum-chain/spacewalk/pull/98/files#diff-9714af5b9a8b07f9a221f24863a019ffd722c4419350fde648343e41145e285eR214-R226). If a match is found, it will add to the `pendingTransactions` vec. The `T` here would be `TransactionEnvelope`, so it's easy for the issue event to find a match from its memo.

I had to remove the subxts since the main target is only collecting the scpmessages and the txsets. On what to do with the pending transactions, I created a [function to call it](https://github.com/pendulum-chain/spacewalk/pull/98/files#diff-9714af5b9a8b07f9a221f24863a019ffd722c4419350fde648343e41145e285eR231-R232)... wherever code needs it.
^ Opinions... @gonzamontiel  @ebma

b-yap:
follow-up question on this @ebma , 
> To be precise, we only have to keep Stellar transactions that contain a 'Payment' Operation with the vault's stellar address as the destination and contain a memo of 'MEMO_HASH' type.

If  in a `TransactionSet` we find **_no_** meaningful stellar transaction for the vault, can we also opt to **_not_** save the TransactionSet as well?

ebma:
@b-yap to answer your second question, yes we could. But overall this is kind of related to this other ticket #95 we defined. You can either do these things now or add them in another pull request, I don't have a strong opinion on this.

b-yap:
@ebma I moved the tests.  :v:

clients/vault/src/oracle/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use collector::*;
+use errors::Error;
+use storage::*;
+use types::*;
+
+/// A filter trait to check whether `T` should be processed.
+pub trait FilterWith<T> {
+	/// unique identifier of this filter
+	fn id(&self) -> u32;
```
ebma, line 21: By the way, could we (the oracle library code) also just take this over so the user of the trait does not have to provide an id? Can we have some global counter variable, maybe, that is just incremented for every new filter? Otherwise the user would have to keep track of the IDs he already used which might be annoying if he creates filters at different places.

b-yap: hhm. yeah, we can do that. Alright. 
But how to delete them (SHOULD we need to delete, that's another question too) later on?

ebma: Ahhhh true, that's why we needed the user to provide it in the first place... 😅 
Well no, let's keep it then. But can we maybe make the ID a bit generic so that either an integer or a &str can be used? It might be more convenient if you can provide some kind of name for the filter and not just a number. Or if you don't want to make it generic, I'd prefer the &str over number as id.

b-yap: gotcha, i'll change the datatype.

PR #99, opened 2022-09-29
Create stellar relay pallet

### LINKED ISSUE #88: Create 'stellar relay' pallet
**Edit**: Some parts were moved to #94 because they fit better logically. 

This pallet is intended to contain things that are related to Stellar and shared by other spacewalk pallets i.e. the function to validate stellar transactions and the function to generate a secure unique ID for issue/redeem requests. _Other naming suggestions are welcome._  

### Storage
~~- nonce, i.e. an integer number incremented every time it is accessed~~
~~- last block number, used when creating new issue/redeem requests and needed to check whether the requests are 'expired'~~
- set of Stellar's tier 1 validators
    - the set has to be stored with additional information on the structure so that we can automatically derive valid quorum sets from it 

### Functions
- `validate_stellar_transaction(tx, externalizedMessages[], txSet)` validates the passed `tx` by checking the externalized messages and transaction set
~~- `get_secure_id(accountID)` returns a  unique Identifier by combining the `accountID` with the stored nonce and increasing it afterwards~~

### Extrinsics
- `updateTier1ValidatorSet(newValidatorSet)` extrinsic that is only callable by root (because this extrinsic should only be callable with a governance proposal or sudo)

### Set of tier 1 validators
We have to store this set with additional meta information to be able to derive a valid quorum set.
A possible solution would be to store the validator structs in a list, where a validator struct is comprised as follows:
```
struct Validator {
   name,
   public_key, // the public key used for signing the messages
   organization, // id of the organization this validator belongs to
   total_org_nodes // the total number of validator nodes that the organization has
}
```

### How to validate the stellar transaction
If the following checks succeed, we can be confident that this transaction was successfully executed on Stellar:
1. Check if the `tx` is included in the provided `txSet`
2. Check if all the received externalized messages belong to and were signed by the respective tier 1 validator node.
3. Derive the `txSetHash` based on the provided `txSet` and check if it matches the `txSetHash` contained in all the provided externalized messages
    - I think this is done by ordering the transactions by their transaction hash and then hashing all of them together, but this has to be confirmed in the stellar-core [[source code](https://github.dev/stellar/stellar-core/blob/19ea1e4e7989139b547346686d804430f730f699/src/herder/TxSetFrame.cpp#L216)](https://github.dev/stellar/stellar-core/blob/19ea1e4e7989139b547346686d804430f730f699/src/herder/TxSetFrame.cpp#L216)
4. Check that the tier 1 nodes that signed the externalized messages build a valid quorum set
    - For all externalized messages, find the corresponding validator in the stored list of validators. If one validator cannot be found, reject the transaction.
    - Collect the distinct organizations that occur in the validator structs related to the externalized messages
        - For each distinct organization, check that more than 1/2 of the `total_org_nodes` of this organization are contained in the externalized messages
    - Calculate the total amount of organizations in the tier 1 validator set
    - Check that the distinct organizations occurring in the validator structs related to the externalized messages are more than 2/3 of the total amount of organizations in the tier 1 validator set

ebma:
This PR adds the stellar relay pallet.

It also adds it to the testchain runtime and fixes its benchmarking setup. The weight template file was copied from [here](https://github.com/paritytech/substrate/blob/polkadot-v0.9.24/.maintain/frame-weight-template.hbs).

Closes #88.


pallets/stellar-relay/README.md
```diff
@@ -0,0 +1,38 @@
+# Stellar Relay Pallet
+
+## Testing
+
+To run the tests use:
+
+```bash
+cargo test --package pallet-stellar-relay --features runtime-benchmarks
```
b-yap, line 8: can we run the test without the feature `runtime-benchmarks`?

pallets/stellar-relay/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+			);
+			// Ensure that the number of organizations does not exceed the limit
+			ensure!(
+				organizations.len() as u32 <= T::OrganizationLimit::get(),
+				Error::<T>::OrganizationLimitExceeded
+			);
+
+			let validator_vec = BoundedVec::<Validator, T::ValidatorLimit>::try_from(validators)
+				.map_err(|_| Error::<T>::BoundedVecCreationFailed)?;
+			Validators::<T>::put(validator_vec);
```
b-yap, line 160: I forgot what the `put` does. 

Let's say the  `ValidatorLimit` is 10,  and we already have 8 in the `Validators` storage.
Then the user wants to update 5 _new_ validators (meaning keys that do not exist in `Validators`). Does this mean the 8 existing validators are removed and _replaced_ with the 5 new ones?

pallets/stellar-relay/README.md
ebma: Yes, you can run them without it (i.e. just `cargo test --package pallet-stellar-relay`). Adding the `runtime-benchmarks` feature will also make it run the test that is contained in 'benchmarks.rs' (which doesn't do much).

pallets/stellar-relay/src/lib.rs
ebma: It _sets_ the value i.e. the existing values are irrelevant and replaced every time the extrinsic is used. I updated the `update_tier_1_validator_set_works()` to better illustrate this behaviour.

pallets/stellar-relay/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+
+	// Extrinsics
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// This extrinsic is used to update/replace the current sets of validators and
+		/// organizations
+		#[pallet::weight(<T as Config>::WeightInfo::update_tier_1_validator_set())]
+		pub fn update_tier_1_validator_set(
```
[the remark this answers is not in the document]
ebma: Hmm, I think it is highly unlikely that you want to change only the organizations and not the validators. But there might be cases where you only want to update the validators and don't have to touch the organizations. 
IMO splitting this extrinsic is not necessary since even if only one of these two things is required to change, the overhead is just one database write which should be negligible.

ebma:
I added an additional change that makes the organization ID generic and configurable in the pallet config.

b-yap (approved):

PR #107, opened 2022-10-13
Add missing tier 1 organizations to genesis config

ebma:
I realized that I missed some of the tier 1 validator organizations (because they were not shown in the sidebar at [Stellarbeat](https://stellarbeat.io/organizations?network=public) but only in the organizations tab).


ebma:
Closing this because not all the organizations that stellarbeat lists as tier 1 organizations are actually in the tier 1 validator set. Only the ones in the big cluster in the center are.

PR #116, opened 2022-10-21
Use cache in github ci workflow

### LINKED ISSUE #93: Add caching for CI jobs
It would be nice to have something like [rust-cache](https://github.com/Swatinem/rust-cache) for our CI jobs to speed up the build process.

ebma:
Adds caches for the CI jobs. The `rust-cache` action caches the dependencies, speeding up their download time if the Cargo.lock file did not change. 'sccache' is used to save the compiled artifacts and thus speed up compilation times. 
Clippy [does not work with sccache](https://github.com/rust-lang/rust-clippy/issues/3840) however, so I had to remove it. IMO it should be fine to remove it because it just spits out some warnings, that we probably ignore most of the time anyway.

These changes can cut down the duration of CI runs from ~30 minutes to ~12 minutes. At least for subsequent runs, where the dependencies did not change but just the code.

It also adds the `--all-features` flag to cargo test, because it is needed for the vault tests (as we have to specify either parachain-metadata, or standalone-metadata, and optionally the multi-address feature).

Closes #93.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,11 +80,6 @@ jobs:
           command: fmt
           args: --all -- --check
 
-      - name: Clippy
-        run: |
-          rustup component add clippy 
-          cargo clippy --all -- -W warnings -W clippy::unwrap_used -W clippy::needless_collect 
-
       - name: Print sccache stats
         run: sccache --show-stats
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,13 @@ jobs:
         # Call `rustup show` as a hack so that the toolchain defined in rust-toolchain.toml is installed
         run: rustup show
 
+      - name: Use Cache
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           shared-key: "shared"
 
       - name: Save sccache
-        uses: actions/cache@v3
+        uses: pat-s/always-upload-cache@v3
```
gonzamontiel, line 56: what is the benefit here?

gonzamontiel (approved):
Looks cool.

.github/workflows/ci.yml
ebma: By default, the actions/cache@v3 only uploads the cache if the run was successful. Unfortunately, the setting to change this is not exported. That's why I exchanged it with this other one which is basically the same except that it uploads the cache regardless of the overall outcome.

PR #121, opened 2022-10-28
Remove public network from stellar relay

ebma:
Removes the `public_network` parameter from the `Organization` and `Validator` struct. This is a good idea because supporting both network types simultaneously would create a lot of overhead for storage and performance.

Refactors the tests so that instead of every test case having to instantiate the validators and organizations, it is configured only once in the mock genesis and the relevant variables are passed into the tests.
Also renames the module from `pallet_stellar_relay` to `stellar_relay` to make it consistent with the other crates in the directory.

I deliberately decided not to make the `public_network` configurable via an extrinsic. This parameter is configurable only once and this has to happen in the genesis config. Making this configurable after genesis bears lots of risk, because the registered vault and issue/redeem requests would become invalid and have to change at the same time which is a scenario that does not really make sense in practice.


testchain/node/src/chain_spec.rs
```diff
@@ -240,6 +279,12 @@ fn testnet_genesis(
 		security: SecurityConfig {
 			initial_status: if start_shutdown { StatusCode::Shutdown } else { StatusCode::Error },
 		},
+		stellar_relay: StellarRelayConfig {
+			validators,
+			organizations,
+			is_public_network: true,
```
gonzamontiel, line 285: `is_public_network` means that it's connected to mainnet, right? I don't get why, even in the testnet config, `is_public_network` needs to be true. Am I understanding the naming wrongly?

gonzamontiel (approved):
I left a question but otherwise lgtm.

testchain/node/src/chain_spec.rs
ebma: Yes, 'public network' is equal to mainnet. Good thing you noticed this, as it is actually an oversight. I changed it to false now.

PR #123, opened 2022-10-31
91 currency pallet

### LINKED ISSUE #91: Create 'currency' pallet
We should extend the functionalities offered by the `orml_tokens` pallet. The 'currency' pallet will be tightly coupled to the `orml_tokens` pallet. 

The currency pallet should offer:
- getters for the native, relay-chain, and wrapped currency ID
- a struct that can hold an amount and currency ID, and offers fixed point arithmetic for increased precision. It might also be useful to add helper functions e.g. for transferring and minting, to this struct

Should be pretty similar to [this](https://github.com/interlay/interbtc/tree/master/crates/currency/src) pallet.

gonzamontiel:
- Makes CurrencyId available from other modules
- Uses `orml_tokens` and traits

Closes  #91


gonzamontiel:
It is weird that so many files have conflicts, I merged `dev/mileston-3` a short time ago. I will solve them and re push,

ebma:
I thought it was because it was based on the main branch and that some changes there were not contained on the dev branch yet. But I now merged main to dev/milestone-3, and it didn't change anything. I also tried force-pushing the last commit again but this also didn't help...

pallets/currency/src/lib.rs
```diff
@@ -74,6 +74,18 @@ pub mod pallet {
...
+		#[pallet::constant]
+		type GetNativeCurrencyId: Get<CurrencyId<Self>>;
+
+		/// Relay chain currency e.g. DOT/KSM
+		#[pallet::constant]
+		type GetRelayChainCurrencyId: Get<CurrencyId<Self>>;
+
+		/// Wrapped currency e.g. IBTC/KBTC
+		#[pallet::constant]
+		type GetWrappedCurrencyId: Get<CurrencyId<Self>>;
```
gonzamontiel, line 87: What would this mean for us? We will have many "wrapped" assets, how does this play along with our defined `CurrencyId`

gonzamontiel (commented):
I left one message, if not urgent we can merge it as-is, but still need to figure that one out.

primitives/src/lib.rs
```diff
@@ -296,3 +313,180 @@ pub struct CustomMetadata {
...
+// 					str::from_utf8(code).unwrap(),
+// 					str::from_utf8(
+// 						stellar::PublicKey::from_binary(*issuer).to_encoding().as_slice()
+// 					)
+// 					.unwrap()
+// 				)
+// 			},
+// 		}
+// 	}
+// }
```
ebma, line 417: What about this part, wouldn't it be nice to be able to print the currency in logs?

For the `Token` you can use the functions that are added in the `create_currency_id` macro, eg
```
...
Self::Token(token_symbol) => {
				write!(f, "{:?} ({:?})", token_symbol.name(), token_symbol.symbol())
			},
			...
```

ebma (approved):
LGTM overall

pallets/currency/src/lib.rs
ebma: That's a very good question. I briefly checked the code of the other crates I worked on, and it's not used a lot so I think we might even be able to get rid of it. But I would keep it for now.

primitives/src/lib.rs
ashneverdawn: Oh yeah. I forgot I'd commented this out temporarily until I figured out how to get it to work with the RuntimeDebug macro which interlay uses. I guess for now it's better to have it on both runtime and wasm than on neither. Uncommented and pushed the change now. Thanks! 👍

primitives/src/lib.rs
```diff
@@ -385,36 +384,42 @@ impl TryInto<stellar::Asset> for CurrencyId {
...
+impl fmt::Debug for CurrencyId {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		match self {
+			Self::Token(token_symbol) => {
+				write!(f, "{:?} ({:?})", token_symbol.name(), token_symbol.symbol())
+			},
+			Self::ForeignAsset(id) => {
+				write!(f, "{:?}", id)
+			},
+			Self::Native => write!(f, "PEN"),
```
ebma, line 396: Maybe change this to say 'native' instead of 'PEN' since it might also be used on other chains.

ebma:
Ready to merge from my side.

Issue #122, opened 2022-10-28
Use horizon to check for transactions targeted towards vault's stellar account


ebma:
We should change the oracle agent to fetch horizon for transactions targeting its Stellar account instead of going through the transactions it receives from the stellar overlay network stream.  We already had a small discussion about this [here](https://github.com/pendulum-chain/spacewalk/pull/98#issuecomment-1282394046), where we concluded that we should keep using the transactions from the overlay network. However, fetching horizon is preferable because, let’s say, our client is offline, either due to network issues or because we shut it down (e.g. for upgrading). The client will miss lots of SCP messages and also all transaction sets and thus transactions. So once we start the client again, he would not know which transaction was targeting it. If we fetch the horizon server for the transactions targeting its Stellar account, we will always have access to the transactions we missed while being offline.

Since the ledger field is equal to the slot (statements proving this are [here](https://github.com/stellar/stellar-core/tree/master/src/herder#key-implementation-details) and [here](https://github.com/stellar/stellar-core/blob/ca2fb06059c15442cb4c9a8c89de1a8fc3579a39/src/herder/HerderImpl.cpp#L1113)), when we fetch transactions from the horizon server (see e.g. [this ](https://horizon-testnet.stellar.org/transactions/9e1dbad238cc70fa602e9f6fb3d78707cee6a767ce3fda094faec21915a68015) request), it includes a ledger field from which we immediately know which slot we are looking for. The client can then check its local maps or files to find the SCP messages for that slot, and if it does not find them fall back to requesting the overlay network or stellar archive nodes.
However, with this approach, we don't even need to store SCP messages in a file. When we continuously fetch the transactions for our vault's stellar account, we will see which transactions are relevant (because they contain a proper transaction memo) and which are not. Since the transaction data we fetch from horizon contains the ledger/slot, we can have some local vector ‘slots_that_are_relevant’, put that slot number in there, then send a GET_SCP_STATE message to the overlay. In our listening logic, we check if the slot we currently received things for is contained in ‘slots_that_are_relevant’.

To wrap this up, we should:
1) Make the oracle agent continuously fetch new transactions that are targeting its account from the horizon server (see the API documentation [here](https://developers.stellar.org/api/resources/accounts/transactions/)). Maybe implement this as a polling mechanism that requests the transactions every 6 seconds or so.
2) Check whether there are new transactions that it should handle (i.e. ones that contain a proper memo that matches one of the registered issue/redeem IDs)
[...]

PR #127, opened 2022-11-03
handle missed slots inside function get_envelopes. implemented option A

cr4pt0:


clients/vault/src/oracle/handler.rs
```diff
@@ -67,6 +71,9 @@ impl ScpMessageActor {
 			ActorMessage::GetPendingProofs { sender } => {
 				let _ = sender.send(self.collector.get_pending_proofs());
 			},
+			ActorMessage::GetScpState { missed_slot } => {
+				overlay_conn.send(StellarMessage::GetScpState(missed_slot.try_into().unwrap())).await;
```
ebma, line 75: This works just like that? I'm just asking because one day we talked about that maybe converting the number to xdr might be necessary but even better if this is not the case.

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -42,17 +43,36 @@ pub enum ProofStatus {
...
 
+	
 	/// Returns either a list of ScpEnvelopes or a ProofStatus saying it failed to retrieve a list.
 	fn get_envelopes(&self, slot: Slot) -> Result<UnlimitedVarArray<ScpEnvelope>, ProofStatus> {
-		let (envelopes, is_from_file) =
-			self._get_envelopes(slot).ok_or(ProofStatus::NoEnvelopesFound(slot))?;
+		let envelopes =
+			self._get_envelopes(slot);
+		
+		if let None = envelopes{
```
ebma, line 60: Let's not only do this when we don't have anything but also when we do have some but not enough.

clients/vault/src/oracle/collector/envelopes_handler.rs
```diff
@@ -43,9 +51,6 @@ impl ScpMessageCollector {
 				txset_hash_map.insert(txset_hash, slot);
 				// let's request the txset from Stellar Node
 				overlay_conn.send(StellarMessage::GetTxSet(txset_hash)).await?;
-
-				// check if we need to transfer the map to a file
-				self.check_write_envelopes_to_file(slot)?;
```
ebma, line 48: Is there a special reason why you remove this? It's not bad since we talked about removing the 'storing to file logic' overall anyways, but maybe this is not the right pr to do it.

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -42,17 +43,36 @@ pub enum ProofStatus {
 	NoTxSetFound(Slot),
 }
 
+const MAX_SLOT_TO_REMEMBER : u64 = 12;
```
ebma, line 46: Maybe move this to `clients/vault/src/oracle/constants.rs`

clients/vault/src/oracle/collector/proof_builder.rs
ebma: I would suggest something like 
```
let fetch_more = if envelopes.is_none() { true } else {
let envelopes = envelopes.unwrap();
envelopes.len()  < get_min_externalized_messages(self.is_public()) 
}

if fetch_more {
... // your logic
```

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -42,17 +43,36 @@ pub enum ProofStatus {
...
+			let action_sender = self.action_sender.clone();
+			if last_slot_index - MAX_SLOT_TO_REMEMBER < slot {
+				tokio::spawn(async move { 
+					action_sender
+					.send(ActorMessage::GetScpState { missed_slot : slot })
+					.await
+				});
+			}
+
+			return Err(ProofStatus::NoEnvelopesFound(slot))
```
ebma, line 72: Maybe change this to return `ProofStatus::LackingEnvelopes` and remove the `ProofStatus::NoEnvelopesFound(slot)` completely. I'd argue that having both variants in the enum is not really beneficial.

ebma (changes requested):
Looks good overall but I added some nitpicky comments.

clients/vault/src/oracle/handler.rs
cr4pt0: Yes. It works 100% like this. xdr is used behind the scene to pass value via network.

clients/vault/src/oracle/collector/envelopes_handler.rs
cr4pt0: sure. you are right.

clients/vault/src/oracle/collector/proof_builder.rs
cr4pt0: Done.

clients/vault/src/oracle/collector/envelopes_handler.rs
cr4pt0: Done.

clients/vault/src/oracle/collector/proof_builder.rs
cr4pt0: Done.

clients/vault/src/oracle/collector/proof_builder.rs
cr4pt0: Done.

ebma (approved):
LGTM now although the formatting might be a bit off. Are you using rustfmt and the precommit script see [here](https://github.com/pendulum-chain/spacewalk#development)?
Maybe add another commit after running `cargo fmt` on the changed files.

b-yap:
Guys I'm going to merge this if that's ok? Since the `GetScpState` is needed for [issue 122](https://github.com/pendulum-chain/spacewalk/issues/122) and I'm checking it out right now

ebma:
Yes you can merge it @b-yap. (Shoutout to you if you push a commit that `cargo fmt`s before doing so :p)

b-yap:
@ebma i did :innocent:

PR #129, opened 2022-11-08
Adapt vault registry pallet

gonzamontiel:
- Assigns always the same Stellar Public Key for deposits on the vault (no derivation scheme)
- Lets you register, change and liquidate vaults
- Still draft since I'm missing more testing


pallets/vault-registry/src/types.rs
```diff
@@ -660,27 +659,17 @@ impl<T: Config> RichVault<T> {
...
 		&mut self,
 		secure_id: H256,
 	) -> Result<StellarPublicKeyRaw, DispatchError> {
-		let public_key = self.new_deposit_public_key(secure_id)?;
-		// let btc_address = BtcAddress::P2WPKHv0(public_key.to_hash());
-		let stellar_address: StellarPublicKeyRaw = [0; 32];
-		// TODO change this to a stellar address
-		Ok(stellar_address)
+		self.new_deposit_public_key(secure_id)
 	}
```
ebma, line 673: Maybe we should completely remove these functions since we don't use deposit addresses and it might be confusing. Although keeping it like this is also okay since `new_deposit_public_key` has the comment that makes it clear. WDYT @gonzamontiel ?

ebma:
I fixed the tests (by re-adding some uncommented lines) and I also re-added the other currencies in the primitives. The reason is that they are used throughout the tests of the various pallets. You are right that we should remove them but I would prefer doing so in a separate PR where we then also make sure that the tests of the other pallets don't fail.

It looks good to me overall, but it would be nice if you could also add the rpc stuff to the testchain runtime (don't forget to downgrade jsonrpsee to v0.13.0) and rerun the benchmark (instructions are in the readme).

pallets/vault-registry/src/types.rs
gonzamontiel: I thought about removing them, my thought after was to keep that abstraction layer and add the comment. It can be a bit confusing or a newcomer, and If we never ever plan to adapt the deposit address in any way, then that function does not make sense and I would remove it.

ebma: Yes, I agree. But I doubt we will ever use/need something like this so I would vote for removing it.

ebma (approved):
LGTM, good job. 👍 

Maybe we should think about adding a fixed formatter for .toml files as well in the future because there are always useless changes. But that's something for another day so let's get this merged.

cr4pt0 (approved):

PR #133, opened 2022-11-15
Improve benchmark script

gonzamontiel:
- Improves the benchmarking script that lives in the `.maintain/` folder
- Adds documentation on how to use the script (I found the most appropriate place was `pallets/README.md`, but the rest of the readmes are untouched) 
- Run benchmarks again

Note: I noticed that after re-running benchmarks again and again, they always change a little bit, so not sure how often we need to re run them.


ebma:
Is this ready for review of do you still want to change things?

And it's true the benchmarks always change a little. I think it's because of
```
--steps=100 \
--repeat=10
```
which are quite low so the few measurements the benchmark does always deviate a bit.

gonzamontiel:
> Is this ready for review or do you still want to change things?
> 
> And it's true the benchmarks always change a little. I think it's because of
> 
> ```
> --steps=100 \
> --repeat=10
> ```
> 
> which are quite low so the few measurements the benchmark does always deviate a bit.

It's ready for review but `cargo test` is failing, I'm wondering why since I didn't modify anything.

.maintain/run-benchmark
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
+
+#### Function definitions
+print_help () {
+  echo "USAGE: \\n\
+  \\trun-benchmark.sh --all\\n\
+  \\trun-benchmark.sh --pallet <pallet-name>\\n\
```
ebma, line 7: Nit: This is not accurate since the filename is 'run-benchmark'. I think I would prefer it to be named 'run-benchmark.sh' anyways, but if we keep the current name we should maybe change this help statement.

pallets/README.md
```diff
@@ -10,31 +10,69 @@ To run the tests use:
...
 ```
 
 ```bash
-# Show all available benchmarks
-./target/release/spacewalk-standalone benchmark pallet --list
-
-# Show benchmarks for a specific pallet
-./target/release/spacewalk-standalone benchmark pallet -p <pallet-name> -e '*' --list
+# Show benchmarks for this pallet
+./target/release/spacewalk-standalone benchmark pallet -p fee -e '*' --list
```
ebma, line 26: Is there a reason you changed it to use the 'fee' pallet in this README? Since this is the 'general' README for all pallets I wouldn't put instructions for a specific pallet in here.

pallets/README.md
```diff
@@ -10,31 +10,69 @@ To run the tests use:
...
---steps 20 \
---repeat 10 \
---output pallets/<pallet-name>/src/weights.rs \
+--chain=dev \
+--pallet=fee \
+--extrinsic='*' \
+--steps=100 \
+--repeat=10 \
+--wasm-execution=compiled \
+--output=pallets/fee/src/default_weights.rs \
```
ebma, line 39: same here

ebma (commented):
I added some nitpicky comments.

But this is some great stuff and really helpful; good job coming up with that. 👍

ebma:
@gonzamontiel can you address the comments from my review so we can merge this soon?

.maintain/run-benchmark
gonzamontiel: you are right, I left `run-benchmark`

gonzamontiel:
> @gonzamontiel can you address the comments from my review so we can merge this soon?

sure thing! it's ready

ebma:
Looks good, thanks for changing it 👍

PR #135, opened 2022-11-17
Stellar Overlay Connection integration tests

cr4pt0:


cr4pt0:
Resolve #140
Add function to disconnect connection to overlay network

clients/stellar-relay/src/connection/overlay_connection.rs
```diff
@@ -114,3 +121,127 @@ impl StellarOverlayConnection {
...
+	}
+
+	fn create_node_and_conn() -> (NodeInfo, ConnConfig) {
+		let secret =
+			SecretKey::from_encoding("SBLI7RKEJAEFGLZUBSCOFJHQBPFYIIPLBCKN7WVCWT4NEG2UJEW33N73")
+				.unwrap();
+		let node_info = NodeInfo::new(19, 21, 19, "v19.1.0".to_string(), &TEST_NETWORK);
+		let cfg = ConnConfig::new("34.235.168.98", 11625, secret, 0, false, true, false);
+		(node_info, cfg)
+	}
```
ebma, line 246: Nit: I would move the helper functions above the first test so that someone reading the code has seen it before it is used.

clients/stellar-relay/src/connection/services.rs
```diff
@@ -277,6 +279,12 @@ pub(crate) async fn connection_handler(
 	loop {
 		match timeout(Duration::from_secs(connector.timeout_in_secs), actions_receiver.recv()).await
 		{
+			Ok(Some(ConnectorActions::Disconnect)) => {
+				log::error!("dropped");
+				drop(w_stream);
+				return Ok(())
+			},
+
```
ebma, line 287: Is there a reason why you put this here and not into the `_connection_handler()` where the other actions are handled?

clients/stellar-relay/src/connection/services.rs
```diff
@@ -277,6 +279,12 @@ pub(crate) async fn connection_handler(
 	loop {
 		match timeout(Duration::from_secs(connector.timeout_in_secs), actions_receiver.recv()).await
 		{
+			Ok(Some(ConnectorActions::Disconnect)) => {
+				log::error!("dropped");
```
ebma, line 283: I would change this message to "Disconnected by dropping the TCP stream". And also, this does not have to be logged with 'error' level since it is intended.

ebma (commented):
It's good overall, I just added some nitpicky comments.

However, I'm wondering if we should add some extra logic to prevent the user from using the overlay connection after the disconnect() is called? But maybe that is unnecessary for now.

clients/stellar-relay/src/connection/services.rs
cr4pt0: There reason why i put this here it is because that inside of _connection_handler we have unique reference
```
async fn _connection_handler(
	actions: ConnectorActions,
	connector: &mut Connector,
	w_stream: &mut tcp::OwnedWriteHalf,
) -> Result<(), Error> {...}
```

We need to call drop(..) function for value and not for reference.


```
#[derive(Debug)]
struct X(i32);
impl Drop for X{
    fn drop(&mut self){
        println!("DROP");
    }
}

fn main() {
    let mut x = X(1);
    r(&mut x);
    println!("{:#?}", x);
    drop(x);
    
}

fn r(v : &mut X)
{
    println!("try drop");
    drop(v);
    println!("try drop");
}

```

cr4pt0:
> 
Once we call `disconnect `we still able call listen method to get message that was already received from stellar node before we called `disconnect`. 
We can add extra logic to `send` method to prevent sending `ConnectorActions::SendMessage(message)` via `mpsc` 
@ebma 
kind of this options
```
pub async fn send(&self, message: StellarMessage) -> Result<(), Error> {
        if !self.is_disconected{
                self.actions_sender
		.send(ConnectorActions::SendMessage(message))
		.await
		.map_err(Error::from)
         }
         Ok(())
}
```

```
pub async fn send(&self, message: StellarMessage) -> Result<(), Error> {
         if self.is_disconected{
                return Error(...)
         }
         self.actions_sender
	 .send(ConnectorActions::SendMessage(message))
	.await
	.map_err(Error::from)
}
```

ebma:
Yeah, sending should not be possible after disconnecting, right?

I think I would prefer the second option from your code proposals. Can you add it to the code?

cr4pt0:
> Yeah, sending should not be possible after disconnecting, right?
> 
> I think I would prefer the second option from your code proposals. Can you add it to the code?

Done!

clients/vault/src/oracle/handler.rs
```diff
@@ -78,7 +78,7 @@ impl ScpMessageActor {
 					.await;
 			},
 			ActorMessage::Disconnect => {
-				overlay_conn.disconnect().await;
+				panic!("Should disconnect from run method")
```
ebma, line 81: This somehow looks worrying but I'm not sure how to improve it.

ebma (commented):
The formatting is off again, please fix that and I will approve afterwards.

cr4pt0:
> The formatting is off again, please fix that and I will approve afterwards.

Done!

PR #136, opened 2022-11-18
122 use horizon to check for transactions targeted towards vaults stellar account

b-yap:
not yet ready!
Needs more testing.

- [x] filter the memo for the the incoming horizon transactions
- [ ] work on the test for the polling ^
- [ ] test for breaking changes (especially @cr4pt0 's code for retrieving scpenvelopes)


clients/vault/examples/oracle.rs
```diff
@@ -54,25 +54,63 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+						p.envelopes().len()
+					);
+					old_slot = 0;
+				},
+				other => {
+					tracing::info!("no proof yet for {}: {:?}", old_slot, other);
+				},
+			}
+			let res = handler.get_size().await?;
+			tracing::info!("  --> envelopes size: {}", res);
```
ebma, line 113: some comments here on what the intention is would be nice to have

clients/vault/src/horizon/client.rs
```diff
@@ -0,0 +1,254 @@
...
+		&self,
+		is_public_network: bool,
+	) -> Result<HorizonTransactionsResponse, Error> {
+		let vault_keypair: SecretKey = SecretKey::from_encoding(&self.vault_secret_key).unwrap();
+		let vault_address = vault_keypair.get_public();
+
+		let request_url = String::from(horizon_url(is_public_network)) +
+			"/accounts/" + str::from_utf8(vault_address.to_encoding().as_slice())
+			.map_err(|_| Error::HttpFetchingError)? +
+			"/transactions?order=desc&limit=1";
```
ebma, line 82: The `limit=1` is problematic. We lived with it back in the days but we should tackle it now. We should fetch more than just the latest transaction because with this it might happen that we miss some.

I would suggest that, we do an initial fetch for the latest transaction, process that and somewhere store the `pagingToken` of that transaction. Now, for subsequent fetches, we can use the `pagingToken` (contained in the latest transaction we fetched) in the next calls, to be able to receive new transactions in batches. 
Info about pagination on stellar is [here](https://developers.stellar.org/api/introduction/pagination/).  But the relevant documentation on how to use a paging token when querying the transactions of an account is [here](https://developers.stellar.org/api/introduction/pagination/) where it is explained how to use it as a `cursor`.
You should use the pagingToken as cursor, set the order to "asc" and the limit to `20` or so.

clients/vault/src/horizon/client.rs
```diff
@@ -0,0 +1,254 @@
...
+					},
+				}
+			}
+		}
+	}
+}
+
+pub async fn poll_horizon_for_new_transactions(
+	vault_secret_key: String,
+	issue_set: Arc<Mutex<IssueRequests>>,
```
ebma, line 174: Nit: Using an [RwLock](https://docs.rs/tokio/0.2.22/tokio/sync/struct.RwLock.html) might be more elegant because it allows distinguishing between read and write access.

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -1,160 +1,261 @@
...
+	/// The available ScpEnvelopes are not enough to create a proof, and the slot is too far back
+	/// to fetch more.
+	LackingEnvelopes,
+	/// More envelopes have been requested from the Stellar Node, so user needs to wait.
+	WaitForMoreEnvelopes,
+	/// No ScpEnvelopes found and the slot is too far back.
+	NoEnvelopesFound,
+	/// Envelopes is fetched again, and user just needs to wait for it.
+	WaitForEnvelopes,
+	/// TxSet is not found and the slot is too far back.
```
ebma, line 64: Whats the difference between `WaitForEnvelopes` and `WaitForMoreEnvelopes`?

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -1,160 +1,261 @@
...
+		slot: Slot,
+		overlay_conn: &StellarOverlayConnection,
+	) -> ProofStatus {
+		// get the SCPEnvelopes
+		let envelopes = match self.get_envelopes(slot) {
+			Ok(envelopes) => envelopes,
+			Err(neg_status) => {
+				match &neg_status {
+					// let's fetch from Stellar Node again
+					ProofStatus::WaitForMoreEnvelopes | ProofStatus::WaitForEnvelopes => {
```
ebma, line 133: Do we really need separate types for this? Maybe just having `WaitForEnvelopes` is enough.

clients/vault/src/system.rs
```diff
@@ -57,22 +74,78 @@ impl VaultService {
...
+		tracing::info!("public key binary {:?}", public_key_binary);
+
+		let pk =
+			PublicKey::from_encoding("GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN")
+				.unwrap();
+		tracing::info!("mainnet {:?}", pk.as_binary());
+		let pk =
+			PublicKey::from_encoding("GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC")
+				.unwrap();
+		tracing::info!("testnet {:?}", pk.as_binary());
```
ebma, line 110: You can remove this again, I just used it to find out the binaries for the stellar issuers of some USDC on stellar haha. No need to keep it. The values are used in the testchain configuration of the (not yet merged) vault communication branch.

clients/vault/src/system.rs
```diff
@@ -57,22 +74,78 @@ impl VaultService {
...
 		let err_provider = self.spacewalk_parachain.clone();
 		let err_listener = wait_or_shutdown(self.shutdown.clone(), async move {
 			err_provider
-				.on_event_error(|e| tracing::debug!("Received error event: {}", e))
+				.on_event_error(|e| tracing::debug!("Received error event: {:?}", e))
 				.await?;
 			Ok(())
 		});
+		// issue handling
+		let issue_set: Arc<Mutex<IssueRequests>> = Arc::new(Mutex::new(IssueRequests::new()));
```
ebma, line 135: Again, I'd replace it with RwLock.

ebma (commented):
It is quite some code that changed, and I only understood how everything is connected while reading the last file. Meaning one day there might be room for some refactoring, eg. by putting the horizon logic into a whole separate module together with other stellar helpers as I did [here](https://github.com/pendulum-chain/spacewalk/tree/130-add-vault-communication/clients/wallet).  But it doesn't have to happen here, we can do it another day.

But overall I like the code changes, there are some neat things like the watch_slot thing and some ugly code vanished in the refactorings. 👍

clients/vault/src/oracle/collector/proof_builder.rs
b-yap: removed those extras variants.

clients/vault/src/oracle/collector/proof_builder.rs
b-yap: removed extra variants

clients/vault/examples/oracle.rs
b-yap: i have added comments. And although it runs okay, it's not yet complete. There's the get txse via the archive to include. I wanted to use this, to test the breaking changes.

Issue #120, opened 2022-10-28
Drop connection after the defined retries and timeouts


b-yap:
In the oracle.rs or even the connect.rs example, the log keeps going with _`peeking empty`_ even after the the retry and timeouts are exhausted.
Reported by @cr4pt0 .

Modify stellar-relay [to stop logging this error and  to stop the program completely.](https://github.com/pendulum-chain/spacewalk/blob/8cf9ccd2cad3de34964cc05a894f94f024bc9aad/clients/stellar-relay/src/connection/services.rs#L171) The retry variable should be incremented every time such error happens.

cr4pt0:
Merged
https://github.com/pendulum-chain/spacewalk/pull/142

PR #139, opened 2022-11-21
Drop connection after the defined retries and timeouts

cr4pt0:


ebma (changes requested):
Ticket #120 is not linked here. Please consider adding some description to your pull requests. Even if it is just one sentence and the wording eg "Closing #120". 

Also, if I understand correctly, the ticket was about also removing this log statement. And besides that, I'm not even sure if this works well in practice because I am not sure how many times this match branch is gonna be called before we receive a 'real' message. Did you test if the relay still works as expected after changing this @RustNinja ? I am just concerned that increasing the retry variable here might lead to the relay disconnecting every time before meaningful messages are received.

PR #142, opened 2022-11-22
Drop connection after the defined retries and timeouts.  Closing issue #120

cr4pt0:
"peeking empty" happens when we connect to node with incorrect overlay_version:

there is new version that works good without "peeking empty" error: 

`let node_info = NodeInfo::new(19, 25, 23, "v19.5.0".to_string(), network);`

this is the old one that we stuck forever with "peeking empty" error:

`let node_info = NodeInfo::new(19, 21, 19, "v19.1.0".to_string(), network);`

There is no chance that we we will receive any message once we go to this case statement. 

```
Ok(Ok(0)) => {
     log::error!("peeking empty");
}
```

introduce retry_read variable. increment each time received zero size

Closing https://github.com/pendulum-chain/spacewalk/issues/120

### DIFF
--- a/clients/stellar-relay/src/connection/services.rs
+++ b/clients/stellar-relay/src/connection/services.rs
@@ -168,6 +168,7 @@ pub(crate) async fn receiving_service(
 			.await
 		{
 			Ok(Ok(0)) => {
+				retry += 1;
 				log::error!("peeking empty");
 			},
 			Ok(Ok(_)) if lack_bytes_from_prev == 0 => {
--- a/clients/stellar-relay/src/connection/services.rs
+++ b/clients/stellar-relay/src/connection/services.rs
@@ -153,6 +153,7 @@ pub(crate) async fn receiving_service(
 	retries: u8,
 ) -> Result<(), Error> {
 	let mut retry = 0;
+	let mut retry_read = 0;
 	let mut proc_id = 0;
 
 	// holds the number of bytes that were missing from the previous stellar message.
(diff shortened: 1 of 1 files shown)


ebma (commented):
Thanks a lot for the description of the PR, and I also think the code is better now. 👍 

But just for some understanding before we merge this: so you are saying that once we get into this `Ok(Ok(0)) => {` match branch we will stay in that forever, ie while testing you experienced that you never got another message afterwards? Is this always the case?

Because if that's true we should maybe even think about immediately returning an error since there is no reason to retry if it will stay the same forever.

cr4pt0:
> 

Yes. Exactly. Once we receive  and went into `Ok(Ok(0)) => {}`  i never receive something new.
Of cause we can return error immediately and never go this statement again.
But I pepefer to log banch a log just to have attention that it happen several time.
Because to return the error at the first time it is a really big changes for such an important component like stellar-relay receiver.

ebma:
But is it a big change to return the error immediately? I don't see why it would be. 
However, since I am not sure what the inner workings of the stellar-overlay are in this regard, it would be fine for me to keep it the way you did it. If we encounter issues with that in the future, the error message should point us into the right direction.

ebma (approved):

cr4pt0:
> But is it a big change to return the error immediately? I don't see why it would be. However, since I am not sure what the inner workings of the stellar-overlay are in this regard, it would be fine for me to keep it the way you did it. If we encounter issues with that in the future, the error message should point us into the right direction.

It is not a big changes to return error immediately but it is a important changes in perspective new functionality.

PR #143, opened 2022-11-24
continuation of https://github.com/pendulum-chain/spacewalk/pull/136

b-yap:


clients/vault/examples/oracle.rs
```diff
@@ -54,25 +55,101 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+			}
+		// for every multiples of 6, let's watch out for a slot.
+		} else if counter % 6 == 0 {
+			// let's watch out for a slot that is 5 steps away from the current one.
+			handler.watch_slot(last_slot + 5).await?;
+
+		// for every multiples of 4, let's try to get proof of an old slot.
+		} else if counter % 4 == 0 {
+			let check_slot = if get_from_random {
+				// let's get a proof of a slot 10 0r 100 steps away from the current one.
```
ebma, line 111: Nit: There is a small typo between 10 and 100

clients/vault/src/horizon/client.rs
```diff
@@ -0,0 +1,279 @@
...
+	last_tx_id: Option<Vec<u8>>,
+}
+
+impl<C: HorizonClient> HorizonFetcher<C> {
+	pub fn new(client: C, vault_secret_key: String) -> Self {
+		Self { client, vault_secret_key, last_tx_id: None }
+	}
+
+	/// Fetch recent transactions from remote and deserialize to HorizonResponse
+	/// Since the limit in the request url is set to one it will always fetch just one
```
ebma, line 68: This comment is thankfully outdated now

clients/vault/src/horizon/client.rs
```diff
@@ -0,0 +1,279 @@
...
+
+		let mut request_url = String::from(horizon_url(is_public_network)) +
+			"/accounts/" + str::from_utf8(vault_address.to_encoding().as_slice())
+			.map_err(|_| Error::HttpFetchingError)? +
+			"/transactions?";
+
+		if !next_page.is_empty() {
+			request_url.push_str(&format!("cursor={}&", next_page));
+		}
+		request_url += "order=desc&limit=20";
```
ebma, line 85: I guess this should actually be `order=asc&limit=20`, or does it work? Because I think otherwise you get the transactions older than that paging token but we want the transactions that are more recent. 

Nit: It might also be a good idea to extract the `limit` to a constant and parse it into the string.

clients/vault/src/horizon/client.rs
```diff
@@ -0,0 +1,279 @@
...
+		};
+
+		let mut paging_token = PagingToken::new();
+		for transaction in transactions {
+			// update the paging_token
+			if let Some(page) = transaction.paging_token() {
+				paging_token = page;
+			} else {
+				paging_token = PagingToken::new();
+			}
```
ebma, line 155: Is this needed? I would assume every transaction has a paging token, or is that not the case?
Since you instantiate the paging_token outside the loop anyways, I guess it is safe to remove the else branch.

ebma: Also, I think I would prefer if we can find the highest paging_token in all transactions and choose it to be the new one explicitly. 
Doing it inside this loop might work but then we rely on the transactions to be in a particular order (ie, the latest transaction has to be the last one in the vector), otherwise the paging token is not set to the latest and we might fetch and process duplicate transactions eventually. 
Since the paging_token should always increase over time it should be safe to assume that the highest is the latest.

clients/vault/src/system.rs
```diff
@@ -293,19 +289,24 @@ where
...
 		shutdown: ShutdownSender,
 	) -> Result<Self, Error> {
-		let stellar_wallet = StellarWallet::from_secret_encoded(&config.stellar_vault_secret_key)?;
+		let is_public_network = spacewalk_parachain.is_public_network().await?;
+
+		let mut stellar_wallet =
+			StellarWallet::from_secret_encoded(&config.stellar_vault_secret_key)?;
+		stellar_wallet.set_is_public_network(is_public_network);
 		let stellar_wallet = Arc::new(stellar_wallet);
+
```
ebma, line 303: I assume you went the extra step of making the service creation async to ensure the wallet's `is_public_network` parameter is set correctly the moment it is created? I would then actually propose removing the `set_is_public_network` method altogether, and pass it into the the `from_secret_encoded` function. 

The reason why I added the `set_is_public_network`  in the first place is that I did not want to make the service creation async like you did but needed to set the network afterward. But generally we should not need to set the network after wallet creation so it's safe to remove.

clients/vault/src/system.rs
```diff
@@ -558,3 +520,42 @@ pub(crate) async fn is_vault_registered(
...
+	let secret = SecretKey::from_encoding(stellar_vault_secret_key).unwrap();
+	let public_key_binary = hex::encode(secret.get_public().as_binary());
+	tracing::info!("public key binary {:?}", public_key_binary);
+
+	let node_info = NodeInfo::new(19, 21, 19, "v19.1.0".to_string(), network);
+	let cfg = ConnConfig::new(tier1_node_ip, 11625, secret, 0, true, true, false);
+
+	// todo: add vault addresses filter
+	let addresses = vec![];
```
ebma, line 556: Can't we actually add the public key of our stellar account ie `secret.get_public()` here? Or what are these addresses for again?

clients/wallet/src/horizon.rs
```diff
@@ -231,36 +234,35 @@ impl<C: HorizonClient> HorizonFetcher<C> {
...
 		for transaction in transactions {
 			let tx = transaction.clone();
 			let id = tx.id.clone();
 			if is_tx_relevant(&tx, issue_hashes.clone()).await {
-				match watcher.watch_slot(tx.ledger.try_into().unwrap()).await {
+				match w.watch_slot(tx.ledger.try_into().unwrap()).await {
 					Ok(_) => {
-						tracing::info!("following transaction {:?}", String::from_utf8(id));
+						tracing::info!("following transaction {:?}", String::from_utf8(id.clone()));
+						println!("following transaction {:?}", String::from_utf8(id));
```
ebma, line 261: Why do we need the additional println statements?

ebma (commented):
So I assume you decided not to write proper tests? Because right now the tests still don't assert anything

cr4pt0 (approved):

ebma:
I added some more refactors and removed the `vault_addresses` field from the scp collector because as far as I can tell it is not used and also not needed because of our watch_slot approach.

PR #144, opened 2022-11-24
Fix vault client integration tests

ebma:
This PR adds and fixes the previously removed integration tests for the vault client. For this, the subxt-client package, whose purpose is to be able to run a standalone chain for testing purposes, was re-added. Also, other tests for the runtime were added (two of which fail unfortunately, but I don't know how to emulate the error properly).

This also adds the RPC APIs for the oracle and security pallets because they were used in the tests. 

PS: The testchain node was changed so that it can be run with manual-seal consensus, meaning that you can call a function that will force the runtime to author a new block immediately. It will not produce blocks automatically. This is convenient because now the tests can manually author blocks with a simple function call instead of waiting 6 seconds.

### DIFF
--- a/clients/runtime/src/tests.rs
+++ b/clients/runtime/src/tests.rs
@@ -1,0 +1,134 @@
+#![cfg(test)]
+
+use std::{convert::TryInto, time::Duration};
+
+use sp_keyring::AccountKeyring;
+
+use primitives::{StellarPublicKeyRaw, TokenSymbol};
+
+use crate::{integration::*, FeedValuesEvent, OracleKey, VaultId};
+
+use super::{
+	CollateralBalancesPallet, CurrencyId, FixedPointNumber, FixedU128, OraclePallet,
+	SecurityPallet, StatusCode, Token, TryFromSymbol, VaultRegistryPallet,
+};
+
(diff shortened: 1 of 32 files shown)


clients/runtime/src/tests.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use super::{
+	CollateralBalancesPallet, CurrencyId, FixedPointNumber, FixedU128, OraclePallet,
+	SecurityPallet, StatusCode, Token, TryFromSymbol, VaultRegistryPallet,
+};
+
+const DEFAULT_TESTING_CURRENCY: CurrencyId = Token(TokenSymbol::KSM);
+const DEFAULT_WRAPPED_CURRENCY: CurrencyId = CurrencyId::AlphaNum4 {
+	code: *b"USDC",
+	issuer: [
```
gonzamontiel, line 19: In literals such as this one, it is always useful to comment the human-readable issuer account for reference and future readers.

gonzamontiel (approved):
It's quite a lot of code and looks overall good. There are a couple of TODO's, but I ran the tests and they passed okay. I'd say LGTM since tests are usefull for other branches.

PR #128, opened 2022-11-08
86 listen to and execute issue events in vault client

b-yap:
This is just a draft!  Not for merging.

Based on branch `90-create-issue-pallet` and `main`. 

The files to focus on are in the clients dir.

PR #146, opened 2022-12-01
Add issue request handling to vault client

ebma:
Supercedes #128.


clients/vault/src/system.rs
```diff
@@ -425,10 +427,24 @@ impl VaultService {
...
+			),
+			(
+				"Listen for Issue Requests",
+				run(issue::listen_for_issue_requests(
+					self.spacewalk_parachain.clone(),
+					secret_key.clone(),
+					issue_map.clone(),
+				)),
+			),
+			(
```
ebma, line 443: Should we also run the `listen_for_executed_issues` here?

b-yap: 😬 oh yes, i forgot. I'll add it right away.

b-yap:
@ebma about checking the amount: is it safe to assume that it's just as simple as 
```
issue_amount_from_event == amount_from_transaction
``` 
?
https://github.com/pendulum-chain/spacewalk/blob/86-listen-and-execute-issue-events/clients/vault/src/issue.rs#L280-L281

other than that, I think this is good for review (since the code will be refactored anyway)

ebma:
Looks good to me. We should fix/re-add the one cancelation integration test once redeem/replace are done however.

PR #147, opened 2022-12-09
Refactor oracle agent architecture

ebma:
This PR simplifies the oracle agent code so that it is easier to use in our vault client. 

A new 'Agent' struct is constructed, that replaces the handler and watcher. The big simplification here is that the calling `get_proof` will now not return a proof-status but the whole proof, meaning that the caller might have to wait longer but using it gets way simpler.

Still WIP


ebma:
I'm finally gonna merge this now.

PR #148, opened 2022-12-13
Add handling for redeem/replace events to vault

ebma:
Adds the handling for redeem and replace events to the vault client.
Also adds the last missing integration tests.

The `send_payment_to_address()` function of the StellarWallet struct now requires a mutable reference to `self` because it now uses and sets a local account sequence number. This number is necessary because, when calling this function multiple times before a ledger close, all transactions would be created with the same sequence number, meaning that only one of them can go through and the others are rejected by the stellar network. Keeping a local copy of this sequence number mitigates this problem because we can increase it for every new transaction call.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10264,6 +10264,7 @@ dependencies = [
  "jsonrpc-core-client",
  "lazy_static",
  "mockall 0.8.3",
+ "nonzero_ext",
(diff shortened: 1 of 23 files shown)


ebma:
I'm merging this without a review because it is urgent

Issue #110, opened 2022-10-19
Extend stellar-relay pallet to allow for seamless update of validator set


ebma:
Currently, the stellar-relay pallet allows updating the validator set via calling an extrinsic, but doing so will immediately change the set on-chain. It would be better to extend this functionality such that it is possible to have two configurations in parallel and then use the newer one once a certain block height is reached.

This would require us to change the stellar-relay pallet as follows:
- Add three more on-chain storage items, so that overall it would be something like
   - `old_organizations`
   - `old_validators` 
   - `new_organizations`
   - `new_validators`
   - `commissioning_block_height`
- Change the `update_tier_1_validator_set(validators, organizations)` extrinsic to `update_tier_1_validator_set(validators, organizations, enactment_block_height)`
   - This extrinsic will then save the current values of `new_organizations` and `new_validators` to `old_organizations` and `old_validators` and save the supplied function arguments `new_organizations`, `new_validators` and `commissioning_block_height` respectively
- Change the `validate_stellar_transaction()` function to use the stored organization and validator based on the current block height and the value of `commissioning_block_height`

ebma:
Closed by #150.

PR #149, opened 2022-12-14
Stellar relay pallet validator set #110

cr4pt0:
Implement #110

### DIFF
--- a/pallets/stellar-relay/src/lib.rs
+++ b/pallets/stellar-relay/src/lib.rs
@@ -527,8 +527,7 @@ pub mod pallet {
 				transaction_set.txes.get_vec().iter().any(|tx| tx.get_hash(&network) == tx_hash);
 			ensure!(tx_included, Error::<T>::TransactionNotInTransactionSet);
 
-			let is_new_data = BlockHeight::<T>::get() > <frame_system::Pallet<T>>::block_number();
-
+			let is_new_data = <frame_system::Pallet<T>>::block_number() >= BlockHeight::<T>::get();
 			
 			let validators = if is_new_data {
 				// Check if all externalized ScpEnvelopes were signed by a tier 1 validator
--- a/pallets/stellar-relay/src/lib.rs
+++ b/pallets/stellar-relay/src/lib.rs
@@ -110,6 +110,17 @@ pub mod pallet {
 	}
 
 	#[pallet::storage]
+	#[pallet::getter(fn organizations_old)]
+	pub type OrganizationsOld<T: Config> =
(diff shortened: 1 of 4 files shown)


pallets/stellar-relay/src/lib.rs
```diff
@@ -448,6 +467,30 @@ pub mod pallet {
 				Error::<T>::OrganizationLimitExceeded
 			);
 
+			let validators_old = Validators::<T>::get();
```
ebma, line 470: At first, I thought that what you are doing here is wrong because of the weird naming. You are assigning the current validators to a variable that indicates it is the old validators. But they are not the old ones (yet). So I would rather call it `current_validators` or something.

pallets/stellar-relay/src/lib.rs
```diff
@@ -448,6 +467,30 @@ pub mod pallet {
...
+			let validators_old = validators_old
+				.into_iter()
+				.collect::<Vec<_>>();
+
+			let validator_vec_old =
+				BoundedVec::<ValidatorOf<T>, T::ValidatorLimit>::try_from(validators_old)
+					.map_err(|_| Error::<T>::BoundedVecCreationFailed)?;
+			ValidatorsOld::<T>::put(validator_vec_old);
+
+			let organizations_old = Organizations::<T>::get();
```
ebma, line 481: Same here

pallets/stellar-relay/src/tests.rs
```diff
@@ -728,6 +740,82 @@ fn update_tier_1_validator_set_works() {
...
+		let validator = Validator {
+			name: Default::default(),
+			public_key: Default::default(),
+			organization_id: organization.id,
+			public_network,
+		};
+		let new_validator_set = vec![validator; 2];
+		let new_organization_set = vec![organization; 2];
+		// let new_validator_set: Vec<ValidatorOf<Test>> = vec![validator; 2];
+		// let new_organization_set: Vec<OrganizationOf<Test>> = vec![organization; 2];
```
ebma, line 790: This can be removed I suppose

ebma (changes requested):
You have to re-do this unfortunately because you based it on the main branch. There were some changes on the dev/milestone-3 branch that remove the 'network' parameter from the struct. Sorry that I did not make this clear, I somehow assumed you would use it that. So please start from the dev/milestone-3 branch and create another PR targeted to it (and not to main).

The test looks good but is there some way we can actually check programmatically in a test if when calling the validation function from something before and after the enactment block height it changes behaviour? I.e. is there a function to set the block height in tests? And if so can you maybe use it to write another test for that scenario?

Also, the formatting is off, again :(((((((((((((((((

ebma:
Superseded by #150.

PR #150, opened 2022-12-16
Dev milestone 3 stellar relay pallet validator set 110

cr4pt0:


ebma (commented):
I added some changes. It's okay now overall but the test I was asking for on the other PR is still missing. You don't think it makes sense or it's not possible to write it the way I imagined it @RustNinja ?

cr4pt0:
> 

 definitely it is possible. i am preparing the unit tests and will commit them.

ebma:
Good job with the tests, I like it 👍

Issue #151, opened 2022-12-16
Add rate limiting


TorstenStueber:
In order to slowly roll out Pendulum and to make it more secure against attacks, implement some kind of volume limiting.

The general idea is that the total issue volume and the total redeem volume within any 24 hour time window must not exceed that limit.

### How to configure
There are two ways to implement the limit: a) as a configuration parameter, b) as a storage value.
For a) one would require a runtime upgrade to change the limit. For b) one can provide a root extrinsic function that would allow to change the stored value.

Since a) is simpler to implement, I propose that we start with a) in the first version.

### Type
The value itself should be an `Option<BalanceOf<T>>`. If the value is not set, then there is no rate limiting.

### Algorithm
The general idea is to check for every redeem or issue whether the total volume in the last 24hours reached the limit. This requires to store every redeem and issue event in the previous 24 hours and to compute an accumulated value.

Due to time and memory constraints one could simplify and approximate this as follows: define regular 24 hour intervals (e.g., every 24 * 3600 / 12 blocks) and just ensure that within any of these intervals the total volume limit is not exceeded. This requires only two variables: 
* a `counter` that keeps track of the total volume in the current interval
* an `index` that stores the interval number of the last issue/redeem happened

For a new issue/redeem: 
1. check whether the current interval is larger than `index`; if that is the case, reset `counter` to 0 and set `index` to the current interval.
2. Then decline the issue/redeem if `counter` + the volume of the issue/redeem exceeds the limit.
3. Increment `counter` by the volume of the issue/redeem

This approximation is crude as it allows that up to double the volume limit can be moved within a 24 hour window (if it happens around the beginning of a new interval). But at least the actual volume cannot be orders of magnitudes higher.

### Discussion
We might want to keep two separate limits for issue and redeem as it might be better to have loser limits on redeems (might be bad user experience if users have to wait 24 hours in order to execute redeems).

What is the implication for liquidations?

cr4pt0:
@TorstenStueber 
> an index that stores the interval number of the last issue/redeem happened

Does `index` variable means a last `block_number` for current interval?
And if the `current_block_number` will be > `index` then we reset `counter` and set new calculated `block_number` to `index` storage value?

TorstenStueber:
Define a fixed length for the interval, say `interval_length` (e.g. the number of blocks in 24 hours: 24 * 60 * 60 / 12).

Then the **interval index** is `block_number / interval_length` (integer division of course). 

[...]

PR #154, opened 2022-12-20
Feature/add rate limiting 151

cr4pt0:
Implement rate limit functionality according to issue #151


pallets/issue/src/lib.rs
```diff
@@ -126,6 +129,8 @@ pub mod pallet {
 		InvalidExecutor,
 		/// Issue amount is too small.
 		AmountBelowMinimumTransferAmount,
+		///Exceed the limit volume for issue request.
```
gonzamontiel, line 132:
```suggestion
		/// Exceeds the volume limit for an issue request.
```

pallets/issue/src/lib.rs
```diff
@@ -126,6 +129,8 @@ pub mod pallet {
...
+		ExceedLimitVolumeForIssueRequest,
```
gonzamontiel, line 133: Where is this error used btw?

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
...
+			if current_index > Some(CurrentInterval::<T>::get()) {
+				CurrentInterval::<T>::put(current_index.unwrap_or_default());
+				CurrentVolumeAmount::<T>::put(current_volume);
+			} else {
+				current_volume = CurrentVolumeAmount::<T>::get();
+			}
+			let new_issue_request = Self::convert_into_limit_currency_id_amount(amount_requested)?;
+			ensure!(
+				new_issue_request.amount().saturating_add(current_volume) > limit_volume,
+				Error::<T>::VaultNotAcceptingNewIssues
```
gonzamontiel, line 745: Should `ExceedLimitVolumeForIssueRequest` be used here? or I misunderstood ?

gonzamontiel (commented):
I left some comments

pallets/issue/src/lib.rs
```diff
@@ -147,6 +152,25 @@ pub mod pallet {
 	pub(super) type IssueMinimumTransferAmount<T: Config> =
 		StorageValue<_, BalanceOf<T>, ValueQuery>;
 
+	#[pallet::storage]
+	pub(super) type LimitVolumeAmount<T: Config> =
+		StorageValue<_, Option<BalanceOf<T>>, ValueQuery>;
+
+	///Currency id for limit rate volume value. DOT, BTC or PEN
```
ebma, line 159: I would not mention BTC here as we don't use BTC. I would change it to 
```
/// CurrencyID that represents the currency in which the volume limit is measured, eg DOT, USDC or PEN
```

pallets/issue/src/lib.rs
```diff
@@ -147,6 +152,25 @@ pub mod pallet {
...
+	pub(super) type LimitVolumeCurrencyId<T: Config> = StorageValue<_, T::CurrencyId, ValueQuery>;
+
+	#[pallet::storage]
+	pub(super) type CurrentVolumeAmount<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
+
+	/// Represent interval define regular 24 hour intervals (every 24 * 3600 / 12 blocks)
+	#[pallet::storage]
+	pub(super) type IntervalLength<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
+
+	// Represent current interval current_block_number / IntervalLength
```
ebma, line 170: I would change this to be a triple slash comment as like the others

pallets/issue/src/lib.rs
```diff
@@ -440,6 +468,8 @@ impl<T: Config> Pallet<T> {
 
 		// mint issued tokens
 		issue_amount.mint_to(&requester)?;
+		//increase volume regarding to rate limits
```
ebma, line 471: I would reword this as 
```
// increase volume according to volume limits
```

pallets/issue/src/lib.rs
cr4pt0: You are rigth!

pallets/issue/src/lib.rs
cr4pt0: done

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
...
+	fn increase_interval_volume(issue_amount: Amount<T>) -> Result<(), DispatchError> {
+		Ok(if let Some(limit_volume) = LimitVolumeAmount::<T>::get() {
+			let issue_volume = Self::convert_into_limit_currency_id_amount(issue_amount)?;
+			let current_volume = CurrentVolumeAmount::<T>::get();
+			let new_volume = current_volume
+				.checked_add(&issue_volume.amount())
+				.ok_or(ArithmeticError::Overflow)?;
+			CurrentVolumeAmount::<T>::put(new_volume);
+		})
+	}
```
ebma, line 717: I would put a simple `Ok(())` below the whole `if let` statement because it looks a bit weird like this

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
 	fn issue_minimum_transfer_amount(currency_id: CurrencyId<T>) -> Amount<T> {
 		Amount::new(IssueMinimumTransferAmount::<T>::get(), currency_id)
 	}
+
+	fn increase_interval_volume(issue_amount: Amount<T>) -> Result<(), DispatchError> {
+		Ok(if let Some(limit_volume) = LimitVolumeAmount::<T>::get() {
+			let issue_volume = Self::convert_into_limit_currency_id_amount(issue_amount)?;
+			let current_volume = CurrentVolumeAmount::<T>::get();
+			let new_volume = current_volume
+				.checked_add(&issue_volume.amount())
```
ebma, line 713: We could use a `.saturating_add()` here because that should fit perfectly for our use case

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
...
+	) -> Result<Amount<T>, DispatchError> {
+		let issue_volume =
+			oracle::Pallet::<T>::convert(&issue_amount, LimitVolumeCurrencyId::<T>::get())
+				.map_err(|_| DispatchError::Other("Missing Exchange Rate"))?;
+		Ok(issue_volume)
+	}
+
+	fn check_volume(amount_requested: Amount<T>) -> Result<(), DispatchError> {
+		let limit_volume: Option<BalanceOf<T>> = LimitVolumeAmount::<T>::get();
+		Ok(if let Some(limit_volume) = limit_volume {
```
ebma, line 730: Again, I'd add a `Ok(())` at the bottom instead of writing it like this. Should make it easier to read

ebma (changes requested):

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
...
+				CurrentInterval::<T>::put(current_index.unwrap_or_default());
+				CurrentVolumeAmount::<T>::put(current_volume);
+			} else {
+				current_volume = CurrentVolumeAmount::<T>::get();
+			}
+			let new_issue_request = Self::convert_into_limit_currency_id_amount(amount_requested)?;
+			ensure!(
+				new_issue_request.amount().saturating_add(current_volume) > limit_volume,
+				Error::<T>::ExceedLimitVolumeForIssueRequest
+			);
```
ebma, line 746: This statement needs to be flipped actually. Like this, it would mean that the error is not thrown when the amounts are above the limit_volume which is exactly the opposite of what we want.

ebma: So it should be 
```
				new_issue_request.amount().saturating_add(current_volume) <= limit_volume,
```
instead

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
...
+	}
+
+	fn check_volume(amount_requested: Amount<T>) -> Result<(), DispatchError> {
+		let limit_volume: Option<BalanceOf<T>> = LimitVolumeAmount::<T>::get();
+		Ok(if let Some(limit_volume) = limit_volume {
+			let current_block: T::BlockNumber = <frame_system::Pallet<T>>::block_number();
+			let interval_length: T::BlockNumber = IntervalLength::<T>::get();
+
+			let current_index = current_block.checked_div(&interval_length);
+			let mut current_volume = BalanceOf::<T>::default();
```
ebma, line 735: I would also change this to 
```
BalanceOf::<T>::zero();
```
to make it more explicit that it is gonna be 0. (Although It should not matter since the default value should be 0 anyways)

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
...
+			if current_index > Some(CurrentInterval::<T>::get()) {
```
ebma, line 736: This should be 
```
if current_index != Some(CurrentInterval::<T>::get()) {
```

because we want to reset the interval and volume when the current index is lower or larger.

pallets/issue/src/lib.rs
cr4pt0: Done

pallets/issue/src/lib.rs
```diff
@@ -147,6 +152,25 @@ pub mod pallet {
...
+	#[pallet::storage]
+	pub(super) type CurrentVolumeAmount<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
+
+	/// Represent interval define regular 24 hour intervals (every 24 * 3600 / 12 blocks)
+	#[pallet::storage]
+	pub(super) type IntervalLength<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
+
+	// Represent current interval current_block_number / IntervalLength
+	#[pallet::storage]
+	pub(super) type CurrentInterval<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
```
ebma, line 172: I thought about it and I would rename this storage item to "LastIntervalIndex". We want to store the index of the last interval where a issue/redeem occurred so that is more fitting I think.

pallets/issue/src/lib.rs
```diff
@@ -674,4 +704,46 @@ impl<T: Config> Pallet<T> {
...
+
+			let current_index = current_block.checked_div(&interval_length);
+			let mut current_volume = BalanceOf::<T>::default();
+			if current_index > Some(CurrentInterval::<T>::get()) {
+				CurrentInterval::<T>::put(current_index.unwrap_or_default());
+				CurrentVolumeAmount::<T>::put(current_volume);
+			} else {
+				current_volume = CurrentVolumeAmount::<T>::get();
+			}
+			let new_issue_request = Self::convert_into_limit_currency_id_amount(amount_requested)?;
```
ebma, line 742: I would rename this variable to `new_issue_request_amount`

ebma (changes requested):
So I think the overall logic should be okay once you add my requested changes. Actually the only important thing is to fix the `ensure!` statement, the rest are just nitpicky comments.

But are you planning on moving this logic to the security crate? I guess you could and you should move the functions you added here to the security crate so that they can be used both in issue and redeem. Afterwards it would be nice if you can add a test to issue/redeem that actually tests that this volume limiting works. It's okay if it is just one test but I would copy and paste it to both the issue/redeem crate.

cr4pt0:
> So I think the overall logic should be okay once you add my requested changes. Actually the only important thing is to fix the `ensure!` statement, the rest are just nitpicky comments.
> 
> But are you planning on moving this logic to the security crate? I guess you could and you should move the functions you added here to the security crate so that they can be used both in issue and redeem. Afterwards it would be nice if you can add a test to issue/redeem that actually tests that this volume limiting works. It's okay if it is just one test but I would copy and paste it to both the issue/redeem crate.

I planned to move this logic to security pallet but it requires more efforts because of dependency that security pallet will need according to new functionality.

pallets/issue/src/lib.rs
```diff
@@ -168,6 +209,10 @@ pub mod pallet {
 		fn build(&self) {
 			IssuePeriod::<T>::put(self.issue_period);
 			IssueMinimumTransferAmount::<T>::put(self.issue_minimum_transfer_amount);
+			LimitVolumeAmount::<T>::put(self.limit_volume_amount);
+			LimitVolumeCurrencyId::<T>::put(self.limit_volume_currency_id);
+			IntervalLength::<T>::put(self.interval_length);
+			LastIntervalIndex::<T>::put(self.last_interval_index);
```
ebma, line 215: The assignment to the `CurrentVolumeAmount` storage item is missing here.

pallets/issue/src/lib.rs
```diff
@@ -262,12 +309,54 @@ pub mod pallet {
...
+
+		#[pallet::weight(<T as Config>::WeightInfo::set_issue_period())]
+		#[transactional]
+		pub fn rate_limit_update(
+			origin: OriginFor<T>,
+			limit_volume_amount: Option<BalanceOf<T>>,
+			limit_volume_currency_id: T::CurrencyId,
+			current_volume_amount: BalanceOf<T>,
+			interval_length: T::BlockNumber,
+			last_interval_index: T::BlockNumber,
```
ebma, line 321: I think I would remove `current_volume_amount` and `last_interval_index` from this extrinsic actually. I don't think that being able to assign these two in this extrinsic gives us a benefit, in fact, I would even argue that having to overwrite the `current_volume_amount` with some value here is bad, because:
Let's assume we don't want to change it and leave the value that is used currently on-chain intact. Now, with the current implementation, we have to provide something for the `current_volume_amount`. We can choose the current on-chain value for that. But by the time we call and finalize this extrinsic, that value might have changed and then we reset it to some earlier value.

cr4pt0:
complete unit tests for rate limit functionality for issue pallet

pallets/issue/src/tests.rs
```diff
@@ -55,6 +57,23 @@ fn request_issue_ok_with_address(
...
+	ext::security::get_secure_id::<Test>.mock_safe(|_| MockResult::Return(get_dummy_request_id()));
+
+	ext::vault_registry::try_increase_to_be_issued_tokens::<Test>
+		.mock_safe(|_, _| MockResult::Return(Ok(())));
+	ext::vault_registry::get_stellar_public_key::<Test>
+		.mock_safe(|_| MockResult::Return(Ok(DEFAULT_STELLAR_PUBLIC_KEY)));
+
+	Issue::_request_issue(origin, amount, vault)
+}
+
```
ebma, line 76: Isn't this function just a duplicate of `request_issue_ok_with_address()`? If so, I'd remove it.

pallets/issue/src/tests.rs
```diff
@@ -179,6 +198,33 @@ fn setup_execute(
...
+	ext::stellar_relay::validate_stellar_transaction::<Test>
+		.mock_safe(move |_, _, _| MockResult::Return(Ok(())));
+	ext::currency::get_amount_from_transaction_envelope::<Test>.mock_safe(move |_, _, currency| {
+		MockResult::Return(Ok(Amount::new(amount_transferred, currency)))
+	});
+
+	Ok(issue_id)
+}
+
```
ebma, line 227: Same here. Maybe I'm just blind. If we need a separate function please explain it to me haha.

pallets/issue/src/tests.rs
```diff
@@ -434,3 +480,202 @@ fn test_set_issue_period_only_root() {
...
+#[test]
+fn test_request_issue_fails_exceed_limit_volume_for_issue_request() {
+	run_test(|| {
+		crate::Pallet::<Test>::_rate_limit_update(
+			std::option::Option::<u128>::Some(1u128),
+			DEFAULT_COLLATERAL_CURRENCY,
+			7200u64,
+		);
+
+		let issue_amount = 3;
```
ebma, line 493: I would change the `issue_amount` to `2` because it should already fail for 2.

pallets/issue/src/tests.rs
ebma: Ah wait, I think I get it now. You added a separate function to be able to get the result from it. 

But since it is basically just a duplicate I think I would prefer if we could have just one function for this. Maybe change it so that the functions you are interested in return a result and change their usages accordingly

pallets/issue/src/tests.rs
```diff
@@ -434,3 +480,202 @@ fn test_set_issue_period_only_root() {
...
+
+		//act
+		let request_issue_result = setup_execute_for_rate_limits(
+			issue_amount,
+			issue_fee,
+			griefing_collateral,
+			amount_transferred,
+		);
+
+		//assert check that limit volume resets after interval exceed\
```
ebma, line 676: typo at the end of the line

ebma (commented):
Overall the tests are good I think. Good job 👍 

But I would only merge it after you also added the tests for the redeem crate on this same PR actually

pallets/issue/src/tests.rs
cr4pt0: the same but without `unwrap`
`Issue::_request_issue(origin, amount, vault)`

pallets/issue/src/tests.rs
cr4pt0: agree!

pallets/issue/src/tests.rs
cr4pt0: remove duplicated functions

ebma:
I added some small changes and I'd say it is okay to merge now. 

But before we do that I want to point out that this implementation differs slightly from what we agreed on in #151. 
In #151 I suggested that we move the storage items for the volume limit to the security crate. @cr4pt0 mentioned some doubts that it might be best to put them in issue and redeem separately for now so that we don't have to change the security crate to also rely on the currency crate. I agreed to this change. However, this now has the implication that there are separate volume limits for issuing and redeeming tokens. It can be considered an advantage because now we have more fine-grained control over the volume limit as we can set different limits (possibly even using a different currency for measuring it). 

I think it is good how we did it now, just wanted to mention it explicitly. @TorstenStueber do you have a strong opinion on this?

TorstenStueber:
@ebma If the only difference is that there are different limits for redeem and issue, then I actually think that this is a good change. I think we should avoid a panic situation of a bank run where users want to redeem a high amount of tokens and then are blocked due to volume limits. We can't totally avoid this with volume limiting for redeems but it is good to be able to set higher limits here (maybe factor 2 or 3).

cr4pt0:
Great. Then i will merge. @TorstenStueber @ebma

PR #157, opened 2022-12-21
Feature/add rate limiting 151 redeem pallet

cr4pt0:
Add rate limit volume functionality and unit tests for redeem pallet.
#154

### DIFF
--- a/pallets/redeem/src/tests.rs
+++ b/pallets/redeem/src/tests.rs
@@ -10,8 +10,9 @@ use vault_registry::{DefaultVault, VaultStatus};
 use crate::{
 	ext,
 	mock::*,
-	types::{RedeemRequest, RedeemRequestStatus},
+	types::{BalanceOf, RedeemRequest, RedeemRequestStatus},
 };
+use sp_runtime::traits::Zero;
 
 type Event = crate::Event<Test>;
 
@@ -1079,11 +1080,9 @@ mod spec_based_tests {
 	}
 }
 
-
 #[test]
 fn test_request_redeem_fails_limits() {
 	run_test(|| {
-
 		crate::Pallet::<Test>::_rate_limit_update(
 			std::option::Option::<u128>::Some(9u128),
 			DEFAULT_COLLATERAL_CURRENCY,
@@ -1138,22 +1137,15 @@ fn test_request_redeem_fails_limits() {
 		let btc_fee = Redeem::get_current_inclusion_fee(DEFAULT_WRAPPED_CURRENCY).unwrap();
 
 		assert_err!(
-			Redeem::request_redeem(
-				RuntimeOrigin::signed(redeemer),
-				amount,
-				stellar_address,
-				VAULT
-			),
+			Redeem::request_redeem(RuntimeOrigin::signed(redeemer), amount, stellar_address, VAULT),
 			TestError::ExceedLimitVolumeForIssueRequest
 		);
 	})
 }
 
-
 #[test]
 fn test_request_redeem_limits_succeeds() {
 	run_test(|| {
-
 		crate::Pallet::<Test>::_rate_limit_update(
 			std::option::Option::<u128>::Some(91u128),
 			DEFAULT_COLLATERAL_CURRENCY,
@@ -1207,14 +1199,12 @@ fn test_request_redeem_limits_succeeds() {
 			.mock_safe(move |_| MockResult::Return(Ok(wrapped(redeem_fee))));
 		let btc_fee = Redeem::get_current_inclusion_fee(DEFAULT_WRAPPED_CURRENCY).unwrap();
 
-		assert_ok!(
-			Redeem::request_redeem(
-				RuntimeOrigin::signed(redeemer),
-				amount,
-				stellar_address,
-				VAULT
-			)
-		);
+		assert_ok!(Redeem::request_redeem(
+			RuntimeOrigin::signed(redeemer),
+			amount,
+			stellar_address,
+			VAULT
+		));
 	})
 }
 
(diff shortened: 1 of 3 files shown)


pallets/redeem/src/tests.rs
```diff
@@ -1078,3 +1079,431 @@ mod spec_based_tests {
...
+				active_replace_collateral: 0,
+				banned_until: None,
+				secure_collateral_threshold: None,
+				status: VaultStatus::Active(true),
+				liquidated_collateral: 0,
+			},
+		);
+
+		let redeemer = USER;
+		let amount = 90;
```
ebma, line 1111: Can we use `10` as the amount here? I just would prefer to use the value that is as close to the valid threshold as possible. To make it clearer we could also introduce an extra variable at the top like
```
let volume_limit: u128 = 9;
crate::Pallet::<Test>::_rate_limit_update(
			std::option::Option::<u128>::Some(volume_limit),
			DEFAULT_COLLATERAL_CURRENCY,
			7200u64,
		);
...
...
let amount =  volume_limit + 1;
```

This way it is very clear why we choose which amount.

pallets/redeem/src/tests.rs
```diff
@@ -1078,3 +1079,431 @@ mod spec_based_tests {
...
+		let redeem_fee = 5;
```
ebma, line 1173: Similar here. We can introduce a variable `volume_limit` and re-use that to set the `amount` equal to it.

pallets/redeem/src/tests.rs
```diff
@@ -1078,3 +1079,431 @@ mod spec_based_tests {
...
+			VAULT
+		));
+	})
+}
+
+#[test]
+fn test_execute_redeem_succeeds_before_exceed_rate_limit() {
+	// POSTCONDITION: `redeemTokens` MUST be called, supplying `redeemRequest.vault`,
+	// `redeemRequest.amountBtc + redeemRequest.transferFeeBtc`, `redeemRequest.premium` and
+	// `redeemRequest.redeemer` as arguments.
```
ebma, line 1211: I'd remove this comment as it is probably out of place.

pallets/redeem/src/tests.rs
```diff
@@ -1078,3 +1079,431 @@ mod spec_based_tests {
...
+			RuntimeOrigin::signed(redeemer),
+			amount,
+			stellar_address,
+			VAULT
+		));
+	})
+}
+
+#[test]
+fn test_execute_redeem_succeeds_before_exceed_rate_limit() {
```
ebma, line 1208: What is this test actually used for? Not sure I understand what it is supposed to show

pallets/redeem/src/tests.rs
```diff
@@ -1078,3 +1079,431 @@ mod spec_based_tests {
...
+			TestError::RedeemCompleted,
+		);
+	})
+}
+
+#[test]
+fn test_execute_redeem_fails_after_exceed_rate_limit() {
+	// POSTCONDITION: `redeemTokens` MUST be called, supplying `redeemRequest.vault`,
+	// `redeemRequest.amountBtc + redeemRequest.transferFeeBtc`, `redeemRequest.premium` and
+	// `redeemRequest.redeemer` as arguments.
```
ebma, line 1303: Also remove this comment

pallets/redeem/src/tests.rs
```diff
@@ -1078,3 +1079,431 @@ mod spec_based_tests {
...
+			TestError::ExceedLimitVolumeForIssueRequest
+		);
+	})
+}
+
+#[test]
+fn test_execute_redeem_after_exceed_rate_limit_reset_interval_succeeds() {
+	// POSTCONDITION: `redeemTokens` MUST be called, supplying `redeemRequest.vault`,
+	// `redeemRequest.amountBtc + redeemRequest.transferFeeBtc`, `redeemRequest.premium` and
+	// `redeemRequest.redeemer` as arguments.
```
ebma, line 1403: Also remove this

pallets/redeem/src/tests.rs
```diff
@@ -1078,3 +1079,431 @@ mod spec_based_tests {
...
+			fee: 0,
+			transfer_fee: btc_fee.amount(),
+		});
+		assert_err!(
+			Redeem::get_open_redeem_request_from_id(&H256([0u8; 32])),
+			TestError::RedeemCompleted,
+		);
+
+		let redeemer = USER;
+		let amount = 2;
```
ebma, line 1490: Can you change it to use an amount of just `1` here? You'd have to change l. 1406 as well of course but I'd prefer to be as close to the 'edge' case as possible, so showing that it errors with an amount of `1` is better I suppose.

ebma (changes requested):
I added some nitpicky comments again. The tests look good overall, I'd just like some things to be clearer.

pallets/redeem/src/tests.rs
cr4pt0: with one we will not exceed the limit rate.
```
ensure!(
				new_issue_request_amount.amount().saturating_add(current_volume) <= limit_volume,
				Error::<T>::ExceedLimitVolumeForIssueRequest
			);
```

the limit is 101 and the prev request redeem was 100.
so we need 2 here.
@ebma

pallets/redeem/src/tests.rs
cr4pt0: we check case when we try execute redeem before we reach the limit and everything works good as expected.

pallets/redeem/src/tests.rs
cr4pt0: done

PR #159, opened 2022-12-22
fix get scp, transactions archive tests. remove file at the end of test

cr4pt0:
remove downloaded archive *.xdr files after test execution.


ebma (changes requested):
The execution of [this](https://github.com/pendulum-chain/spacewalk/blob/430da0518df7b6289c3b8128337663576da2e2b8/clients/vault/src/oracle/agent.rs#L246) will also lead to two xdr files being created. @cr4pt0 can you add something to that test so that those files are removed at the end of the test?

cr4pt0:
> this

@ebma done

PR #163, opened 2022-12-26
add #[require_transactional] to all functions that require transactional

RustNinja:
add #[require_transactional] to all functions that require transactional

### DIFF
--- a/pallets/fee/src/lib.rs
+++ b/pallets/fee/src/lib.rs
@@ -10,6 +10,7 @@ extern crate mocktopus;
 use codec::EncodeLike;
 use frame_support::{
 	dispatch::{DispatchError, DispatchResult},
+	require_transactional,
 	traits::Get,
 	transactional, PalletId,
 };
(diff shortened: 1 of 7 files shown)


ebma (changes requested):
Did you test if this works with all tests and benchmarks. Because I remember that after you added the `require_transactional` to some function in the stellar-relay pallet (which made sense) the benchmarks for it did not work anymore because it is calling this internal helper function but the benchmark extrinsic is not a `transaction` itself somehow. 
I didn't look into that more deeply 2 weeks ago but just removed it again because there was not so much time but maybe you can look into it now? Just run some benchmark for a pallet by following the steps in the respective README and I guess you will see the issue I'm talking about

cr4pt0:
@ebma 
There are still errors in tests because of `require_transactional`.

ebma:
Can you fix them then or is it too complicated @cr4pt0?

erasmus:
@RustNinja what should we do with this PR here? 🤔

erasmus:
@RustNinja what should we do with this PR here? 🤔

RustNinja:
> Can you fix them then or is it too complicated @RustNinja ?

let's close it. because in other case we need rewrite tests. and there is no need to use require transaction for our functions because the is no stage updates and then error panic.
It is just safety that in future we will not write any code rust will update storage even when error dispatched from extrinsic.

Issue #160, opened 2022-12-22
Add metrics reporting to vault client


ebma:
We should make our vault client report metrics about its status and things it is doing so that we have some way of monitoring. This feature exists in interlay's implementation but was removed for now because it is not crucial right now. 

I did not look into the details of how this is done, but looking at [this](https://github.com/interlay/interbtc-clients/blob/master/vault/src/metrics.rs) file should be a good starting point.

PR #165, opened 2022-12-29
Add metrics reporting to vault client

RustNinja:
Add metrics functionality to vault client. Vault will send metrics to Prometheus.
issue #160


clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,1341 @@
...
+				.map(|i| i.symbol().to_string())
+				.unwrap_or_default(),
+			vault_id
+				.wrapped_currency()
+				.inner()
+				.map(|i| i.symbol().to_string())
+				.unwrap_or_default()
+		);
+		Self::new_with_label(label.as_ref())
+		// todo!()
```
gonzamontiel, line 172: is this comment needed?

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,1341 @@
...
 
+use crate::{
+	// execution::parachain_blocks_to_bitcoin_blocks_rounded_up,
+	system::{VaultData, VaultIdManager},
+	Error,
+};
+use async_trait::async_trait;
+// use bitcoin::{json::ListTransactionResult, GetTransactionResultDetailCategory, SignedAmount,
+// TransactionExt};
+use futures::{try_join, StreamExt, TryFutureExt};
```
gonzamontiel, line 11: apparently these imports are unused, can you remove them? And also the commented imports

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,1341 @@
...
 use lazy_static::lazy_static;
+use runtime::{
+	prometheus::{
+		gather, proto::MetricFamily, Encoder, Gauge, GaugeVec, IntCounter, IntGauge, IntGaugeVec,
+		Opts, Registry, TextEncoder,
+	},
+	types::currency_id::CurrencyIdExt,
+	CollateralBalancesPallet, CurrencyId, CurrencyInfo, Error as RuntimeError, FeedValuesEvent,
+	FixedU128, IssuePallet, IssueRequestStatus, OracleKey, RedeemPallet, RedeemRequestStatus,
+	ReplacePallet, SecurityPallet, SpacewalkParachain, SpacewalkReplaceRequest, UtilFuncs, VaultId,
```
gonzamontiel, line 21: Also here, let's try to clean the imports (ReplacePallet not used, etc)

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,1341 @@
...
+	expired_count: Gauge,
+}
+
+#[derive(Clone, Debug)]
+pub struct PerCurrencyMetrics {
+	locked_collateral: Gauge,
+	collateralization: Gauge,
+	required_collateral: Gauge,
+	remaining_time_to_redeem_hours: Gauge,
+	btc_balance: XLMBalance,
```
gonzamontiel, line 136: this should be renamed to `lumen_balance` or `xlm_balance`

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,1341 @@
...
+		}
+	}
+}
+
+async fn publish_time_to_first_deadline<V: VaultDataReader, P: RedeemPallet + SecurityPallet>(
+	parachain_rpc: &P,
+	vault_id_manager: &V,
+	redeems: &[(H256, SpacewalkReplaceRequest)],
+) {
+	todo!()
```
gonzamontiel, line 514: Is this intended? So this part will be completed in another PR? or is this still a draft PR?

gonzamontiel (dismissed):
Thanks for this, very useful indeed. I left some minor comments and also a question, was this ready for review? I saw some todo's left.

cr4pt0:
@gonzamontiel 
YEs. It is ready for review because this pr cover everything related to metrics from parachain.

The next one will be the pr with metrics for stellar network by vault id.

clients/vault/src/metrics.rs
cr4pt0: so this function related to xlm network. so will add/modify logic in next pr.

ebma (changes requested):
I checked it out and it works overall. I re-added a task that periodically publishes the metrics and changed the default port to 9616 because when running a testchain node locally, the prometheus instance of that node already uses that port.

However, there are lots of warning about unused things that should be cleaned up, and also @gonzamontiel s comments still have to be addressed. 
Furthermore, I think it would be nice if you included the other metrics that you wanted to add for stellar on this PR already, or is there a good reason not to? Having these metrics is not critical for the audit and I would prefer merging a clean version to dev/milestone-3.

PS: When running the vault client one can query the metrics in browser from `http://127.0.0.1:9616/metrics/graph`. Maybe we should point that out in some readme.

clients/vault/src/metrics.rs
cr4pt0: Done.

clients/vault/src/metrics.rs
cr4pt0: Done

clients/vault/src/metrics.rs
cr4pt0: Done

ebma:
@cr4pt0 what about this PR by the way? I'm pinging you again because it would actually be nice to have metrics for our vault once we launch on our main networks.

cr4pt0:
> @cr4pt0 what about this PR by the way? I'm pinging you again because it would actually be nice to have metrics for our vault once we launch on our main networks.

@ebma 
sure. i finished the implementation. but you are right i need to remove warnings.
will do after we finish with chain extension and foucoco upgrade runtime.

ebma:
Can you please also address the other issues I mention in my last comment? 😬

>  [...] Also, there are plenty of occurrences of BTC and Bitcoin which should be either renamed or removed. The test module included in metrics.rs also either has to be changed or removed; it is of no use to us the way it is now.

RustNinja:
> Can you please also address the other issues I mention in my last comment? 😬
> 
> > [...] Also, there are plenty of occurrences of BTC and Bitcoin which should be either renamed or removed. The test module included in metrics.rs also either has to be changed or removed; it is of no use to us the way it is now.

@ebma 
Done

clients/runtime/src/rpc.rs
```diff
@@ -456,6 +470,9 @@ pub trait UtilFuncs {
 	/// Gets the current height of the parachain
 	async fn get_current_chain_height(&self) -> Result<u32, Error>;
 
+	/// Gets the ID of the native currency.
+	fn get_native_currency_id(&self) -> CurrencyId;
```
ebma, line 474: Wouldn't we also need a utility function for the relay chain currency ID?

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,522 @@
...
+		Opts::new("redeem_count", "Number of redeems"),
+		&[CURRENCY_LABEL, REQUEST_STATUS_LABEL]
+	)
+	.expect("Failed to create prometheus metric");
+	pub static ref NATIVE_CURRENCY_BALANCE: Gauge =
+		Gauge::new("native_currency_balance", "Native Currency Balance")
+			.expect("Failed to create prometheus metric");
+	pub static ref FEE_BUDGET_SURPLUS: GaugeVec =
+		GaugeVec::new(Opts::new("fee_budget_surplus", "Fee Budget Surplus"), &[CURRENCY_LABEL])
+			.expect("Failed to create prometheus metric");
```
ebma, line 85: I think we can remove this one as we will not use it

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,522 @@
...
 	pub static ref RESTART_COUNT: IntCounter =
 		IntCounter::new("restart_count", "Number of service restarts")
 			.expect("Failed to create prometheus metric");
 }
+const STELLAR_NATIVE_ASSET_TYPE: [u8; 6] = [110, 97, 116, 105, 118, 101]; //"native"
```
ebma, line 90: Can't we just write `... = *b"native"` ?

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,522 @@
...
+#[async_trait]
+impl VaultDataReader for VaultIdManager {
+	async fn get_entries(&self) -> Vec<VaultData> {
+		self.get_entries().await
+	}
+}
+
+pub const fn one() -> f64 {
+	10u128.pow(12u32) as f64
+}
```
ebma, line 130: I think this function is quite useless and only used at the top. Maybe we could use the `one()` method that exists on our assets on the `CurrencyId` enum already.

clients/runtime/src/types.rs
```diff
@@ -114,6 +115,24 @@ mod metadata_aliases {
...
+
+	pub trait CurrencyIdExt {
+		fn inner(&self) -> Result<u8, Error>;
+	}
+
+	impl CurrencyIdExt for CurrencyId {
+		fn inner(&self) -> Result<u8, Error> {
+			match self {
+				CurrencyId::XCM(x) => Ok(*x),
+				_ => Err(Error::CurrencyNotFound),
```
ebma, line 130: This needs to implement something for our `CurrencyId::Stellar` types as well.

ebma: We should consider not returning an `u8` as a result here but a string or a Vec<> maybe. u8 is not enough to provide a meaningful description.

ebma: Note that this is used to describe the currency in our metric reporting so it should be unambiguous

clients/vault/src/metrics.rs
```diff
@@ -1,27 +1,522 @@
...
+				};
+			} else {
+				tracing::warn!("Incorrect stellar asset type");
+			}
+			vault.metrics.asset_balance.actual.set(wrapped_to_collateral as f64);
+		},
+		Err(e) => {
+			// unexpected error, but not critical so just continue
+			tracing::warn!("Failed to get balance: {}", e);
+			vault.metrics.asset_balance.actual.set(0 as f64);
```
ebma, line 313: Something has to be wrong with the current setup, because when I run it, I see that the Stellar balance shown in metrics is set to 0.

ebma (changes requested):
There seems to be something wrong with the fetching of the Stellar balance. Also, the inner() function for the CurrencyId is not working correctly, we need to extend that. 

For the Stellar balance logic, maybe consider writing test cases that show that some functions do work as expected. Also, please test the vault locally by running it and opening 'http://127.0.0.1:9616/metrics' in the browser. You can check in the metrics here if the Stellar balance problem was solved or not.

clients/vault/src/metrics.rs
ebma: [It does exist.](https://github.com/pendulum-chain/spacewalk/blob/c81caf406ce63b7dffbe8443fcff6e29e77ca96d/primitives/src/lib.rs#L478) I added it recently.

clients/runtime/src/rpc.rs
ebma: I checked interlays code and they also don't use it. Fine if we don't add it then I guess.

clients/vault/src/metrics.rs
RustNinja: Oh. Got it. Thanks. Did not see it previous time! Will use your code!

clients/runtime/src/types.rs
RustNinja: @ebma is it ok to use debug formatter for CurrencyId?

clients/vault/src/metrics.rs
RustNinja: Done.

clients/vault/src/metrics.rs
RustNinja: Done.

clients/vault/src/metrics.rs
RustNinja: Done.

clients/runtime/src/types.rs
```diff
@@ -114,6 +115,34 @@ mod metadata_aliases {
...
+			match self {
+				CurrencyId::Native => Ok("Native".to_owned()),
+				CurrencyId::XCM(foreign_currency_id) =>
+					Ok(format!("XCM {}", foreign_currency_id).to_owned()),
+				CurrencyId::Stellar(stellar_asset) => match stellar_asset {
+					Asset::StellarNative => Ok("Stellar Native".to_owned()),
+					Asset::AlphaNum4 { code, issuer } =>
+						Ok(format!("Stellar asset <{:?}>:<{:?}>", code, issuer)),
+					Asset::AlphaNum12 { code, issuer } =>
+						Ok(format!("Stellar asset <{:?}>:<{:?}>", code, issuer)),
```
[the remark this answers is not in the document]
RustNinja: Done.

clients/vault/src/metrics.rs
RustNinja: i found the problem.
there is no rpc method` oracle_wrappedToCollateral `
we did not expose any RPC for oracle pallet API
do we need to expose it? what do you think?
@ebma

RustNinja: Now the error reproduced on test `spacewalk-standalone`
but the reason seems that there is no price feed for currency id
`AssetTypeCreditAlphanum4(AlphaNum4 { asset_code: [85, 83, 68, 67], issuer: PublicKeyTypeEd25519([20, 209, 150, 49, 176, 55, 23, 217, 171, 154, 54, 110, 16, 50, 30, 226, 102, 231, 46, 199, 108, 171, 97, 144, 240, 161, 51, 109, 72, 34, 159, 139]) })`
and error is 
`Err(Rpc(ClientError(Call(Custom(ErrorObject { code: InternalError, message: "Execution error: Module(ModuleError { index: 19, error: [0, 0, 0, 0], message: None })", data: None })))))`

ebma: But we do have the price for USD so the problem can't actually be that there is no price, or am I wrong? Is some conversion buggy?

clients/runtime/src/types.rs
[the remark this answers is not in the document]
ebma: Thanks for changing it. But I checked and saw that this now of course shows the actual bytes and no nice human-readable asset code and issuer. Can you change it so that code and issuer are shown as an ascii string instead of the bytes?

clients/vault/src/metrics.rs
RustNinja: @ebma 
Parachain status is offline. So error means that there is no fresh price feed for 
USDC
It is required currency id to be online for all parachain 
AssetTypeCreditAlphanum4(AlphaNum4 { asset_code: [85, 83, 68, 67], issuer: PublicKeyTypeEd25519([20, 209, 150, 49, 176, 55, 23, 217, 171, 154, 54, 110, 16, 50, 30, 226, 102, 231, 46, 199, 108, 171, 97, 144, 240, 161, 51, 109, 72, 34, 159, 139]) })

ebma: Ah yeah it's probably because there is an issue with the testchain. I'll create a new ticket. But for now, if you are testing locally you can quick-fix the testchain by changing the XCMCurrencyConversion in `testchain/runtime/src/lib.rs` to this

```
impl XCMCurrencyConversion for SpacewalkNativeCurrencyKey {
	fn convert_to_dia_currency_id(token_symbol: u8) -> Option<(Vec<u8>, Vec<u8>)> {
		if token_symbol == 0 {
			return Some((b"Kusama".to_vec(), b"KSM".to_vec()))
		}

		// We assume that the blockchain is always 0 and the symbol represents the token symbol
		let blockchain = vec![0u8];
		let symbol = vec![token_symbol];

		Some((blockchain, symbol))
	}

	fn convert_from_dia_currency_id(blockchain: Vec<u8>, symbol: Vec<u8>) -> Option<u8> {
		// We assume that the blockchain is always 0 and the symbol represents the token symbol
		if blockchain.len() != 1 && blockchain[0] != 0 || symbol.len() != 1 {
			return None
		}
		return Some(symbol[0])
	}
}
```

clients/runtime/src/types.rs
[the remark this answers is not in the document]
RustNinja: Done.

clients/vault/src/metrics.rs
RustNinja: @ebma 
I tested to bridge tokens with Solar Wallet
FCC6AD0A-E155-4D22-8ECB-7D8DF65DFAB2

and after issuing the token on standalone side it shows
```
stellar_balance{currency="XCM(0)_\"USDC\":\"GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC\"",type="actual"} 112000000
stellar_balance{currency="XCM(0)_\"USDC\":\"GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC\"",type="required_lowerbound"} 120000000
stellar_balance{currency="XCM(0)_\"USDC\":\"GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC\"",type="required_upperbound"} 120000000
```

ebma (approved):
I fixed the remaining problems. This is now ready for merge once CI passes.

RustNinja:
> I fixed the remaining problems. This is now ready for merge once CI passes.

Thanks 🙏
 Wow. Amazing 🚀

PR #167, opened 2023-01-02
update redeem/issue weights according to rate limit functionality

cr4pt0:
update weights for issue and redeem pallet after implementation of rate limits #151


ebma (approved):
Looks good to me 👍

PR #168, opened 2023-01-02
Read storage after ensure signed in nomination pallet

cr4pt0:
Read storage after ensure signed nomination pallet
Switch reading from storage and ensuring signed.


ebma (approved):
Looks good, nice catch 👍

PR #170, opened 2023-01-03
Set set_active_block_number function for testing purpose unavailable for runtime

cr4pt0:
Set `set_active_block_number` function for testing purpose unavailable for runtime

### DIFF
--- a/pallets/security/src/lib.rs
+++ b/pallets/security/src/lib.rs
@@ -308,9 +308,7 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// for testing purposes only!
-	/// Std is never enabled in the Runtime.
-	/// This function can only be called from testing but not from the pallet itself:
-	#[cfg(feature = "std")]
+	#[cfg(feature = "testing-utils")]
 	pub fn set_active_block_number(n: T::BlockNumber) {
 		ActiveBlockCount::<T>::set(n);
 	}
--- a/pallets/security/src/lib.rs
+++ b/pallets/security/src/lib.rs
@@ -308,6 +308,9 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// for testing purposes only!
+	/// Std is never enabled in the Runtime.
+	/// This function can only be called from testing but not from the pallet itself:
+	#[cfg(feature = "std")]
 	pub fn set_active_block_number(n: T::BlockNumber) {
 		ActiveBlockCount::<T>::set(n);
 	}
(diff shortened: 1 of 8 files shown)


pallets/security/src/lib.rs
```diff
@@ -308,6 +308,9 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// for testing purposes only!
+	/// Std is never enabled in the Runtime.
+	/// This function can only be called from testing but not from the pallet itself:
+	#[cfg(feature = "std")]
```
ebma, line 313: Can we use `#[cfg(test)]` instead or does this not work for some reason?

cr4pt0: I tried. No. We can't.

cr4pt0: It does not work because when you use `#[cfg(test)]` this code will not include to binary and other packages can't reach this function.

cr4pt0: you can restrict it to testing, which only works from the **same crate** since the
'test' feature **does not propagate** to dependencies:
```
#[cfg(test)]
pub fn some_fn(n: T::BlockNumber) {
    SomeStorage::<T>::set(n);
}
```

but other packages use this function. so we can't use `#[cfg(test)]` in current case
@ebma

ebma: Okay got it. But can we use the 'testing-utils' feature for this then, like we do in other crates, eg stellar-relay? In stellar-relay some helper functions that are also used in other crates [are exposed](https://github.com/pendulum-chain/spacewalk/blob/debc953b78986c011ca55b265c808e97db61655b/pallets/stellar-relay/src/lib.rs#L19-L20) when the 'testing-utils' feature is imported, maybe we can do something similar here. 

I am sure your approach with `std` also works, but it seems a bit hacky and I think we can do better by using another feature.

cr4pt0: Wow! Amazing! You are right. 
Thank you so much. I changed it. 
use `#[cfg(feature = "testing-utils")]` instead of `#[cfg(feature = "std")]`
Ready for merge.
@ebma

ebma:
Thanks for changing it @cr4pt0, looks great now. Can you check if compiling the node with 
```
cargo build --package spacewalk-standalone --release --features runtime-benchmarks
```
still works? Just to be sure that there are no conflicts with dependencies somehow. I'll approve it once you tested that

cr4pt0:
> spacewalk-standalone

@ebma you are right. 
Fixed. Now it compiling the node with
`cargo build --package spacewalk-standalone --release --features runtime-benchmarks`

ebma:
Okay great. From my experience we should always run 
```
cargo build --package spacewalk-standalone --release --features runtime-benchmarks
```
after changing something about the pallets. This seems to catch many issues if there are any. I'll approve it then.

ebma (approved):

PR #173, opened 2023-01-05
remove unused metadata scale file

cr4pt0:


ebma (commented):
I think we could actually also remove the `metadata-parachain-testnet.scale` file while we're at it.

ebma:
> I think we could actually also remove the metadata-parachain-testnet.scale file while we're at it.

@cr4pt0 can you do this so we can merge?

cr4pt0:
@ebma Done

PR #174, opened 2023-01-05
Fix wrong issue amount conversion

ebma:
During manual testing I realized two things: 
1) when the client is shutdown and re-started again, pending redeem and replace requests were picked up and executed successfully but 'old' issue requests were not 
and 2) the `BalanceConversion` struct we introduced some time ago makes it hard/confusing to fulfill an issue request with the appropriate stroop amount on Stellar. I changed the rate used in the `BalanceConversion` to be `1` so it does not actually do anything anymore. If I understand correctly this struct was intended to convert between the native on-chain balance and Stellar balances. That's why I decided to keep it around and added a comment.

This PR fixes both issues.


ebma:
@cr4pt0 can you review this PR?

cr4pt0 (approved):

cr4pt0:
> @cr4pt0 can you review this PR?

Sure. Done

PR #181, opened 2023-01-09
Prevent replay attacks by checking memo hash

ebma:
This pull request changes the issue, redeem and replace pallet such that they check if the to-be-used Stellar transaction includes the respective request ID as a memo. This is crucial for mitigating replay attacks.


pallets/stellar-relay/Cargo.toml
```diff
@@ -24,6 +24,7 @@ frame-system = { default-features = false, git = "https://github.com/paritytech/
 frame-benchmarking = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.36", optional = true }
 
 sp-std = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.36" }
+sp-core = { default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.36" }
```
b-yap, line 27: I just noticed; if this is only added just so we could use the `H256` datatype, `spacewalk-primitives` is actually available: 
https://github.com/pendulum-chain/spacewalk/blob/dev/milestone-3/primitives/src/lib.rs#L9

The `H256` is exposed already.

I think we can actually remove `sp-core` altogether.

ebma: Hmm yes you're right, that's the only reason I added sp-core. But I just checked the other pallets that use H256 and from what I can tell they all import it from sp_core instead of the primitives. Do you think we should change it to use the one from primitives?

b-yap: ^ that would be ideal, imo.  One less dependency in the toml files. :eyes:

ebma: Haha, true there would be one less dependency, but maybe we should try to depend on the spacewalk primitives as little as possible in the crates? Not so sure about it. Anyways, it does not really matter at this point so I'll change it to use the one from primitives.

ebma:
@b-yap, the rest is okay and I'm good to merge the PR?

pallets/stellar-relay/Cargo.toml
ebma: Nevermind, after changing the code to use the one from primitives I realized that the primitives dependency is declared as 'optional' and only included when the 'testing-utils' feature is set...... so I would say we keep it like this 😅

cr4pt0 (approved):

b-yap:
lgtm :thumbsup:

Issue #179, opened 2023-01-09
Change or remove comments and variables about 'bitcoin' and 'btc'


ebma:
At the moment, we have lots of doc comments and some variables that are still referring to Bitcoin or BTC. We should change these comments and variables such that they relate to spacewalk and not Bitcoin/interBTC. This means that simply 'find-replacing' every mention of the word 'Bitcoin' with 'Stellar' might not be feasible because this would not always make sense. We need to check case by case how to reformulate the comment so that it makes sense in our context.

PR #183, opened 2023-01-09
Use Stellar, XLM naming instead of BTC, Bitcoin

cr4pt0:
update comments, extrinsic, unit tests and regenerate metadata.scale file
#179


ebma:
I added some changes because it was easier/faster to do these little things myself instead of pointing them out everywhere.

ebma (approved):

PR #184, opened 2023-01-10
Restrict set_exchange_rate oracle function. Allow only for tests and benchmarks.

cr4pt0:


pallets/oracle/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 //! # Oracle Pallet
 //! Based on the [specification](https://spec.interlay.io/spec/oracle.html).
 
-#![deny(warnings)]
+#![warn(warnings)]
```
ebma, line 4: Is this change necessary or did you forget to change it back?

ebma (approved):
Looks good to me overall. I checked and the tests are all still working. Goob job 👍 

We can merge this once you address my other comment.

pallets/oracle/src/lib.rs
cr4pt0: @ebma 
I did it because the test failed because of `unused import: `DispatchResult``

Should I fix it or we can leave  it like this?

ebma: Can't we just remove the unused import then?

cr4pt0: No. Because `DispatchResult` for other code when there is no feature `testing-utils`

ebma: I fixed it by conditionally importing it.

PR #186, opened 2023-01-11
update orml dependencies to v0.9.36

b-yap:
for https://github.com/pendulum-chain/pendulum/pull/89, to avoid compile errors with the orml dependencies


ebma (changes requested):
Makes sense to change it 👍 This means we might be able to get rid of our `patch` statements in the root Cargo.toml. @b-yap can you remove them and check if everything still works?

cr4pt0 (approved):

b-yap:
@ebma i've removed the patch

ebma (approved):

PR #187, opened 2023-01-11
Redesign spacewalk oracle to be compatible with any oracle provider.

RustNinja:
- add dia.rs
- add oracle_mock.rs functionality to provide easy way to mock test runtime.
- remove feed_values, register, unregister_vault extrinsics from spacewalk oracle.
- remove old Storage from spacewalk oracle (RawUpdatedValue, UpdatedValue, AggregatedValues)
- introduce DiaOracleAdapter struct
- implement `DataProviderExtended` for `DiaOracleAdapter`
- add functionality to retrieve data from DIA oracle with Spacewalk OracleKey
- update all tests and mock.rs in other spacewalk pallets (redeem, replace, issue, vault-registry)
- update all benchmarking tests
- integrate DIA-oracle and new spacewalk oracle pallet to testchain runtime

### DIFF
--- a/clients/runtime/src/tests.rs
+++ b/clients/runtime/src/tests.rs
@@ -51,9 +51,6 @@ async fn test_getters() {
 		async {
 			assert_eq!(parachain_rpc.get_parachain_status().await.unwrap(), StatusCode::Error);
 		},
-		// async {
-		// 	assert!(parachain_rpc.get_replace_dust_amount().await.unwrap() > 0);
-		// },
 		async {
 			assert!(parachain_rpc.get_current_active_block_number().await.unwrap() == 0);
 		}
--- a/clients/runtime/src/types.rs
+++ b/clients/runtime/src/types.rs
@@ -73,7 +73,6 @@ mod metadata_aliases {
 			Balance,
 			CurrencyId,
 		>;
-	// pub use crate::metadata::runtime_types::spacewalk_primitives::CurrencyId;
 
 	pub type SpacewalkIssueRequest =
 		metadata::runtime_types::spacewalk_primitives::issue::IssueRequest<
--- a/pallets/oracle/src/tests.rs
+++ b/pallets/oracle/src/tests.rs
@@ -14,7 +14,6 @@ fn feed_values_succeeds() {
 		let key = OracleKey::ExchangeRate(CurrencyId::XCM(DOT));
 		let rate = FixedU128::checked_from_rational(100, 1).unwrap();
 
-		// Oracle::is_authorized.mock_safe(|_| MockResult::Return(true));
 		let result = Oracle::_feed_values(3, vec![(key.clone(), rate)]);
 		assert_ok!(result);
 
--- a/pallets/staking/src/tests.rs
+++ b/pallets/staking/src/tests.rs
@@ -4,8 +4,6 @@ use primitives::ForeignCurrencyId;
 /// Tests for Staking
 use crate::mock::*;
 
-// type Event = crate::Event<Test>;
-
 #[macro_export]
 macro_rules! fixed {
 	($amount:expr) => {


pallets/oracle/Cargo.toml
```diff
@@ -20,6 +20,9 @@ frame-system = { git = "https://github.com/paritytech/substrate", branch = "polk
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.36", default-features = false, optional = true }
 pallet-timestamp = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.36", default-features = false }
 
+orml-oracle = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.33", default-features = false }
```
ebma, line 23: We updated orml dependencies to `0.9.36` on the dev/milestone-3 branch yesterday. So you should merge the dev branch into this one and change the branch here to `polkadot-v0.9.36`.

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,75 @@
+use dia_oracle::DiaOracle;
+use orml_oracle::{DataProviderExtended, TimestampedValue};
+pub use primitives::{oracle::Key as OracleKey, CurrencyId, TruncateFixedPointToInt};
+use sp_std::marker;
+
+use sp_runtime::traits::Convert;
+
+struct X;
```
ebma, line 8: Why `X`?

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+pub struct DiaOracleAdapter<
+	DiaPallet: DiaOracle,
+	UnsignedFixedPoint,
+	Moment,
+	ConvertKey,
+	ConvertPrice,
+	ConvertMoment,
```
ebma, line 28: If we need all three I would group them under a new trait and name it something like eg. `TypeConversion`.

```
trait TypeConversion<TargetKey, TargetPrice, TargetMoment> {
   type ConvertKey: Convert<TargetKey, Option<(Vec<u8>, Vec<u8>)>> + Convert<(Vec<u8>, Vec<u8>), Option<TargetKey>>,
   type ConvertPrice:Convert<TargetPrice, Option<u128>> + Convert<u128, Option<TargetPrice>>,
   type ConvertMoment: Convert<TargetMoment, Option<u64>> + Convert<u64, Option<TargetMoment>>,
>
```

then in the `where` statement of the `impl` block you just specify the target types. The source types should always be the same and we don't need to make them generic since they are set by DIA.

cr4pt0: Thanks. Good idea! 
I just take as example [fungibles_adapter.rs](https://github.com/paritytech/polkadot/blob/master/xcm/xcm-builder/src/fungibles_adapter.rs#L63-L72) from xcm-builder

There are two different type without combination in one trait.
```
ConvertAssetId: Convert<MultiLocation, AssetId>,
ConvertBalance: Convert<u128, Balance>,
```

Do we need introduce one more trait `TypeConversion` to group them together?
@ebma

pallets/oracle/src/dia.rs
cr4pt0: Mock. Will remove it!

pallets/oracle/Cargo.toml
cr4pt0: Thanks

pallets/oracle/src/dia.rs
ebma: I think it would be better to group it into another trait since we don't have just two but three. But that's just my personal opinion. Would you prefer to keep the three conversions separated?

cr4pt0: For now i would like to keep them separated. Try to mock in mock.rs and then complicate and introduce type to combine them together.
Because now there is a problem more complicated: `fn _set_exchange_rate`
https://github.com/pendulum-chain/spacewalk/blob/dev/milestone-3/pallets/oracle/src/lib.rs#L392-L401
Used by other pallet and benchmarking..
What is the best way to solve and replace it?
Because with new oracle-extension implementation there is no way to update storage...

ebma: Hmm, that's problematic indeed. How about we make our `DiaOracleAdapter` not only implement the DataProviderExtended trait but also the [DataFeeder](https://docs.rs/orml-traits/0.4.0/orml_traits/data_provider/trait.DataFeeder.html) trait? Then we can change this `_set_exchange_rate()` function to call the [feed_value](https://docs.rs/orml-traits/0.4.0/orml_traits/data_provider/trait.DataFeeder.html#tymethod.feed_value) function.

cr4pt0: yes. we can. but we will not use datafeed trait in prodaction. only for testing. and this trait bound will be redundant for prodaction runtime.

ebma: How about we only require the additional `DataFeeder` trait bound when the `testing-utils` feature is set on the `oracle` crate then? The `_set_exchange_rate` function is also only available if that feature is set

pallets/fee/Cargo.toml
```diff
@@ -33,6 +33,7 @@ mocktopus = "0.8.0"
 frame-benchmarking = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.36", default-features = false }
 pallet-timestamp = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.36", default-features = false }
 currency = { path = "../currency", default-features = false, features = ["testing-constants"] }
+security = { path = "../security", features = ['testing-utils'] }
```
ebma, line 36: Why is this necessary?

pallets/issue/Cargo.toml
```diff
@@ -55,6 +55,8 @@ oracle = { path = "../oracle", features = ['testing-utils'] }
 orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.33" }
 orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.33" }
 
+dia-oracle = { git = "https://github.com/TorstenStueber/oracle-pallet", branch = "master", default-features = false }
```
ebma, line 58: (As we learned in todays sync meeting) this can be changed to the 'real' repo now.

pallets/oracle/src/lib.rs
```diff
@@ -159,25 +147,25 @@ pub mod pallet {
 	pub struct GenesisConfig<T: Config> {
 		pub max_delay: u32,
 		pub authorized_oracles: Vec<(T::AccountId, Vec<u8>)>,
```
ebma, line 149: We can remove the authorized oracles from the genesis config since we don't need them anymore (and they are not used).

pallets/oracle/src/lib.rs
```diff
@@ -110,39 +122,15 @@ pub mod pallet {
...
-	/// Time until which the aggregate is valid
-	#[pallet::storage]
-	pub type ValidUntil<T: Config> = StorageMap<_, Blake2_128Concat, OracleKey, T::Moment>;
-
 	/// Maximum delay (milliseconds) for a reported value to be used
 	#[pallet::storage]
 	#[pallet::getter(fn max_delay)]
 	pub type MaxDelay<T: Config> = StorageValue<_, T::Moment, ValueQuery>;
 
 	// Oracles allowed to set the exchange rate, maps to the name
```
ebma, line 130: This comment is wrong now. Please update it accordingly.

pallets/oracle/src/lib.rs
```diff
@@ -251,26 +220,29 @@ pub mod pallet {
 impl<T: Config> Pallet<T> {
 	// public only for testing purposes
```
ebma, line 221: This comment is wrong I think.

pallets/oracle/src/lib.rs
```diff
@@ -251,26 +220,29 @@ pub mod pallet {
...
+			let is_outdated = current_time > price.timestamp + max_delay;
+			if !is_outdated {
+				updated_items.push((key.clone(), Some(price.value)));
 			}
 		}
-
+		let updated_items_len = updated_items.len();
 		if !updated_items.is_empty() {
 			Self::deposit_event(Event::<T>::AggregateUpdated { values: updated_items });
 		}
```
ebma, line 242: Doing it this way makes it different from the original implementation.
In the 'original' implementation, the entries of `updated_items` can be set to `None` or `Some(price)`. But since you exit the for loop in l. 232 if the price is empty, the `updated_items` cannot contain values of type `None`. 

I think we should change it so that a value of `None` is pushed to `updated_items` if no price was found.

pallets/oracle/src/lib.rs
```diff
@@ -251,26 +220,29 @@ pub mod pallet {
...
 
 		let current_status_is_online = Self::is_oracle_online();
-		let new_status_is_online = !raw_values_updated.is_empty() &&
-			raw_values_updated.iter().all(|(key, _)| Aggregate::<T>::get(key).is_some());
+		let new_status_is_online = oracle_keys.len() > 0 && updated_items_len == oracle_keys.len();
```
ebma, line 245: Adding the change in my previous comment implies that we also need to change this logic. If we allow the 'push' of `None` values to the updated_items vec, we need to check here if all items actually have a value of `Some(x)` before we set the status back to online.

pallets/oracle/src/lib.rs
```diff
@@ -281,16 +253,26 @@ impl<T: Config> Pallet<T> {
...
 	// public only for testing purposes
-	pub fn _feed_values(oracle: T::AccountId, values: Vec<(OracleKey, T::UnsignedFixedPoint)>) {
-		for (key, value) in values.iter() {
-			let timestamped =
-				TimestampedValue { timestamp: Self::get_current_time(), value: *value };
-			RawValues::<T>::insert(key, &oracle, timestamped);
-			RawValuesUpdated::<T>::insert(key, true);
+	#[cfg(feature = "testing-utils")]
+	pub fn _feed_values(
+		oracle: T::AccountId,
```
ebma, line 260: Since the introduction of this `oracle` parameter to this function results in lots of (actually unnecessary) changes to the existing code, since all the calls to `_feed_values` have to be adjusted, I would suggest we remove this parameter again and in l. 267, when calling 
```
T::DataFeedProvider::feed_value(oracle.clone(), k.clone(), timestamped)
				.expect("Expect store value by key");
```
we just use some hardcoded account ID. Since this is only for testing purposes I don't see a problem with always using the same account for it. We can just do something like
```
let feeder = AccoundId::from([1u8;32])  
T::DataFeedProvider::feed_value(feeder, k.clone(), timestamped)
				.expect("Expect store value by key");
```

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,207 @@
...
+	}
+}
+pub struct MockConvertPrice;
+impl Convert<u128, Option<FixedU128>> for MockConvertPrice {
+	fn convert(price: u128) -> Option<FixedU128> {
+		Some(FixedU128::from_inner(price))
+	}
+}
+
+pub struct MockMoment;
```
ebma, line 202: I'd prefer a different name for this. Maybe similar to the others `MockConvertMoment`

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,207 @@
...
+		Some(TimestampedValue { value, timestamp })
+	}
+
+	///do not need this funtion implementation
+	fn get_all_values() -> Vec<(OracleKey, Option<TimestampedValue<UnsignedFixedPoint, Moment>>)> {
+		panic!("spacewalk oracle extension does not requre implementation of DataProviderExtended get_all_values function")
+	}
+}
+
+pub struct MockDiaOracleConvertor;
```
ebma, line 123: This should maybe be called `MockOracleKeyConvertor` since that's what it's doing.

pallets/oracle/src/oracle_mock.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[derive(Clone)]
+struct Data {
+	pub key: (Vec<u8>, Vec<u8>),
+	pub price: u128,
+	pub timestamp: u64,
+}
+
+thread_local! {
+	static COINS: RefCell<Vec<Data>> = RefCell::new(vec![]);
```
ebma, line 22: This is a bit hacky. But if it works, fine for me since it is only used for testing purposes.

ebma (changes requested):
I think the approach is okay overall. I added my thoughts on the most important parts.

Also, compilation with all features enabled does not work. Please make sure that `cargo test --all --all-features` works.

pallets/fee/Cargo.toml
cr4pt0: Because of this
`Security::set_active_block_number(1);`

and declaration is like this
```
/// for testing purposes only!
    pub fn set_active_block_number(n: T::BlockNumber) {
	ActiveBlockCount::<T>::set(n);
}
```

pallets/oracle/src/lib.rs
cr4pt0: Done

pallets/oracle/src/lib.rs
cr4pt0: Done

pallets/oracle/src/lib.rs
cr4pt0: Done

pallets/issue/Cargo.toml
cr4pt0: Done

pallets/oracle/src/dia.rs
cr4pt0: Done

pallets/oracle/src/dia.rs
cr4pt0: Done.

pallets/oracle/src/lib.rs
cr4pt0: Thanks. Good suggestion. 
I tried this approach to mock account id. for this testing function.
But it does not work

```
let feeder = T::AccountId::from([1u8; 32]);
    |                      ------------------ ^^^^^^^^^ expected associated type, found array `[u8; 32]`
    |                      |
    |                      arguments to this function are incorrect
    |
    = note: expected associated type `<T as frame_system::Config>::AccountId`
                         found array `[u8; 32]`
```

But this is really works. Thanks.

```
use primitives::AccountId;
let feeder = AccountId::from([1u8; 32]);
```

pallets/oracle/Cargo.toml
cr4pt0: Done

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,121 @@
...
+where
+	Dia: DiaOracle,
+	ConvertKey: Convert<OracleKey, Option<(Vec<u8>, Vec<u8>)>>
+		+ Convert<(Vec<u8>, Vec<u8>), Option<OracleKey>>,
+	ConvertPrice: Convert<u128, Option<UnsignedFixedPoint>>,
+	ConvertMoment: Convert<u64, Option<Moment>>,
+{
+	fn get_no_op(key: &OracleKey) -> Option<TimestampedValue<UnsignedFixedPoint, Moment>> {
+		let dia_key: Option<(Vec<u8>, Vec<u8>)> = ConvertKey::convert(key.clone());
+		let Some((blockchain,symbol)) = dia_key else {
```
b-yap, line 99: since this function already returns an `Option`, you can actually use the [`?` operator](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/reference/expressions/operator-expr.html#the-question-mark-operator);
```rust
let (blockchain,symbol) = ConvertKey::convert(key.clone())?;
```

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,121 @@
...
+        };
+
+		let Ok(coin_info) = Dia::get_coin_info(blockchain, symbol) else {
+            return None;
+        };
+
+		let Some(value) = ConvertPrice::convert(coin_info.price) else{
+            return None;
+        };
+		let Some(timestamp) = ConvertMoment::convert(coin_info.last_update_timestamp) else{
```
b-yap, line 110: same thing here

pallets/oracle/src/lib.rs
ebma: I assume `T::AccountId::new(Default::default())` also does not work?

cr4pt0: It also does work. @ebma 

```
let r = T::AccountId::new(Default::default());
    |                               ^^^ function or associated item not found in `<T as frame_system::Config>::AccountId`
```

```
let r = T::AccountId::from(Default::default());
    |                                    ^^^^^^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `<T as frame_system::Config>::AccountId`
```

pallets/oracle/src/dia.rs
cr4pt0: thanks! You are right!

pallets/oracle/src/lib.rs
cr4pt0: removed option 
```
AggregateUpdated {
			values: Vec<(OracleKey, T::UnsignedFixedPoint)>,
		},
```
there is no need for option in AggregateUpdated pallet event

testchain/node/src/chain_spec.rs
```diff
@@ -340,5 +351,11 @@ fn testnet_genesis(
 			replace_griefing_collateral: FixedU128::checked_from_rational(1, 10).unwrap(), // 10%
 		},
 		nomination: NominationConfig { is_nomination_enabled: false },
+		dia_oracle_module: DiaOracleModuleConfig {
+			authorized_accounts: authorized_oracles,
+			supported_currencies: vec![AssetId::new(b"Bitcoin".to_vec(), b"BTC".to_vec())],
```
ebma, line 356: I guess we should change this away from btc?

testchain/runtime/src/lib.rs
```diff
@@ -396,9 +402,97 @@ where
...
+pub struct ConvertMoment;
+impl Convert<u64, Option<Moment>> for ConvertMoment {
+	fn convert(a: u64) -> Option<Moment> {
+		Some(a)
+	}
+}
+pub struct DataCollector;
+impl DataProvider<OracleKey, TimestampedValue<UnsignedFixedPoint, Moment>> for DataCollector {
+	fn get(key: &OracleKey) -> Option<TimestampedValue<UnsignedFixedPoint, Moment>> {
+		todo!()
```
ebma, line 467: What about this?

testchain/runtime/src/lib.rs
```diff
@@ -396,9 +402,97 @@ where
...
+}
+impl orml_oracle::DataFeeder<OracleKey, TimestampedValue<UnsignedFixedPoint, Moment>, AccountId>
+	for DataCollector
+{
+	fn feed_value(
+		who: AccountId,
+		key: OracleKey,
+		value: TimestampedValue<UnsignedFixedPoint, Moment>,
+	) -> sp_runtime::DispatchResult {
+		todo!()
```
ebma, line 478: and this?

ebma (changes requested):
These are lots of complex changes. And it works overall, so good job @cr4pt0 :+1:. 
However there are some things I'd like to have improved. Some minor things like variable renamings but also missing implementations (that are not 100% required but IMO we should not cut corners here). 


Also, when running all tests, the vault-registry benchmark and `vault_integration_tests::test_cancel_scheduler_succeeds` failed. It is also not possible to just run the integration tests without the --all flag. At least for me, running `cargo test --package vault --features standalone-metadata` does not compile anymore while `cargo test --all --all-features` works and also runs the integration tests

So please address my comments and make sure that these tests also work.

clients/runtime/src/rpc.rs
```diff
@@ -902,11 +912,15 @@ impl OraclePallet for SpacewalkParachain {
...
+	async fn has_updated(&self, blockchain: Vec<u8>, symbol: Vec<u8>) -> Result<bool, Error> {
+		use crate::metadata::runtime_types::dia_oracle::dia::AssetId;
+		let asset_id = AssetId { blockchain, symbol };
+		let result = self
+			.query_finalized_or_error(
+				metadata::storage().dia_oracle_module().coin_infos_map(&asset_id),
+			)
+			.await;
+		Ok(result.is_ok())
 	}
```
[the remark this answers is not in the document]
cr4pt0: You are right!. We really just need remove all this function. we do need and do not use this function anymore.

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,114 @@
...
+            return None;
+        };
+
+		Some(TimestampedValue { value, timestamp })
+	}
+
+	///do not need this funtion implementation
+	fn get_all_values() -> Vec<(OracleKey, Option<TimestampedValue<UnsignedFixedPoint, Moment>>)> {
+		panic!("spacewalk oracle extension does not requre implementation of DataProviderExtended get_all_values function")
+	}
```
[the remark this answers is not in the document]
cr4pt0: Thanks! You are right. 
But the problem is that we even can't implement this function for DIA oracle. Because they did not provide the function in `DiaOracle` trait that will return list of key or all values. This is limitation of DIA oracle pallet.
here is like DIAOracle trait looks like
```
pub trait DiaOracle {
    /// Returns the coin info by given name
    fn get_coin_info(blockchain: Vec<u8>, symbol: Vec<u8>) -> Result<CoinInfo, DispatchError>;

    /// Returns the price by given name
    fn get_value(blockchain: Vec<u8>, symbol: Vec<u8>) -> Result<PriceInfo, DispatchError>;
}
```

testchain/runtime/src/lib.rs
cr4pt0: Added comments and change macros from `todo` to `unimplemented`
`DataFeeder` required to implement `DataProvider` trait but there is no need to implement get function.
Not required to implement `DataProvider` get function

cr4pt0:
> These are lots of complex changes. And it works overall, so good job @cr4pt0 👍. However there are some things I'd like to have improved. Some minor things like variable renamings but also missing implementations (that are not 100% required but IMO we should not cut corners here).
> 
> Also, when running all tests, the vault-registry benchmark and `vault_integration_tests::test_cancel_scheduler_succeeds` failed. It is also not possible to just run the integration tests without the --all flag. At least for me, running `cargo test --package vault --features standalone-metadata` does not compile anymore while `cargo test --all --all-features` works and also runs the integration tests
> 
> So please address my comments and make sure that these tests also work.

`test_cancel_scheduler_succeeds` did not work at all in dev/milestone-3 branch. i fix it now.

cr4pt0:
> These are lots of complex changes. And it works overall, so good job @cr4pt0 👍. However there are some things I'd like to have improved. Some minor things like variable renamings but also missing implementations (that are not 100% required but IMO we should not cut corners here).
> 
> Also, when running all tests, the vault-registry benchmark and `vault_integration_tests::test_cancel_scheduler_succeeds` failed. It is also not possible to just run the integration tests without the --all flag. At least for me, running `cargo test --package vault --features standalone-metadata` does not compile anymore while `cargo test --all --all-features` works and also runs the integration tests
> 
> So please address my comments and make sure that these tests also work.

@ebma 
Thank you so much. Very useful comments and requests for changes.
Fixed everything that you mention!

clients/runtime/src/integration/mod.rs
```diff
@@ -80,27 +82,26 @@ pub async fn setup_provider(client: SubxtClient, key: AccountKeyring) -> Spacewa
...
+async fn wait_for_aggregate(parachain_rpc: &SpacewalkParachain, key: &OracleKey) {}
 
 pub async fn set_exchange_rate_and_wait(
 	parachain_rpc: &SpacewalkParachain,
 	currency_id: CurrencyId,
 	value: FixedU128,
 ) {
 	let key = OracleKey::ExchangeRate(currency_id);
-	assert_ok!(parachain_rpc.feed_values(vec![(key.clone(), value)]).await);
+	let t = DiaOracleKeyConvertor::convert(key.clone()).unwrap();
```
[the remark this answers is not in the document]
cr4pt0: Done.

clients/runtime/src/integration/mod.rs
```diff
@@ -80,27 +82,26 @@ pub async fn setup_provider(client: SubxtClient, key: AccountKeyring) -> Spacewa
...
 	parachain_rpc.manual_seal().await;
 	// we need a new block to get on_initialize to run
 	assert_ok!(timeout(TIMEOUT_DURATION, wait_for_aggregate(parachain_rpc, &key)).await);
 }
 
 pub async fn set_stellar_fees(parachain_rpc: &SpacewalkParachain, value: FixedU128) {
-	assert_ok!(parachain_rpc.set_stellar_fees(value).await);
+	// assert_ok!(parachain_rpc.set_stellar_fees(value).await);
+	let key = OracleKey::FeeEstimation;
+	let t = DiaOracleKeyConvertor::convert(key.clone()).unwrap();
```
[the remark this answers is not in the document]
cr4pt0: Done

clients/runtime/src/rpc.rs
```diff
@@ -826,42 +826,52 @@ pub trait OraclePallet {
...
-	/// Sets the estimated fee required to get a Stellar transaction included in the next block.
-	///
-	/// # Arguments
-	/// * `value` - the estimated fee rate
-	async fn set_stellar_fees(&self, value: FixedU128) -> Result<(), Error> {
-		self.with_retry(
-			metadata::tx().oracle().feed_values(vec![(OracleKey::FeeEstimation, value)]),
-		)
-		.await?;
+		for i in values {
```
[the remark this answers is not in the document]
cr4pt0: Done

clients/runtime/src/tests.rs
```diff
@@ -28,10 +29,11 @@ fn dummy_public_key() -> StellarPublicKeyRaw {
 
 async fn set_exchange_rate(client: SubxtClient) {
 	let oracle_provider = setup_provider(client, AccountKeyring::Bob).await;
-	let key = OracleKey::ExchangeRate(DEFAULT_TESTING_CURRENCY);
+	let key = primitives::oracle::Key::ExchangeRate(DEFAULT_TESTING_CURRENCY);
+	let t = DiaOracleKeyConvertor::convert(key.clone()).unwrap();
```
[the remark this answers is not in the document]
cr4pt0: Done.

pallets/oracle/src/lib.rs
```diff
@@ -110,39 +126,15 @@ pub mod pallet {
...
-	#[pallet::storage]
-	pub type ValidUntil<T: Config> = StorageMap<_, Blake2_128Concat, OracleKey, T::Moment>;
-
 	/// Maximum delay (milliseconds) for a reported value to be used
 	#[pallet::storage]
 	#[pallet::getter(fn max_delay)]
 	pub type MaxDelay<T: Config> = StorageValue<_, T::Moment, ValueQuery>;
 
-	// Oracles allowed to set the exchange rate, maps to the name
+	// Oracles keys required by runtime
```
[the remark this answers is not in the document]
cr4pt0: Done.

testchain/node/src/chain_spec.rs
cr4pt0: Done.

clients/runtime/src/integration/mod.rs
```diff
@@ -80,27 +82,26 @@ pub async fn setup_provider(client: SubxtClient, key: AccountKeyring) -> Spacewa
 const SLEEP_DURATION: Duration = Duration::from_millis(1000);
 const TIMEOUT_DURATION: Duration = Duration::from_secs(20);
 
-async fn wait_for_aggregate(parachain_rpc: &SpacewalkParachain, key: &OracleKey) {
-	while parachain_rpc.has_updated(key).await.unwrap() {
-		// should be false upon aggregate update
-		sleep(SLEEP_DURATION).await;
-	}
-}
+async fn wait_for_aggregate(parachain_rpc: &SpacewalkParachain, key: &OracleKey) {}
```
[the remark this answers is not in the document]
cr4pt0: You right. Since we remove logic about aggregation in spacewalk oracle there is no needs to wait for aggregate. 
Will remove it.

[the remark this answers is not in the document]
cr4pt0: Done. Thanks

pallets/oracle/src/lib.rs
ebma: Okay, then I'm fine with keeping it the way you implemented it.

clients/runtime/src/tests.rs
```diff
@@ -79,34 +81,6 @@ async fn test_too_low_priority_matching() {
...
-		event_listener,
-		invalid_oracle.feed_values(vec![(key.clone(), exchange_rate)]),
-		oracle_provider.feed_values(vec![(key, exchange_rate)])
-	);
-
-	// ensure first set_exchange_rate failed and second succeeded.
-	result.1.unwrap_err();
-	result.2.unwrap();
-}
-
```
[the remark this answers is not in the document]
ebma: @cr4pt0 what about this?

pallets/oracle/src/dia.rs
```diff
@@ -0,0 +1,114 @@
...
+						return Some((
+							KSM_DIA_BLOCKCHAIN.as_bytes().to_vec(),
+							KSM_DIA_SYMBOL.as_bytes().to_vec(),
+						)),
+					_ => unimplemented!(),
+				},
+				CurrencyId::Native => unimplemented!(),
+				CurrencyId::StellarNative => unimplemented!(),
+				CurrencyId::AlphaNum4 { .. } => unimplemented!(),
+				CurrencyId::AlphaNum12 { .. } => unimplemented!(),
```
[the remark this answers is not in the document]
ebma: @cr4pt0 and this?

pallets/oracle/src/dia.rs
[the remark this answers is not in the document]
ebma: Hmm, I see. But we can query the coin info of all items by using the `Dia::prices_map()` getter function (see [here](https://github.com/pendulum-chain/oracle-pallet/blob/af8517a2d9eb469c9681195e6653aa519f05cf28/pallets/dia-oracle/src/lib.rs#L109-L111)). So I guess it actually should be possible?

testchain/runtime/src/lib.rs
ebma: Can you change it to `unimplemented()` then and also state the reason why it's not implemented in a comment, or the macro itself?

testchain/runtime/src/lib.rs
ebma: Can you change it to `unimplemented()`then and also state the reason why it's not implemented in a comment, or the macro itself?

b-yap (approved):
I have no more comments to add.

@ebma?

clients/runtime/src/tests.rs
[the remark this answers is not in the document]
cr4pt0: @ebma 
Because there is no need in this tests because there is no Event about feeding in oracle spacewalk.
It is not relevant already. `FeedValuesEvent` does not exist in new redesigned oracle in spacewalk.

pallets/oracle/src/dia.rs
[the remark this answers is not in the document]
cr4pt0: > Hmm, I see. But we can query the coin info of all items by using the `Dia::prices_map()` getter function (see [here](https://github.com/pendulum-chain/oracle-pallet/blob/af8517a2d9eb469c9681195e6653aa519f05cf28/pallets/dia-oracle/src/lib.rs#L109-L111)). So I guess it actually should be possible?

Thanks. You are right. But. this function not included to DIAOracle trait that they provide.
We use bound 
```
impl<Dia, UnsignedFixedPoint, Moment, ConvertKey, ConvertPrice, ConvertMoment>
	DataProviderExtended<OracleKey, TimestampedValue<UnsignedFixedPoint, Moment>>
	for DiaOracleAdapter<Dia, UnsignedFixedPoint, Moment, ConvertKey, ConvertPrice, ConvertMoment>
where
	Dia: DiaOracle,
	ConvertKey: Convert<OracleKey, Option<(Vec<u8>, Vec<u8>)>>
		+ Convert<(Vec<u8>, Vec<u8>), Option<OracleKey>>,
	ConvertPrice: Convert<u128, Option<UnsignedFixedPoint>>,
	ConvertMoment: Convert<u64, Option<Moment>>,
{}
```
**Dia: DiaOracle**
Of course we can add more not flexible logic to retrieve data about all keys from DIA oracle but actually we **do not need it**. Because we store all required keys for spacewalk oracle in storage `OracleKeys<Vec<OracleKey>>`.

clients/runtime/src/tests.rs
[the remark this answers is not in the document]
ebma: Yes that might be true. But that was not the point of my comment.
I was trying to say that although the actual implementation of this test is using the Oracle, this might just be coincidental and they could have used any other extrinsics to test it. The name of the test 'test_subxt_processing_events_after_dispatch_error' indicates exactly this; that it is not about the oracle but about testing something about subxt. Hence why I was asking if we can change the implementation and keep the test.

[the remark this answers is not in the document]
ebma: I think the main reason they used the oracle pallet for this test is that it is easy to execute one extrinsic successfully and let another fail. And you said that we don't have the `FeedValuesEvent` but that is not 100% true since we do have this event but it is only emitted in tests. Since this is a test however, it should be emitted, no?

pallets/oracle/src/dia.rs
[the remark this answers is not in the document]
ebma: Okay got it, makes sense @cr4pt0 thanks for the explanation. We could point this out by adding a comment in the `unimplemented()` function calls but I'm fine if we don't

clients/runtime/src/tests.rs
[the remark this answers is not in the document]
cr4pt0: No. `FeedValuesEvent` emitted only for Mock runtime. for tests.
In `test_subxt_processing_events_after_dispatch_error` function feed_values calls DIA oracle extrinsic and not an spacewalk oracle. 
Seriously. other tests did the same. 
For example `test_automatic_issue_execution_succeeds` did 
```
assert_event::<ExecuteIssueEvent, _>(TIMEOUT, user_provider.clone(), move |x| {
	x.vault_id == vault_id.clone()
})
.await;
```
and create the user_provider
`let user_provider = setup_provider(client.clone(), AccountKeyring::Dave).await;`
I do not see any reason to keep this function or reimplement it. Because the original purpose of 
`test_subxt_processing_events_after_dispatch_error` was to check that the is an error when feed oracle with incorrect oracle provider. 
@ebma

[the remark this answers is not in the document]
ebma: To be sure what the original purpose of the function was we would have to ask developers from interlay. I still argue that the name of the test, that is 'test_subxt_processing_events_after_dispatch_error' tells us its purpose and not its implementation. 
But if you don't want to re-implement it, then that's okay and we don't. I just want to make sure you get my point.

ebma (approved):
I'd say it's good enough for now. I'll create a follow-up task for refactoring/deduplicating the mock oracle logic, but we can tackle that some time later.

Thanks for all your efforts(and suffering) @RustNinja  and also @b-yap. We can merge it now.

PR #188, opened 2023-01-12
Use DataProvider type in oracle from Pallet config

cr4pt0:
- add `DataProvider` to oracle pallet config
- Add new Storage `Vec<OracleKeys>`
- Add new extrinsic to update the Storage with `oracle_keys`
- load prices `on_initialize` to track the price status and move parachain offline or online
- remove unused `Storages`

### DIFF
--- a/pallets/oracle/src/lib.rs
+++ b/pallets/oracle/src/lib.rs
@@ -11,7 +11,7 @@ extern crate mocktopus;
 use codec::{Decode, Encode, MaxEncodedLen};
 #[cfg(feature = "testing-utils")]
 use frame_support::dispatch::DispatchResult;
-use frame_support::{dispatch::DispatchError, traits::Time, transactional};
+use frame_support::{dispatch::DispatchError, transactional};
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


pallets/oracle/src/lib.rs
```diff
@@ -211,60 +162,20 @@ pub mod pallet {
...
-		/// * `account_id` - the account Id of the oracle
+		/// * `oracle_key` - list of oracle keys
 		#[pallet::call_index(2)]
 		#[pallet::weight(<T as Config>::WeightInfo::remove_authorized_oracle())]
 		#[transactional]
-		pub fn remove_authorized_oracle(
+		pub fn update_oracle_key(
 			origin: OriginFor<T>,
-			account_id: T::AccountId,
+			oracle_key: Vec<OracleKey>,
```
ebma, line 174: This should be named `update_oracle_keys()` and the included variable should be `oracle_keys` since we use a vec here.

ebma (commented):
Makes sense overall

PR #236, opened 2023-01-27
remove Key::FeeEstimation. update metadata.scale. fix issue with types.

cr4pt0:
- remove Key::FeeEstimation. 
- update metadata.scale. 
- fix issue with types. ArithmeticError moved to other module so the tests for new metadata scale failed. fixed the problem.


clients/runtime/src/types.rs
```diff
@@ -246,7 +251,7 @@ mod dispatch_error {
 		Unsupported,
 	);
 
-	convert_enum!(RichArithmeticError, ArithmeticError, Underflow, Overflow, DivisionByZero,);
+	// convert_enum!(RichArithmeticError, ArithmeticError, Underflow, Overflow, DivisionByZero,);
```
ebma, line 254: Why is this commented out?

pallets/oracle/src/tests.rs
```diff
@@ -205,7 +205,7 @@ fn test_is_invalidated() {
 #[test]
 fn set_xlm_tx_fees_per_byte_succeeds() {
```
ebma, line 206: I guess we can remove this test altogether, since by removing the FeeEstimation item, we remove the need for the test

ebma (changes requested):

clients/runtime/src/types.rs
cr4pt0: fixed.
reuse auto implement macros instead of match statement for ArithmeticError

pallets/oracle/src/tests.rs
cr4pt0: Done.

clients/runtime/src/types.rs
```diff
@@ -214,7 +214,10 @@ mod vault_id {
 
 mod dispatch_error {
 	use crate::metadata::{
-		runtime_types::sp_runtime::{ArithmeticError, ModuleError, TokenError, TransactionalError},
+		runtime_types::{
+			sp_arithmetic::ArithmeticError as OtherArithmeticError,
```
ebma, line 218: Why can't we leave the name as 'ArithmeticError'? Why does it need to be 'OtherArithmeticError'?

cr4pt0: Oh. Sure . You are right. Now we can. Fixed.
There was a line with the import of other ArithmeticError.
I removed that line and now there is no need for alias.
Thanks. FIxed.

clients/vault/tests/vault_integration_tests.rs
```diff
@@ -212,7 +210,6 @@ async fn test_redeem_succeeds() {
 		get_exchange_rate(&vault_provider, DEFAULT_TESTING_CURRENCY).await;
 		get_exchange_rate(&vault_provider, DEFAULT_WRAPPED_CURRENCY).await;
 		get_exchange_rate(&vault_provider, DEFAULT_NATIVE_CURRENCY).await;
-		assert_eq!(keys.len(), 3);
```
ebma, line 215: Is it necessary to completely remove this line? I assume there is now some kind of mismatch and the assertion does not hold anymore. But if that's the case, can't we simply change the number we expect?

ebma:
Just FYI, I asked something related to this change on Slack and we have to wait until it is discussed before merging this.

clients/vault/tests/vault_integration_tests.rs
cr4pt0: Yes. Right. Done!

ebma: Does this still work? Removing lines 211-215 does not have an impact on the outcome of the `test_redeem_succeeds()` integration test? 😳

ebma: @cr4pt0 did you check if this particular test still works after the changes?

ebma:
The discussion was resolved on Slack and we decided not to have the inclusion fee for redeem requests, thus we can go ahead and remove the `FeeEstimation` key. There are two warnings about unused imports that should be fixed and I asked about the other test case, which I would like to have resolved before merging @cr4pt0.

clients/vault/tests/vault_integration_tests.rs
cr4pt0: Yes. It works. i added that line trying solve the issue that i have with integration tests.
But lines that we removed are not necessary,` test_with_vault` function feed values for tests.
@ebma

cr4pt0:
> The discussion was resolved on Slack and we decided not to have the inclusion fee for redeem requests, thus we can go ahead and remove the `FeeEstimation` key. There are two warnings about unused imports that should be fixed and I asked about the other test case, which I would like to have resolved before merging @cr4pt0.

great! fixed warning.
pr ready for merge.

ebma (approved):
Ready for merge, good job 👍

PR #237, opened 2023-01-30
211 update test case for the ci jobs

b-yap:
Base code is actually from @cr4pt0's 233 branch, so https://github.com/pendulum-chain/spacewalk/pull/236 should be merged first.

* made sure to use the `shutdown()` for the write half of stream
* made sure to drop the connector when disconnected
* ^ if/when everything works, we won't need the `is_disconnected` field, since the _`send()`_ function will return **_`Err(SendFailed("channel closed"))`_**


cr4pt0 (approved):

clients/stellar-relay-lib/src/connection/services.rs
```diff
@@ -298,10 +305,12 @@ pub(crate) async fn connection_handler(
...
 			},
 
-			Ok(None) => {},
+			Ok(None) => {
+				log::info!("why is it none?");
+			},
 
 			Err(elapsed) => {
-				log::error!("{} for receiving messages.", elapsed.to_string());
+				log::info!("{} for receiving messages.", elapsed.to_string());
```
ebma, line 313: This message probably also is not accurate (?). If I understand correctly it should be something like "Connection timed out after {} seconds."

clients/stellar-relay-lib/src/tests/mod.rs
```diff
@@ -84,41 +97,54 @@ async fn stellar_overlay_should_receive_tx_set() {
...
-							.send(StellarMessage::GetTxSet(txset_hash))
-							.await
-							.unwrap();
-					}
-				},
-				StellarMessage::TxSet(set) => {
-					tx_set_vec.push(set);
-					break
+	let mut overlay_connection = Arc::new(Mutex::new(
+		StellarOverlayConnection::connect(node_info.clone(), cfg).await.unwrap(),
```
ebma, line 102: Is the comma at the end of line necessary or is it a typo?

clients/stellar-relay-lib/src/tests/mod.rs
```diff
@@ -84,41 +97,54 @@ async fn stellar_overlay_should_receive_tx_set() {
...
+	})
+	.await
+	.expect("time has elapsed");
+
+	overlay_connection
+		.lock()
+		.await
+		.disconnect()
+		.await
+		.expect("Should be able to disconnect");
```
ebma, line 144: Do we really need to disconnect twice? There already is a disconnect happening in l. 127, so I'm not sure if/why we need both.

ebma (approved):
LGTM

PR #240, opened 2023-02-06
implement pull node

b-yap:
Based on https://github.com/stellarbeat/js-stellar-node-connector/commit/3b032d6f545d4c421b976e970d54e0ada60d7cf9
To fix the occassional network disconnection every 2 or 3 minutes


ebma (approved):
I couldn't figure out what exactly makes the difference here. I assume that it's the change of `flags` to 100 but the stellar types don't have documentation about it.
Since it works anyways and it's not that important we know how the flags work, I'm happy to approve it though

cr4pt0 (approved):

PR #245, opened 2023-02-08
244 - Change handling of open requests

b-yap:


cr4pt0 (approved):

ebma (changes requested):
We should change it to store unsuccessful requests and retry them instead.

cr4pt0 (approved):

clients/vault/src/execution.rs
```diff
@@ -259,6 +259,42 @@ impl Request {
...
+						continue
+					}
+					break
+				},
+				Err(error) => {
+					retry_count += 1; // increase retry count
+					tracing::error!("Failed to get proof for slot {}: {:?}", slot, error);
+				},
+			}
+		}
```
ebma, line 292: Should we add a line that prints some log if we exhausted all our retry counts and it still did not work? Or should we maybe in l. 280 mention that we are retrying and also mention the retry count?

clients/vault/src/execution.rs
```diff
@@ -259,6 +259,42 @@ impl Request {
...
+			match oracle_agent.get_proof(slot).await {
+				Ok(proof) => {
+					if let Err(e) =
+						request.execute(parachain_rpc.clone(), tx_env.clone(), proof).await
+					{
+						tracing::error!("Failed to execute request #{}: {}", request.hash, e);
+						retry_count += 1; // increase retry count
+						continue
+					}
+					break
```
ebma, line 285: Can we rewrite it with a match statement to make it more obvious that we break the while function on successful submit? It took me a second to figure out why we `continue` in one case and `break` afterwards. Maybe that makes it clearer.

ebma (approved):

PR #254, opened 2023-02-13
 implement one function for currencyid

b-yap:
this is for ~~[PR 90](https://github.com/pendulum-chain/pendulum/pull/109#discussion_r1101240189)~~ of Pendulum project.
new PR opened: [PR 116](https://github.com/pendulum-chain/pendulum/pull/116)


cr4pt0 (approved):

primitives/src/lib.rs
```diff
@@ -507,6 +507,20 @@ pub enum CurrencyId {
 	AlphaNum12 { code: Bytes12, issuer: AssetIssuer },
 }
 
+impl CurrencyId {
+	fn decimals() -> u8 {
+		// scaling allows for seven decimal places of precision
+		7
```
ebma, line 513: While this is good in theory, I would appreciate it if we could maybe make this only apply to the Stellar currencies. 

This however would mean that we'd have to change the `CurrencyId` enum. We could change it to something like 
```
pub enum CurrencyId {
	XCM(ForeignCurrencyId),
	Native,
        Stellar(Asset),
}
pub enum Asset {
	StellarNative,
	AlphaNum4 { code: Bytes4, issuer: AssetIssuer },
	AlphaNum12 { code: Bytes12, issuer: AssetIssuer },
}

impl Asset {
	fn decimals() -> u8 {
		// scaling allows for seven decimal places of precision
		7
         }
         ...
}
```

ebma: WDYT @b-yap ?

b-yap: done. I reused the `CurrencyInfo` trait for the `Asset`.

Issue #242, opened 2023-02-07
Fix minor issues


ebma:
There are some things/issues we should address. These include
- [ ] An unnecessary conversion of a vector (since we do not filter here anymore). Location: `/pallets/stellar-relay/src/lib.rs` L460 and L469
- [ ] Duplicated Code in the `TryFrom` implementations for CurrencyId in ` /primitives/src/lib.rs` L509. We should implement the logic only once and reuse it in both
- [ ] The Stellar secret key is exposed in the CLI invocation of the vault client. We should change it such that this secret key is rather read from a local file because other processes on the host machine could otherwise see this secret key.
- [ ] There is no weight for the `mint_tokens_for_reimbursed_redeem()` but it is using the weights for `set_redeem_period`. We should add another benchmark function that calls `mint_tokens_for_reimbursed_redeem()` so that the weight for this is added to the default_weights.rs. See `/pallets/redeem/src/lib.rs` L390. 
- [ ] There is an unsafe integer cast happening in the `unlookup()` function of the BalanceConversion implementation `/primitives/src/lib.rs` lib.rs  L683. The type conversion from type i64 to type u128 may flip the value's sign. We should add safety checks here.
- [ ] We have an unnecessary reference to the secret key of the stellar account in `/clients/vault/src/system.rs` in L415, which we don't need because we are only interested in the public key. Instead of storing the secret key in a local variable we can just use `wallet.get_secret_key().get_public();` in the statement.

ebma:
Closed by #264

PR #255, opened 2023-02-13
242 - secret exposed in cli

b-yap:
this the 3rd item of [issue 242](https://github.com/pendulum-chain/spacewalk/issues/242)


ebma (approved):
The code looks good. I am wondering if we would maybe want to have a more 'sophisticated' config file, that is also able to take more than one parameter.

But let's merge this now and tackle the other thing later. Maybe in issue 261 instead.

PR #256, opened 2023-02-13
242 - unsafe integer cast for `BalanceConversion`'s `StaticLookUp` implementation

b-yap:
This is item #5 of [#242](https://github.com/pendulum-chain/spacewalk/issues/242)
if conversion rate is too big for i64, set it to what its max is.
if the i64 value reaches negative, make it zero for u128.

^ Unless that's the wrong approach?

### DIFF
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -489,7 +489,7 @@ create_currency_id! {
 		PHA("Phala", 10) = 14,
 		ZTG("Zeitgeist", 10) = 15,
 		USD("Statemine", 10) = 16,
-		
+
 		DOT("Polkadot", 10) = 20,
 	}
 }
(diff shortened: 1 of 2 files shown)


primitives/src/lib.rs
```diff
@@ -696,7 +696,10 @@ impl StaticLookup for BalanceConversion {
 	}
 
 	fn unlookup(stellar_stroops: Self::Target) -> Self::Source {
-		(stellar_stroops * CONVERSION_RATE as i64) as u128
+		let conversion_rate = i64::try_from(CONVERSION_RATE).unwrap_or(i64::MAX);
+
+		let value = stellar_stroops * conversion_rate;
```
ebma, line 701: Maybe we should even use `saturating_mul()` here, just to be safe

ebma: @b-yap WDYT?

b-yap: done

cr4pt0 (approved):

ebma (approved):

PR #264, opened 2023-02-14
Fix minor issues

### LINKED ISSUE #242: Fix minor issues

ebma:
Closes #242.


ebma:
@b-yap can you resolve the issues with conflicting files? Afterward I think we can merge it, AFAIK all issues mentioned in the ticket are addressed.

PR #265, opened 2023-02-14
implement disconnection for overlay conn in agent.rs

cr4pt0:


clients/vault/src/oracle/agent.rs
```diff
@@ -170,6 +171,8 @@ impl OracleAgent {
 		self.message_sender = Some(sender.clone());
 
 		let collector = self.collector.clone();
+		let actions_sender = overlay_conn.get_actions_sender();
+		let r = overlay_conn.disconnect_status();
```
ebma, line 175: I'd rename this variable. Maybe something like `disconnect_action`

clients/stellar-relay-lib/src/connection/overlay_connection.rs
```diff
@@ -22,6 +22,15 @@ pub struct StellarOverlayConnection {
...
 
+impl StellarOverlayConnection {
+	pub fn get_actions_sender(&self) -> mpsc::Sender<ConnectorActions> {
+		self.actions_sender.clone()
+	}
+	pub fn disconnect_status(&self) -> ConnectorActions {
+		ConnectorActions::Disconnect
+	}
+}
+
```
ebma, line 33: Let's maybe just move the two functions to the `impl` block below. But put it at the very bottom of it.

clients/stellar-relay-lib/src/connection/overlay_connection.rs
```diff
@@ -22,6 +22,15 @@ pub struct StellarOverlayConnection {
 	is_disconnected: bool,
 }
 
+impl StellarOverlayConnection {
+	pub fn get_actions_sender(&self) -> mpsc::Sender<ConnectorActions> {
+		self.actions_sender.clone()
+	}
+	pub fn disconnect_status(&self) -> ConnectorActions {
```
ebma, line 29: I'd rename this to `get_disconnect_action()` to make it clearer that the function returns the `ConnectorAction` to disconnect.

ebma (changes requested):

clients/stellar-relay-lib/src/connection/overlay_connection.rs
cr4pt0: Good point. Done.

clients/stellar-relay-lib/src/connection/overlay_connection.rs
cr4pt0: You are right. done.

ebma (commented):
Thanks for adding my request changes @cr4pt0. 
But now that I look at it a second time I'm wondering if it would not make sense to just combine the two functions `get_actions_sender` and `get_disconnect_action` into one? At the moment we use them together anyways, so maybe we should just create one function `send_disconnect_action()` that does both? WDYT @cr4pt0 ?

cr4pt0:
> 

Yes. we can do it.
There are two ways:
- create a new function that will return a tuple
- implement send_disconnect for Sender.

ebma:
Ah right, the reason we have these two functions is because the `overlay_conn` is moved as part of the spawn. I forgot about it. Well then that's okay. Can you just add a comment like this? Then it's good for merge.
```
++ // Get action sender and disconnect action before moving `overlay_conn` into the closure
     let actions_sender = overlay_conn.get_actions_sender();
     let disconnect_action = overlay_conn.get_disconnect_action();
```

cr4pt0:
> Ah right, the reason we have these two functions is because the `overlay_conn` is moved as part of the spawn. I forgot about it. Well then that's okay. Can you just add a comment like this? Then it's good for merge.
> 
> ```
> ++ // Get action sender and disconnect action before moving `overlay_conn` into the closure
>      let actions_sender = overlay_conn.get_actions_sender();
>      let disconnect_action = overlay_conn.get_disconnect_action();
> ```

Ready.

ebma (approved):

PR #266, opened 2023-02-15
do not update the old set and new set if current value equal to updated

### LINKED ISSUE #260: Validate new validator set before updating
We should validate the validator set that is passed to the `update_tier_1_validator_set` extrinsic in `stellar-relay`. 
More details can be found in the `LIT-01` item of the audit report.

RustNinja:
The extrinsic _update_tier_1_validator_set() from the stellar-relay pallet allows the root user to change the Tier 1 Validators from Stellar.

Such a method does not validate the new validators set. Such a set could have repeated or empty values or it could be the same set as the old one. It's a good practice to include such checks to prevent those errors from happening.

Recommendation
We advise the team to add the necessary checks to prevent the new vector from being the same as the old one and avoid repeated and empty values to be included in the new validator set.

- actually we do not need to validate for empty vector because we need to support functionality that remove validators and organisation
- only think is reasonable to include validation to not update the Storage if old set is equals to new set.

Closes #260


pallets/stellar-relay/src/lib.rs
```diff
@@ -475,19 +474,26 @@ pub mod pallet {
...
+
+			//update only when new organization or validators not equal to old organization or validators
+			if new_organization_vec != current_organizations ||
+				new_validator_vec != current_validators
+			{
+				OldValidators::<T>::put(current_validators.clone());
+				OldOrganizations::<T>::put(current_organizations.clone());
+				Validators::<T>::put(new_validator_vec.clone());
+				Organizations::<T>::put(new_organization_vec.clone());
+			}
```
ebma, line 496: I think we should do this differently. Since we need to add other checks I'd propose we do something like this:
```
// Check if validators and organizations change
ensure!(new_organization_vec != current_organizations && new_validator_vec != current_validators, Error::<T>::SameValidatorsAndOrganizations);

// Check for duplicates in the validators vec
// TODO figure out the best way to check for duplicates in this vec (should not be too computationally expensive)
// return an appropriate error if we encounter a duplicate

// Check for duplicates in the organizations vec
// TODO figure out the best way to check for duplicates in this vec (should not be too computationally expensive)
// return an appropriate error if we encounter a duplicate

// Set the values after all checks passed
OldValidators::<T>::put(current_validators.clone());
OldOrganizations::<T>::put(current_organizations.clone());
Validators::<T>::put(new_validator_vec.clone());
Organizations::<T>::put(new_organization_vec.clone());
```

Note that I deliberately used an `&&` operator instead of `||` because it might be that we want to update just an organization **or** something about the validators. If we used `||` then we'd always have to change something

ebma (changes requested):

pallets/stellar-relay/src/lib.rs
cr4pt0: @ebma 
add validation for duplication organisation id and public_key in validators.
fix unit tests to pass different organisation id and public key.

pallets/stellar-relay/src/tests.rs
```diff
@@ -448,8 +452,11 @@ fn update_tier_1_validator_set_works() {
 			public_key: Default::default(),
 			organization_id: organization.id,
 		};
-		let new_validator_set = vec![validator; 2];
-		let new_organization_set = vec![organization; 2];
+
+		let mut validator_1 = validator.clone();
+		validator_1.public_key = BoundedVec::<u8, FieldLength>::try_from(vec![1u8; 128]).unwrap();
```
ebma, line 457: This actually boils down to `validator_1` having the same public key as previously (in line 425). So we could just clone and re-use the validator_1 we had previously I guess (?)

pallets/stellar-relay/src/tests.rs
```diff
@@ -534,13 +541,14 @@ fn verify_signature_works_for_mock_message() {
 fn update_tier_1_validator_store_old_organization_and_validator_and_block_height_works() {
 	run_test(|_, _, _| {
 		let organization = Organization { id: 0, name: Default::default() };
+
 		let validator = Validator {
 			name: Default::default(),
-			public_key: Default::default(),
+			public_key: BoundedVec::<u8, FieldLength>::try_from(vec![1u8; 128]).unwrap(),
```
ebma, line 547: Why are you not using `Default::default()` anymore?

pallets/stellar-relay/src/tests.rs
```diff
@@ -569,8 +577,11 @@ fn update_tier_1_validator_store_old_organization_and_validator_and_block_height
 			public_key: Default::default(),
 			organization_id: organization.id,
 		};
-		let new_validator_set = vec![validator; 2];
-		let new_organization_set = vec![organization; 2];
+		let mut validator_2 = validator.clone();
+		validator_2.public_key = BoundedVec::<u8, FieldLength>::try_from(vec![1u8; 128]).unwrap();
```
ebma, line 581: Here you are assigning the same key as in line 547. Is this a mistake or supposed to be like this? I would actually expect this test to fail now. Or with the changes you added in this PR we do want this test to fail. Because now we call the update extrinsic with two validators that have the same public key which is problematic and should throw the 'Duplicate...' error

ebma (changes requested):
I requested some changes. 

It would also be nice to have a new test function `update_tier_1_validator_set_fails_for_duplicate_values` in which we test if both the error for duplicated validators and duplicated organizations are thrown.

pallets/stellar-relay/src/lib.rs
```diff
@@ -457,23 +459,74 @@ pub mod pallet {
...
+					.and_modify(|e| {
+						*e += 1;
+					})
+					.or_insert(1);
+			}
+
+			ensure!(
+				organizations.len() == organization_id_set.len(),
+				Error::<T>::DuplicationOrganizationId
+			);
```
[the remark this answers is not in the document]
cr4pt0: Done.

pallets/stellar-relay/src/lib.rs
```diff
@@ -117,6 +117,8 @@ pub mod pallet {
 		TransactionSetHashCreationFailed,
 		TransactionSetHashMismatch,
 		ValidatorLimitExceeded,
+		DuplicationOrganizationId,
+		DuplicationValidatorPublicKey,
```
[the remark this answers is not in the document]
cr4pt0: Done

pallets/stellar-relay/src/lib.rs
```diff
@@ -457,23 +459,74 @@ pub mod pallet {
...
 				BoundedVec::<ValidatorOf<T>, T::ValidatorLimit>::try_from(validators)
 					.map_err(|_| Error::<T>::BoundedVecCreationFailed)?;
-			Validators::<T>::put(new_validator_vec);
 
 			let new_organization_vec =
 				BoundedVec::<OrganizationOf<T>, T::OrganizationLimit>::try_from(organizations)
 					.map_err(|_| Error::<T>::BoundedVecCreationFailed)?;
-			Organizations::<T>::put(new_organization_vec);
+
+			//update only when new organization or validators not equal to old organization or
```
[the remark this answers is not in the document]
cr4pt0: Done

pallets/stellar-relay/src/lib.rs
```diff
@@ -457,23 +459,74 @@ pub mod pallet {
...
+					.and_modify(|e| {
+						*e += 1;
+					})
+					.or_insert(1);
+			}
+
+			ensure!(
+				validators.len() == validators_public_key_set.len(),
+				Error::<T>::DuplicationValidatorPublicKey
+			);
```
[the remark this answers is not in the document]
cr4pt0: Done.

pallets/stellar-relay/src/tests.rs
cr4pt0: Agree. Done

pallets/stellar-relay/src/tests.rs
cr4pt0: I want explicitly specify the public key for validator.

pallets/stellar-relay/src/tests.rs
cr4pt0: Validator and validator 1 has a different pub key.
test should fail and it works as expected.

Updated the test to specify more explicit public key with using 1 and 2 digit to fill the vector.

cr4pt0:
> I requested some changes.
> 
> It would also be nice to have a new test function `update_tier_1_validator_set_fails_for_duplicate_values` in which we test if both the error for duplicated validators and duplicated organizations are thrown.

Thanks.  Good catch ! Done!

ebma (changes requested):
The CI is failing because of the benchmarks. Can you look into that and also re-run the benchmark afterwards to re-generate the default weights?

cr4pt0:
> The CI is failing because of the benchmarks. Can you look into that and also re-run the benchmark afterwards to re-generate the default weights?

Done
@ebma

pallets/stellar-relay/src/lib.rs
```diff
@@ -457,23 +459,74 @@ pub mod pallet {
...
 			let current_organizations = Organizations::<T>::get();
-			OldOrganizations::<T>::put(current_organizations);
+			// Filter organizations for selected network type
+			let current_organizations = current_organizations.into_iter().collect::<Vec<_>>();
+
+			let current_organizations =
+				BoundedVec::<OrganizationOf<T>, T::OrganizationLimit>::try_from(
+					current_organizations,
+				)
+				.map_err(|_| Error::<T>::BoundedVecCreationFailed)?;
```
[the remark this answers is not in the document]
ebma: Okay but even if you need them below, we can just clone them and don't need to do all these conversions, do we? `Validators::<T>::get()` and `Validators::<T>::get()` returns a bounded vec already, so just using 
```
let current_validators = Validators::<T>::get();
let current_organizations = Organizations::<T>::get();
```
should be enough I guess?

ebma (changes requested):
@RustNinja I looked into it again and it looks good overall. I added another comment because I think we can still simplify the code in that one place. Besides that you just need to resolve the merge conflicts and I guess it would be ready for merge

pallets/stellar-relay/src/lib.rs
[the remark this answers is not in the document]
RustNinja: Totally agree! Done

ebma (changes requested):
Now we just need to resolve the clippy complaints that are shown in CI @RustNinja 😅

ebma (approved):

ebma:
Merging this because the ubuntu CI passed and the macOS CI probably did not because the integration test was running in another job as well.

PR #268, opened 2023-02-17
Add support for fiat currency prices

ebma:
These changes make it possible for the offchain-worker to feed fiat prices to the on-chain storage.

Changes the currency ID of the griefing collateral to the relay chain token because we do not have the price information for our native asset at the moment.


cr4pt0 (approved):

pallets/oracle/src/dia.rs
```diff
@@ -30,8 +48,16 @@ impl Convert<OracleKey, Option<(Vec<u8>, Vec<u8>)>> for DiaOracleKeyConvertor {
...
 				CurrencyId::Native => unimplemented!(),
-				CurrencyId::StellarNative => unimplemented!(),
-				CurrencyId::Stellar(primitives::Asset::AlphaNum4 { .. }) => unimplemented!(),
+				CurrencyId::StellarNative => Some((
+					STELLAR_DIA_BLOCKCHAIN.as_bytes().to_vec(),
+					STELLAR_DIA_SYMBOL.as_bytes().to_vec(),
+				)),
+				CurrencyId::Stellar(primitives::Asset::AlphaNum4 { code, .. }) => {
+					let fiat_quote =
+						construct_fiat_usd_symbol_for_currency(code.to_ascii_uppercase());
```
b-yap, line 57: i guess the `to_ascii_uppercase()` call is not necessary, since the function itself will do it too.

testchain/node/src/service.rs
```diff
@@ -223,6 +224,17 @@ pub fn new_full(mut config: Configuration) -> Result<(TaskManager, RpcHandlers),
...
+		// Initialize seed for signing transaction using off-chain workers. This is a convenience
+		// so learners can see the transactions submitted simply running the node.
+		// Typically these keys should be inserted with RPC calls to `author_insertKey`.
+		let keystore = keystore_container.sync_keystore();
+		// Note that the `KeyTypeId` here is very important and the content must match the one
+		// defined in the DIA oracle pallet. Otherwise the signer key is not available for the
+		// off-chain worker of the pallet.
+		pub const KEY_TYPE: KeyTypeId = KeyTypeId(*b"dia!");
+		sp_keystore::SyncCryptoStore::sr25519_generate_new(&*keystore, KEY_TYPE, Some("//Bob"))
+			.expect("Creating key with account Alice should succeed.");
```
b-yap, line 236: Looks like it's Bob, not Alice. :joy: But it does't matter, really.

pallets/oracle/src/dia.rs
b-yap, line 57: `to_ascii_uppercase()` is also performed in the function itself. Probably you can just pass the `code` directly.

b-yap (changes requested):
just everything minor

pallets/oracle/src/dia.rs
ebma: Well actually it did only convert it to uppercase for the comparison but not when using it in the concatenation. Which is weird behavior so I moved it around a bit and added a comment 😅

b-yap (approved):

PR #276, opened 2023-02-23
274 Improve handling of issue request execution

b-yap:
I introduced 1 new structure and enum: `LedgerTask` and `LedgerTaskStatus` respectively.

```rust
pub enum LedgerTaskStatus {
	NotStarted,
	ProcessSuccess,
	RecoverableError,
	Error(String),
}

pub struct LedgerTask {
	ledger: Ledger,
	task_execution_receiver: Receiver<LedgerTaskStatus>,
	retries_remaining: u8,
	latest_status: LedgerTaskStatus,
}
```
These should be sufficient enough to make sure that a new task is only run IF and only IF the task failed, and this failure is recoverable.


cr4pt0 (approved):

clients/wallet/src/lib.rs
```diff
@@ -3,11 +3,13 @@ use substrate_stellar_sdk::TransactionEnvelope;
...
 pub use stellar_wallet::StellarWallet;
+pub use task::*;
 
 pub mod error;
 mod horizon;
 mod stellar_wallet;
+mod task;
 pub mod types;
 
 pub type Ledger = u32;
```
ebma, line 14: I think the lingo is a bit confusing. If I'm not mistaken a `Ledger` is equal to a `Slot`. While 'Ledger' is used to describe a block in the Stellar world, the term 'slot' is rather used in the SCP consensus paper etc. But at the end of the day it boils down to the same thing when it comes to the number we store for a ledger and a slot. For one transaction `x` the slot and ledger will be equal to each other. 

Thus, I would prefer if we find a better way for describing this. If I remember correctly we already have a type for `Slot`, maybe we could just re-use that one. (I know that this is not something you introduced in this PR specifically, it just became a bit more apparent now that we use that term more often.)

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+pub const LEDGER_TASK_MAX_RETRIES: u8 = 3;
+/// Determines the status of the task of processing a transaction
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum LedgerTaskStatus {
+	NotStarted,
+	/// Process Done and it was a success
+	ProcessSuccess,
+	/// The error is acceptable, and is best to just retry again
+	RecoverableError,
+	/// Something happened when
```
ebma, line 13: Seems like words are missing

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+
+	/// the number of retries allowed to process this specific task
+	retries_remaining: u8,
+
+	/// The latest status of this task
+	latest_status: LedgerTaskStatus,
+}
+
+impl LedgerTask {
+	/// Creates a task structure the `TransactionEnvelope`
```
ebma, line 31: Not sure if a word is missing here

	/// Creates a task structure **for** the `TransactionEnvelope`

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+			_ => false,
+		}
+	}
+
+	pub fn ledger(&self) -> Ledger {
+		self.ledger
+	}
+
+	/// Checks the status of the task
+	pub fn status(&mut self) -> LedgerTaskStatus {
```
ebma, line 62: If this function is able to mutate the state of the struct itself we should maybe change the name to `update_status`

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+			task_execution_receiver,
+			// The decrement action is called everytime a new receiver is set.
+			retries_remaining: LEDGER_TASK_MAX_RETRIES,
+			latest_status: LedgerTaskStatus::NotStarted,
+		}
+	}
+
+	/// Sets a new one shot receiver.
+	/// Returns true if it was successful.
+	pub fn change_receiver(&mut self, receiver: Receiver<LedgerTaskStatus>) -> bool {
```
ebma, line 44: Maybe change the name to something like `try_recover_with_new_receiver()` or something else. With the current name, it's not obvious why I would call this function.

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+	/// Process Done and it was a success
+	ProcessSuccess,
+	/// The error is acceptable, and is best to just retry again
+	RecoverableError,
+	/// Something happened when
+	Error(String),
+}
+
+pub struct LedgerTask {
+	ledger: Ledger,
```
ebma, line 18: I think this implementation is so general that in theory, we could generify it so that it is not specific to just 'ledgers' but to anything.  You implemented some struct that has some kind of ID (the `ledger` field) and is able to hold info about its current status, and can retry some number of times. 
If we generify the ID passing we could also use this in other contexts. Counter-argument would be that we over-engineer this while we don't need to, but since it's little extra work I would propose we add those changes too. WDYT @b-yap ?

ebma: In any case, I would like to rename this. The name would be different if we generify, however.

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
+use crate::Ledger;
+use tokio::sync::oneshot::{channel, error::TryRecvError, Receiver, Sender};
+
+pub const LEDGER_TASK_MAX_RETRIES: u8 = 3;
```
ebma, line 4: We should consider passing this constant to the constructor function instead of always using it

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+		let ledger = 10;
+		let (sender, receiver) = channel();
+
+		let task = LedgerTask::create(ledger, receiver);
+
+		assert_eq!(task.ledger, ledger);
+		assert_eq!(task.retries_remaining, LEDGER_TASK_MAX_RETRIES);
+		assert_eq!(task.latest_status, LedgerTaskStatus::NotStarted);
+
+		drop(sender);
```
ebma, line 120: Do we need to drop?

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+			assert_eq!(task.status(), LedgerTaskStatus::ProcessSuccess);
+		}
+		{
+			let (sender, receiver) = channel();
+			let mut task = LedgerTask::create(20, receiver);
+
+			sender
+				.send(LedgerTaskStatus::Error("this is a test".to_string()))
+				.expect("should be able to send status");
+			assert_eq!(task.status(), LedgerTaskStatus::Error("this is a test".to_string()));
```
ebma, line 141: Minor nit: We could assign the `"this is a test".to_string()` to a variable and use that variable in both the send and the assert statement.

clients/wallet/src/task.rs
```diff
@@ -0,0 +1,206 @@
...
+		// For tasks flagged as `NotStarted` , wait for it.
+		// For tasks flagged as `ProcessSuccess`, they will be removed in the for loop
+		// For tasks flagged as `Error`, these will be removed.
+		// continue with the loop
+		None
+	}
+}
+
+#[cfg(test)]
+mod test {
```
ebma, line 106: It would be nice to have a test case that shows what happens once the maximum number of retries is exhausted and the 'recover' function is called.

ebma (changes requested):
I like the changes overall. I added my thoughts on some specifics and would like to have your feedback on this @b-yap

clients/wallet/src/lib.rs
b-yap: created a [new issue](https://github.com/pendulum-chain/spacewalk/issues/277) and let's continue from there

clients/vault/src/issue.rs
```diff
@@ -159,6 +159,75 @@ fn get_issue_id_from_tx_env(tx_env: &TransactionEnvelope) -> Option<IssueId> {
...
+				tracing::error!("{}",e);
+				// we cannot process this again, so remove it from the list.
+				ledger_map.remove(ledger);
+				false
+			}
+
+		}
+	});
+
+	drop(ledger_map);
```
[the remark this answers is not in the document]
b-yap: removed

clients/wallet/src/task.rs
b-yap: updated, 🙇

clients/wallet/src/task.rs
b-yap: done

clients/wallet/src/task.rs
b-yap: updated to `fn set_receiver()` and **_private_**; only the method `fn recover_with_new_sender()` calls it.

clients/wallet/src/task.rs
b-yap: i think it doesn't hurt to have a default value, so I changed it to  `DEFAULT_MAX_RETRIES` and moved it inside the struct itself, to avoid confusion. 
However, i did add a new method: `fn new_with_max_retries(...)` with 3 params; the last param being the `max_retries`

clients/wallet/src/task.rs
b-yap: removed

clients/wallet/src/task.rs
b-yap: done

clients/wallet/src/task.rs
b-yap: check test case `max_retries_exhausted()`

clients/wallet/src/task.rs
b-yap: I just changed it to `SlotTask` because there's no alternate name provided. Do you have any suggestions (even just a temp one would be fine  :v:)?

And yes, creating a generic out of this struct  did cross my mind, but I cannot find yet on applying it elsewhere.
Should we create a ticket and explore the "possibilities" from there?  :clap: :grinning:

clients/wallet/src/task.rs
ebma: That's a great idea, thank you 👍

clients/vault/src/issue.rs
```diff
@@ -159,6 +163,74 @@ fn get_issue_id_from_tx_env(tx_env: &TransactionEnvelope) -> Option<IssueId> {
...
+	}
+
+	// Not finding the ledger in the map means there's no existing task for this ledger.
+	let (sender, receiver) = tokio::sync::oneshot::channel();
+	tracing::trace!("Creating a task for ledger {}", ledger);
+
+	let ledger_task = SlotTask::new(*ledger, receiver);
+	processed_map.insert(*ledger, ledger_task);
+
+	Some(sender)
```
ebma, line 188: this is now a bit inconsistent with 'ledger' vs 'slot'
you should probably update the comment and variable name too 😅

cr4pt0 (approved):

web-30 (approved):

ebma:
I ran the tests locally and they work, so I'm merging this

Issue #282, opened 2023-03-01
Remove usages of `unwrap` and `expect` from spacewalk pallets


ebma:
We need to remove the usages of these two functions from code that is executed in the runtime (ie. the spacewalk pallets) because substrate runtimes should never panic.

We should add proper error handling with match statements instead.

PR #287, opened 2023-03-03
282 remove usages of unwrap and expect from spacewalk pallets

RustNinja:
update spacewalk's pallet.
replace `unwrap` and `expect` functions with proper handling.
Now there are nothing that will panic in runtime
issue #282


pallets/stellar-relay/src/lib.rs
```diff
@@ -117,6 +117,7 @@ pub mod pallet {
 		TransactionSetHashCreationFailed,
 		TransactionSetHashMismatch,
 		ValidatorLimitExceeded,
+		FailedToComputenonGenericTxSetContentHash,
```
ebma, line 120: Nit: The 'n' for non should be capitalized
```
		FailedToComputeNonGenericTxSetContentHash,
```

pallets/stellar-relay/src/testing_utils.rs
```diff
@@ -197,7 +197,7 @@ pub fn build_dummy_proof_for<T: crate::Config>(
 	txes.push(transaction_envelope.clone()).unwrap();
 	let transaction_set = TransactionSet { previous_ledger_hash: Hash::default(), txes };
 
-	let tx_set_hash = crate::compute_non_generic_tx_set_content_hash(&transaction_set);
```
ebma, line 200: Don't we need to handle this now? You changed it to return an `Option` now, but the `create_scp_envelope` function that consumes it a few lines below still expects the real type and not an option I think.

pallets/stellar-relay/src/tests.rs
```diff
@@ -105,7 +105,7 @@ fn create_valid_dummy_scp_envelopes(
 	txes.push(transaction_envelope.clone()).unwrap();
 	let transaction_set = TransactionSet { previous_ledger_hash: Hash::default(), txes };
 
-	let tx_set_hash = crate::compute_non_generic_tx_set_content_hash(&transaction_set);
```
ebma, line 108: Same here, I think we need to use a `match` or something

ebma (commented):
Maybe I'm wrong about the changes I point out. The tests seem to be working(?)

pallets/stellar-relay/src/lib.rs
cr4pt0: Done.

pallets/stellar-relay/src/testing_utils.rs
cr4pt0: I added `unwrap` method but now i replace it with 
`.expect("Should compute non generic tx set content hash");`

pallets/stellar-relay/src/tests.rs
cr4pt0: I added `unwrap` method but now i replace it with 
`.expect("Should compute non generic tx set content hash");`

PR #294, opened 2023-03-09
292 add extrinsics to change the values of the minimum transfer amounts

cr4pt0:
- introduce `minimum_transfer_amount_update` to `issue`, `replace` and `redeem` pallets
- it allow update minimum amount value in StorageMap using root


ebma (approved):
LGTM

PR #299, opened 2023-03-13
298 Make vault delay the check for pending Stellar transactions for open requests

b-yap:
From @ebma : 
> I think just adding a delay here of let's say 12 seconds before querying the stellar transactions of the account should suffice to fix it. Because I think actually we have everything in place to prevent this bug in the execute_open_requests() function but I guess that the timing was so unlucky that the transaction was not yet completely processed/accepted by the Stellar network so that fetching the transactions of the vault account did not include this transaction yet, which is why the vault client assumed it needed to do another payment.


RustNinja (approved):

ebma:
We don't need this anymore due to #319.

PR #302, opened 2023-03-14
301 trim stellar secret key before using it in vault client

### LINKED ISSUE #301: Trim Stellar secret key before using it in vault client
We should strip whitespace and new line characters from the Stellar secret key before using it.

ebma:
This change is necessary to fix the vaults cloud deployment.

Closes #301.


RustNinja (approved):

b-yap (approved):
I reran the failed test and it went ok. :facepunch:

PR #318, opened 2023-03-21
317 add xlm to vault client auto register

### LINKED ISSUE #317: Add XLM to Vault Client auto-register
Change the auto-register format such that a vault operator can use it to automatically register for any Stellar asset, including the native asset

ebma:
This changes the parsing of the `auto-register` config parameter such that `XLM` can also be used as a wrapped asset.

Closes #317.

### DIFF
--- a/clients/runtime/src/assets.rs
+++ b/clients/runtime/src/assets.rs
@@ -8,8 +8,17 @@ pub trait TryFromSymbol: Sized {
 }
 
 impl TryFromSymbol for CurrencyId {
+	/// This can build a currency from a string in the following formats:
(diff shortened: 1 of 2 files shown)


RustNinja (approved):
there is example in README.md
`--stellar-vault-secret-key` instead of `--stellar-vault-secret-key-filepath`

that throw the error:
error: Found argument '--stellar-vault-secret-key' which wasn't expected, or isn't valid in this context

        Did you mean '--stellar-vault-secret-key-filepath'?

TorstenStueber (approved):

Issue #315, opened 2023-03-21
Retry Stellar transactions indefinitely


ebma:
We recently changed the logic in the `execute_open_requests()` function to retry up to some specific number of times (ie `3`). Now, while this improves our resilience to some network delays, we should actually make the vault client retry the submission indefinitely because it might happen that an attacker manages to spam/delay the vault's account for a longer timeframe, thus making the limited number of retries problematic.
I would suggest that we extend the [task](https://github.com/pendulum-chain/spacewalk/blob/main/clients/wallet/src/task.rs) implementation to also be able to retry indefinitely, (as long as recoverable errors are encountered).

ebma:
But an estimate of 21 would mean that this task is so huge that it should be split up into multiple tasks 😅 . I'll invite for planning poker

b-yap:
okay, so maybe it's an 8. I'll update my estimate.

Currently, I'm listing down what "errors" are deemed acceptable and what are "not". And how do we communicate it back to the user. Especially when resubmitting stellar transactions that were in the cache, how do we let the user know that a few of them failed? 

For recoverable errors, let's say (stellar transactions) tx1, tx2, ... until tx10: if at tx6 it fails repeatedly (do you think it can happen, @ebma , @TorstenStueber ?) that means tx7 and beyond will have to wait indefinitely too.

b-yap:
```rust
pub enum Error {
	#[error("Invalid secret key")]
	InvalidSecretKey,
	#[error("Error fetching horizon data: {0}")]
	HttpFetchingError(#[from] FetchError),
	#[error("Could not build transaction: {0}")]
	BuildTransactionError(String),
	#[error("Transaction submission failed. Title: {title}, Status: {status}, Reason: {reason}, Envelope XDR: {envelope_xdr:?}")]
	HorizonSubmissionError {
		title: String,
		status: u16,
		reason: String,
		envelope_xdr: Option<String>,
	},
	#[error("Could not parse string: {0}")]
	Utf8Error(#[from] std::str::Utf8Error),
	#[error("Could not decode XDR")]
	DecodeError,
	#[error("Could not sign envelope")]
	SignEnvelopeError,

	#[error(transparent)]
	CacheError(CacheError),
}
```
This is the error returned from `submit_transaction(...)` 
@ebma 

[...]

PR #319, opened 2023-03-22
Cache transactions before submitting to Stellar Node

b-yap:
Closes https://github.com/pendulum-chain/tasks/issues/8

1. A `TransactionEnvelope` is saved locally **before** submitting to Stellar. It will be removed afterwards, regardless of the response.
    - Examples are found in the **_`resources`_** folder of the wallet crate. The content is a `Vec<u8>` xdr format. The filename is the transaction's **_sequence number_**.
2. To extract the _sequence number_, I added a new function on the `TransactionEnvelopeExt` trait in `spacewalk-primitives` crate: **_`fn sequence_number(&self) -> Option<SequenceNumber>`_**
3. When the vault starts/restarts — assuming the location of the cache remains unchanged — all files found in the cache will be resubmitted (ascending order by sequence number). 
   - This happens **before** calling the `fn execute_open_requests(...)`. (See **system.rs**).
4. The caching and resubmission operations are found in `StellarWallet`. 
5. Added extra errors dedicated for Caching.
  - redesigned `HorizonSubmissionError` to expose the timeout factor.

----

### How to begin the review:
1. Look inside **cache.rs** for the new struct:
    -  ```rust
       #[derive(Debug, Clone)]
       pub struct TxEnvelopeStorage {
	      path: String,
	      inner_path: String,
        }
        ``` 
        The `path` is the root directory and the `inner_path` is the inner(actual) directory where the files will be saved. Combined, the _full path_ would be `<path>/<inner_path>` as defined in `fn full_path(&self)`.
    -  Some operations like the delete/**_remove_** are cfg/feature gated depending on the need:
       -  Automatic deletion is disabled during unit testing because the files are used in other test cases.
       -  Removing the entire directory is ONLY for testing purposes (see `pub fn remove_all(&self)`), since we need the cache to still exist, should the vault goes down. 

2. Inside **stellar_wallet.rs**:
    -  Added  the `cache` field to `StellarWallet`:
       ```rust
        pub struct StellarWallet {
             ... 
	    /// Used for caching Stellar transactions before they get submitted.
	    cache: TxEnvelopeStorage,
        }
        ```
        
     - To initialize the `cache` field, a path must be provided by the user. This is done through the updated:
         ```rust
         from_secret_encoded_with_cache(
		    ...
		    cache_path: String,
	    ) -> Result<Self, Error> {
	    ...
         ```
     - The resubmission method is found in `StellarWallet`:
         ```rust 
              /// Submits transactions found in the wallet's cache to Stellar.
	     /// Returns a tuple: (list of txs successfully submitted, list of errors of txs not resubmitted)
	    pub async fn resubmit_transactions_from_cache(
		    &mut self,
	    ) -> (Vec<(TransactionResponse>, Vec<Error>) {
	    ...
         ```
     - Extra step in `async fn send_payment_to_address(...)` method of `StellarWallet` is calling the `save_to_local(...)` method of the cache, to save the envelope in a local path.
     - A separate private method is where the transactions will be submitted, and eventually deleted:
       ```rust
       async fn submit_transaction(
		  &mut self,
		 ... 
	    ) -> Result<(TransactionResponse, TransactionEnvelope), Error> {
	    ...
        ```
        It will be called by both the  public methods `resubmit_transaction_from_cache(...)` and `send_payment_to_address(...)`.
     -  Added 1 test case: **`resubmit_transactions_works()`** to test the new method.  
3. Inside error.rs:
   - redesign of one variant:
       ```rust
       #[error("Transaction submission failed. Title: {title}, Status: {status}, Reason: {reason}, Envelope XDR: {envelope_xdr}")]
	    HorizonSubmissionError { title: String, status: u16, reason: String, envelope_xdr: String },
       ```
       It will show the _timeout_ response, which is found in the [response's status code](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/timeout).
   - added more errors dedicated for caching.  
5. In the tests of the `vault` crate:
   - Because of how the `StellarWallet` is initialized, maintaining  the `lazy_static!` style became difficult. Instead, I have them declared in the testing functions:
    ```rust
      ...
      - async fn test_with<F, R>(execute: impl FnOnce(SubxtClient) -> F) -> R
      + async fn test_with<F, R>(execute: impl FnOnce(SubxtClient, ArcRwLock<StellarWallet>) -> F) -> R
      ...
      
      - execute: impl FnOnce(SubxtClient, VaultId, SpacewalkParachain) -> F,
      + execute: impl FnOnce(SubxtClient, ArcRwLock<StellarWallet>, VaultId, SpacewalkParachain) -> F,
    ```
  - The **clients/runtime/integration/mod.rs** (see the `default_provider_client(...)`) luckily provided the _`tempdir`_, which I used to define the cache location. 
          
------

Things left to do:
- [x] delete the file after a response is returned from the Stellar Node (or even when it returns an error)
- [x] expose the delete method outside of the wallet crate
- [x] update the vault with the new code
- [x] ~~discuss how to forcefully restart the vault in the integ test cases~~ _maybe the resubmission **unit test** will suffice_.
- [x] _discuss the exact location to store these caches_. See (https://github.com/pendulum-chain/spacewalk/pull/319#discussion_r1147829364)
- [x] _discuss where should the location ^ be defined. Right now it's an additional param in `StellarWallet`'s `fn from_secret_encoded(...)`. Doesn't look ideal. Need opinion. What about a hardcoded value?_ See (https://github.com/pendulum-chain/spacewalk/pull/319#discussion_r1147829364)


b-yap:
there's 1 PR with a [running CI](https://github.com/pendulum-chain/spacewalk/actions/runs/4509612741/jobs/7939582112?pr=269). i'll re-run once it finishes.

clients/wallet/src/cache.rs
```diff
@@ -0,0 +1,344 @@
...
+
+		let mut res: Vec<(TransactionEnvelope, SequenceNumber)> = directory
+			.into_iter()
+			.filter_map(|entry| {
+				let path = entry.ok()?.path();
+				extract_tx_envelope_from_path(path)
+			})
+			.collect();
+
+		// the envelopes have been filtered out with those having sequence numbers already
```
ebma, line 137: I don't understand what this comment means. What was filtered out? I get that in the next line we order them in ascending order but I guess the comment is unrelated to that?

clients/wallet/src/cache.rs
```diff
@@ -0,0 +1,344 @@
...
+		res.sort_by(|(_, first), (_, next)| first.cmp(&next));
+
+		res
+	}
+}
+
+/// a helper function to convert a String content into `Vec<u8>`
+fn parse_string_to_vec_u8(value: &str) -> Option<Vec<u8>> {
+	let remove_white_space = value.replace(' ', "");
```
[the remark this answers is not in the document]
ebma: I am not sure I understand what this function is supposed to do. Maybe we need a better name or a comment. Because if it's only about converting a string to a `Vec<u8>` I guess we could use `value.as_bytes().to_vec()` 😅

clients/wallet/src/stellar_wallet.rs
```diff
@@ -25,20 +27,53 @@ pub struct StellarWallet {
...
+		let _ = self.cache.save_to_local(sequence, envelope.clone());
+
+		let client_result = horizon_client
+			.submit_transaction(envelope.clone(), self.is_public_network)
+			.await
+			.map(|response| (response, envelope));
+
+		// for unit testing purposes, file won't be removed immediately.
+		#[cfg(not(test))]
+		let _ = self.cache.remove_transaction(sequence);
```
ebma, line 50: We'd have to check the result of the submission before removing it from the cache. If for example, it's a timeout error, we would need to keep this transaction and retry it. If however, it's eg. a tx_bad_seq error, we can discard the transaction basically. 

Or we change the `resubmit_transactions_from_cache()` method to return errors + the transaction envelope for which the error occurred, so that the transaction envelope is not lost overall, and the caller of the method can decide how to continue.

clients/wallet/src/stellar_wallet.rs
```diff
@@ -180,19 +244,21 @@ mod test {
...
+	fn wallet() -> Arc<RwLock<StellarWallet>> {
+		Arc::new(RwLock::new(
 			StellarWallet::from_secret_encoded(
 				&STELLAR_VAULT_SECRET_KEY.to_string(),
 				IS_PUBLIC_NETWORK,
+				"resources/test_2".to_string(),
 			)
-			.unwrap()
-		));
+			.unwrap(),
```
ebma, line 254: Is the ',' necessary?

clients/wallet/src/cache.rs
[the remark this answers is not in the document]
ebma: Okay looking at the files in the resources I think I get what the problem is you are trying to solve. But then we need to change the name to make it explicit what we we want to parse from a file. I'd rename it to something like `parse_xdr_file_content_to_vec()` and add a descriptive comment that this is necessary because space characters are written to the file when saving.

ebma (changes requested):
I think it's really great how you implemented it. I added some comments with how I think we could improve the logic.

clients/wallet/src/cache.rs
```diff
@@ -0,0 +1,344 @@
...
+		tracing::info!("path for caching stellar transactions: {full_path}");
+
+		TxEnvelopeStorage { path, inner_path }
+	}
+
+	/// Returns true if transaction was saved to a local path.
+	/// The filename will be the transaction's sequence number.
+	pub fn save_to_local(
+		&self,
+		sequence: SequenceNumber,
```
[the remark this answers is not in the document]
b-yap: You are definitely right; I forgot to remove them all, after adding the _`fn sequence_number`_ of `TransactionEnvelopeExt`.

sequence param removed.

clients/wallet/src/cache.rs
```diff
@@ -0,0 +1,344 @@
...
+		transaction: TransactionEnvelope,
+	) -> bool {
```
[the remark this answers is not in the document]
b-yap: added a few errors dedicated for the cache. :bow:

clients/wallet/src/cache.rs
```diff
@@ -0,0 +1,344 @@
...
+		let path = Path::new(&full_file_path);
+
+		if !path.exists() {
+			return None
+		}
+		read_tx_envelope_from_path(path).map(|(tx, _)| tx)
+	}
+
+	/// Returns all transactions found in local path
+	pub fn get_tx_envelopes(&self) -> Vec<(TransactionEnvelope, SequenceNumber)> {
```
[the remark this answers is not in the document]
b-yap: done

clients/wallet/src/cache.rs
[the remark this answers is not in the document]
b-yap: renamed to **_`parse_xdr_string_to_vec`_**

clients/wallet/src/cache.rs
```diff
@@ -0,0 +1,344 @@
...
+				format!("Invalid xdr string: {num_as_char} at pos: {pos}")
+			))
+		})
+		.collect()
+}
+
+/// Returns a tuple of `TransactionEnvelope` and its sequence number,
+/// if successfully extracted from the path.
+/// Otherwise returns a None and and the file will be deleted.
+fn extract_tx_envelope_from_path<P: AsRef<Path> + std::fmt::Debug + Clone>(
```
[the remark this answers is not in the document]
b-yap: removed

clients/wallet/src/cache.rs
```diff
@@ -0,0 +1,344 @@
...
+
+	#[doc(hidden)]
+	#[cfg(any(test, feature = "testing-utils"))]
+	/// Removes all transactions, including the directory
+	pub fn remove_all(&self) -> bool {
+		std::fs::remove_dir_all(self.full_path()).is_ok()
+	}
+
+	#[allow(dead_code)]
+	#[cfg(any(test, feature = "testing-utils"))]
```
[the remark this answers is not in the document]
b-yap: It's useful for testing, but highly unlikely that we want to access one single envelope for other purpose, aside from resubmitting it. And when we resubmit, it's all or nothing. 

It's nice to have tbh, but it's also a dead code since no actual logic is calling this method (again, aside from testing).

clients/wallet/src/stellar_wallet.rs
b-yap: Did you mean Stellar's [timeout](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/timeout) response? 
I've updated the `HorizonSubmissionError` from 1 param, into a struct with 4 fields. One of the fields is `envelope_xdr`.  The user can deal with it as he like.

clients/wallet/src/horizon.rs
```diff
@@ -16,6 +16,60 @@ pub type Ledger = u32;
...
+					}
+				},
+				_ => {
+					let detail = resp["detail"].as_str().unwrap_or(unknown);
+
+					Error::HorizonSubmissionError {
+						title: title.to_string(),
+						status,
+						reason: detail.to_string(),
+						envelope_xdr: "".to_string(),
```
ebma, line 63: Maybe we should use an `Option<>` type for this if it can be empty?

ebma: No wait, actually this is important because the items of the error object will be used for resubmission eventually. So let's also make this an `Option<String>` for the error if possible.

clients/wallet/src/error.rs
```diff
@@ -8,16 +9,45 @@ pub enum Error {
...
+	FileDoesNotExist(String),
+
+	#[error("Cache Error: Cannot read file: {0:?}")]
+	ReadFileFailed(String),
+
+	#[error("Cache Error: Invalid file: {0:?}")]
+	InvalidFile(String),
+
+	#[error("Cache Error: Cannot decode xdr format for file: {0:?}")]
+	DecodeFileFailed(String),
```
ebma, line 52: Nit: If all of these are 'Cache Errors' we could actually group them under a new sub enum called `CacheError` which makes it apparent that these are Cache errors without having to mention it in the error message. But this is up to you if you think it's better to change it that way, I don't have hard feelings about this @b-yap.

clients/vault/src/system.rs
```diff
@@ -437,7 +437,8 @@ impl VaultService {
 		let is_public_network = wallet.is_public_network();
 
 		// re-submit transactions in the cache
-		wallet.resubmit_transactions_from_cache().await;
+		// todo: handle timeouts
```
RustNinja, line 440: @b-yap 
do you still plan to handle somehow timeouts?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -80,14 +166,69 @@ impl StellarWallet {
...
+
+		let (envs, mut errors) = match self.cache.get_tx_envelopes() {
+			Ok(x) => x,
+			Err(errors) => return (passed, errors),
+		};
+
+		for env in envs.into_iter() {
+			match self.submit_transaction(env, &horizon_client).await {
+				Ok(response) => passed.push(response),
+				Err(e) => errors.push(e),
```
RustNinja, line 187: @b-yap 
are we able to add logs if `submit_transaction` failed? 
For now there is no logging and the value returned from `resubmit_transactions_from_cache` not handling at all.
`let _ = wallet.resubmit_transactions_from_cache().await;`  (line 441)

clients/vault/src/system.rs
b-yap: good question. I'm not sure if it should be done in this PR, because there's 2 ways for this, [as Stellar suggested](https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/timeout):
> 1. Keep resubmitting the same transaction (with the same sequence number) and wait until it finally is added to a new ledger, or
> 2. Increase the fee in order to prioritize the transaction.

If we choose the first one, next question would be: how many times should we keep resubmitting? What to do if it's really down? rewrite to cache again and simply crash the vault? Or go to the 2nd option?

If we choose the 2nd one, that means updating the envelope itself. We should alert the user about this. 

Maybe I'm overthinking and I need someone else's input. :neckbeard: @ebma

clients/wallet/src/error.rs
b-yap: I started with a separate `CacheErrors` at first, but then there's a `Result<_,`**_`Vec<Error>`_**`>` return type, which means iterating over the list of `CacheError`s and converting them to `Error`.

Instead, I have the `CacheError(` variant as you suggested originally, and I introduced a struct **`CacheError`** with its specific _ErrorKind_, the **_`CacheErrorKind`_**. 
I wanted to clean up its display, hence the `#[error(transparent)]`.

clients/wallet/src/horizon.rs
b-yap: done

clients/wallet/src/error.rs
ebma: The new implementation is great, good job 👍

clients/vault/src/system.rs
```diff
@@ -431,9 +432,17 @@ impl VaultService {
...
 
-		let wallet = self.stellar_wallet.read().await;
+		let mut wallet = self.stellar_wallet.write().await;
 		let vault_public_key = wallet.get_public_key();
 		let is_public_network = wallet.is_public_network();
+
+		// re-submit transactions in the cache
+		let (_, errors) = wallet.resubmit_transactions_from_cache().await;
+		if !errors.is_empty() {
+			// todo: handle timeouts
```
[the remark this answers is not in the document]
ebma: just saw the comment is already raised above, so I'll resolve this one.

clients/vault/src/system.rs
ebma: I agree that this should not be handled in this PR but in a separate one. I extended issue #315 to not also introduce indefinite retries but also to add this to all places where we submit Stellar transactions.

clients/wallet/src/stellar_wallet.rs
ebma: Sounds good

clients/wallet/src/cache.rs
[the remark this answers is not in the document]
ebma: Okay, fine to keep it like that then

clients/wallet/src/horizon.rs
```diff
@@ -16,6 +16,61 @@ pub type Ledger = u32;
...
+	}
+
+	let error = match response.json::<serde_json::Value>().await {
+		Err(e) => Error::HttpFetchingError(e),
+		Ok(resp) => {
+			let unknown = "unknown";
+			let title = resp["title"].as_str().unwrap_or(unknown);
+			let status = u16::try_from(resp["status"].as_u64().unwrap_or(400)).unwrap_or(400);
+
+			match title {
```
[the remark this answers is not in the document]
b-yap: they both have the same status code:
https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-failed
https://developers.stellar.org/api/errors/http-status-codes/horizon-specific/transaction-malformed. 

But I've updated the code still.

ebma (approved):
LGTM. Can be merged if CI passes.

PR #327, opened 2023-04-04
Update signature of get_required_collateral_for_wrapped RPC function

RustNinja:
- updated get_required_collateral_for_wrapped RPC function in vault-registry
- updated testchain runtime. `wrapped_currency_id` in testchain does not passed to the pallet function. use default value as  `GetWrappedCurrencyId::get()`

```
const WRAPPED_CURRENCY_ID: CurrencyId = CurrencyId::AlphaNum4(
	*b"USDC",
	[
		20, 209, 150, 49, 176, 55, 23, 217, 171, 154, 54, 110, 16, 50, 30, 226, 102, 231, 46, 199,
		108, 171, 97, 144, 240, 161, 51, 109, 72, 34, 159, 139,
	],
);
```

### DIFF
--- a/clients/runtime/src/rpc.rs
+++ b/clients/runtime/src/rpc.rs
@@ -526,7 +526,7 @@ pub trait VaultRegistryPallet {
 		&self,
 		amount_wrapped_asset: u128,
 		wrapped_currency_id: CurrencyId,
-		collateral_currency: CurrencyId,
+		collateral_currency_id: CurrencyId,
 	) -> Result<u128, Error>;
 
 	async fn get_required_collateral_for_vault(&self, vault_id: VaultId) -> Result<u128, Error>;
@@ -670,7 +670,7 @@ impl VaultRegistryPallet for SpacewalkParachain {
 		&self,
 		amount_wrapped_asset: u128,
 		wrapped_currency_id: CurrencyId,
-		collateral_currency: CurrencyId,
+		collateral_currency_id: CurrencyId,
 	) -> Result<u128, Error> {
 		let head = self.get_finalized_block_hash().await?;
 		let result: BalanceWrapper<_> = self
@@ -681,7 +681,7 @@ impl VaultRegistryPallet for SpacewalkParachain {
 				rpc_params![
 					BalanceWrapper { amount: amount_wrapped_asset },
 					wrapped_currency_id,
-					collateral_currency,
+					collateral_currency_id,
 					head
 				],
 			)
--- a/clients/runtime/src/rpc.rs
+++ b/clients/runtime/src/rpc.rs
@@ -525,6 +525,7 @@ pub trait VaultRegistryPallet {
 	async fn get_required_collateral_for_wrapped(
 		&self,
 		amount_wrapped_asset: u128,
+		wrapped_currency_id: CurrencyId,
 		collateral_currency: CurrencyId,
 	) -> Result<u128, Error>;
 
@@ -668,6 +669,7 @@ impl VaultRegistryPallet for SpacewalkParachain {
 	async fn get_required_collateral_for_wrapped(
 		&self,
 		amount_wrapped_asset: u128,
+		wrapped_currency_id: CurrencyId,
 		collateral_currency: CurrencyId,
 	) -> Result<u128, Error> {
 		let head = self.get_finalized_block_hash().await?;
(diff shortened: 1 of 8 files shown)


pallets/vault-registry/rpc/runtime-api/src/lib.rs
```diff
@@ -44,7 +44,7 @@ sp_api::decl_runtime_apis! {
 
 		/// Get the minimum amount of collateral required for the given amount of token
 		/// with the current threshold and exchange rate
-		fn get_required_collateral_for_wrapped(amount: BalanceWrapper<Balance>, currency_id: CurrencyId) -> Result<BalanceWrapper<Balance>, DispatchError>;
+		fn get_required_collateral_for_wrapped(amount_wrapped: BalanceWrapper<Balance>, wrapped_currency_id: CurrencyId,  collateral_currency_id: CurrencyId) -> Result<BalanceWrapper<Balance>, DispatchError>;
```
ebma, line 47: (minor) nit: I think there are two spaces after the comma before `collateral_currency_id` lol

testchain/runtime/src/lib.rs
```diff
@@ -907,9 +907,9 @@ impl_runtime_apis! {
 			VaultRegistry::get_collateralization_from_vault_and_collateral(vault, &amount, only_issued)
 		}
 
-		fn get_required_collateral_for_wrapped(amount_wrapped: BalanceWrapper<Balance>, currency_id: CurrencyId) -> Result<BalanceWrapper<Balance>, DispatchError> {
+		fn get_required_collateral_for_wrapped(amount_wrapped: BalanceWrapper<Balance>, _wrapped_currency_id: CurrencyId,  collateral_currency_id: CurrencyId) -> Result<BalanceWrapper<Balance>, DispatchError> {
 			let amount_wrapped = Amount::new(amount_wrapped.amount, GetWrappedCurrencyId::get());
```
ebma, line 911: This was supposed to use the passed `wrapped_currency_id` instead of `GetWrappedCurrencyId::get()`. The `GetWrappedCurrencyId` was supposed to get removed as a whole.

ebma: While `GetWrappedCurrencyId` is used in the chain spec for the default pairs, I think it would be better to move it there then as it is the only place where we really need it (assuming that you add the change to this function to use the passed `wrapped_currency_id`).

ebma: So please move the definition of the default wrapped currency ID to the `chain_spec.rs` and remove it from here.

clients/runtime/src/rpc.rs
```diff
@@ -525,6 +525,7 @@ pub trait VaultRegistryPallet {
 	async fn get_required_collateral_for_wrapped(
 		&self,
 		amount_wrapped_asset: u128,
+		wrapped_currency_id: CurrencyId,
 		collateral_currency: CurrencyId,
```
ebma, line 529: Nit: I'd rename `collateral_currency` to `collateral_currency_id` to make it consistent.

clients/runtime/src/rpc.rs
```diff
@@ -668,6 +669,7 @@ impl VaultRegistryPallet for SpacewalkParachain {
...
```
ebma, line 673: Same here

ebma (changes requested):

clients/runtime/src/rpc.rs
RustNinja: Done

clients/runtime/src/rpc.rs
RustNinja: Done

pallets/vault-registry/rpc/runtime-api/src/lib.rs
RustNinja: Done

testchain/runtime/src/lib.rs
RustNinja: Done.

RustNinja:
PR with updates from request changes : https://github.com/pendulum-chain/spacewalk/pull/330

ebma (approved):
Thanks for adding my changes, looks good to me 🎉 

Ready for merge.

PR #335, opened 2023-04-11
Add Pagination when getting/fetching transactions in horizon

### LINKED ISSUE #308: Add pagination to check all requests of a vaults Stellar account
At the moment the vault will only fetch its 200 latest transactions when going online. This makes the vault vulnerable to spamming because an attacker could spam the vault's Stellar account with transactions to make it unable to see 'real'/important other transactions. 
We should change it such that the vault always checks all of its transactions so that this spamming does not cause any issues. For this to work we have to be able to fetch more than 200 transactions, thus add pagination. Maybe it makes sense to create a helper function that handles this pagination without the calling code having to know about the limit of 200 transactions. 
Furthermore, we could also think about storing a local indicator of the latest paging token so that the vault client can pick up the transaction fetching from that cursor instead of always having to fetch all transactions.

b-yap:
closes #308 

Uses the `paging_token` of a transaction as a "cursor". The cursor is a _url_tag_ to page through a list of transactions in horizon.

The cursor number value is different between:
* fetching transactions for executing open requests
  - this is at start up of the vault. 
  - it is more of looking backwards, so the list of transactions are actually in DESCENDING order.
  - found in stellar_wallet.rs: `fn get_all_transactions_iter(...)` 
* listening for future transactions
  - this is when the vault is already running
  - it is more of looking forward, so the list of transactions are actually in ASCENDING order.
  - found in horizon.rs: `fn fetch_transactions_iter(...)`

#### Things left to do:

- [x] the limit is **200**, but I mistakenly set it to _500_. 
- [x] retrieving all transactions, and not just 200.
   - [x] when is this called? when executing open requests?
   - [x] incorporate the `"_links"` field of `HorizonResponse`. right now it only contains the `"_embedded"` portion. 
- [x] fix integration testing
- [ ] add todos for methods/functions that has the cursor currently set to 0.
----

### How to begin the review: 

1. Look inside cache.rs:
   -  ```rust
       /// Contains the path where the transaction envelope and the cursor/paging token will be saved.
      #[derive(Debug, Clone)]
      pub struct WalletStateStorage {
	    path: String,
	    inner_path: String,
      }
      ```
      Previously _`TxEnvelopeStorage`_,  which now has additional methods to save/retrieve the last cursor.
   - the additional methods are  `get_last_cursor` and `save_cursor`.
   - renamed methods dedicated for `TxEnvelope`: **`save_tx_envelope`**, **`remove_all_tx_envelopes`**, etc.
   - added 1 test case: `fn test_cursors()`

2. Look inside stellar_wallet.rs:
     - ~~inside `fn submit_transaction(...)`: before returning a value, the `paging_token` of the transaction is saved by calling the `save_cursor(...)` method~~ :
     - make the `get_last_cursor` and `save_cursor` accessible through the wallet
     -  ~~The `cursor` and the `limit` is removed. The cursor value will be derived from accessing wallet's `get_last_cursor` method, and the limit stays 200 (and this is a constant value)~~
     - removed `get_latest_transactions`, added `fn get_all_transactions_iter(...)`
  
3. Look inside horizon.rs:
    - added the `_links` in _`HorizonTransactionsResponse`_ :
        ```rust
            #[derive(Deserialize, Debug)]
            pub struct HorizonTransactionsResponse {
	            pub _embedded: EmbeddedTransactions,
	            _links: HorizonLinks,
            }
        ```
    - created a new structure, for iterating over lists of transactions:
      ```rust
             /// An iter structure equivalent to a list of TransactionResponse
            pub struct TransactionsResponseIter<C> {
	            /// holds a maximum of 200 items
	            pub(crate) records: Vec<TransactionResponse>,
	            /// the url for the next page
	            pub(crate) next_page: String,
	            /// a client capaable to do get operation
	            pub(crate) client: C,
            }
      ```
    - added 2 fields in `HorizonFetcher`: `last_sequence` and `last_cursor` for pagination
    - removed `fetch_latest_txs` and created a new function: `fn fetch_transactions_iter(...)`
   
4. Look inside execution.rs:
    - replaced  `get_latest_transactions(false)`  to the new iter: `wallet_get_all_transactions_iter()`, to iterate over transactions instead of returning a vector


clients/wallet/src/cache.rs
```diff
@@ -21,42 +24,102 @@ macro_rules! unwrap_or_return {
 	};
 }
 
-/// Contains the path where the transaction envelope will be saved.
 #[derive(Debug, Clone)]
-pub struct TxEnvelopeStorage {
+pub struct Cache {
```
RustNinja, line 28: @b-yap 
I would appreciate if you rename it from Cache to something more specific or re-export it with other, more specific name.

b-yap: @RustNinja this cache serves as a storage/backup for the wallet's **_pagination_** and _**tx envelopes**_. And who knows in the future, it could be for something else as well. 

is it ok if I rename it to "WalletStateStorage" ? Because it saves the ff. info in a file:
1. at what cursor/page the latest transactions will be, 
2. the pending transactions of the wallet

@ebma  do you have any ideas for the name?

b-yap: see [a9b9113](https://github.com/pendulum-chain/spacewalk/pull/335/commits/a9b9113a30bf3aef95bba8c45f09e32ef4909d1d)

ebma: I think 'WalletStateStorage' is a fitting name 👍

clients/wallet/src/stellar_wallet.rs
```diff
@@ -47,9 +48,21 @@ impl StellarWallet {
 
 		let submission_result = horizon_client
 			.submit_transaction(envelope.clone(), self.is_public_network)
-			.await;
+			.await
+			.map(|resp| {
+				// use the paging token to save the latest cursor
+				if let Err(e) = self.cache.save_cursor(resp.paging_token) {
```
ebma, line 54: I am not sure if it makes sense to save the cursor here. I think this might introduce unwanted side effects. Especially considering my main comment about the scenario with the race condition.

I would only write a new paging token to the cache after fetching requests.

ebma (commented):
I think we unfortunately need some more sophisticated logic for storing the paging token. We need to save the paging token of _handled_ requests. And we cannot just store one value for this I fear. I will explain in the following.

Let's assume the vault goes online for the very first time and has plenty of open requests that it has to fulfill. Let's say the vault has 201 new Stellar transactions which are all related to issue requests that the vault needs to fulfill. 
Now, the vault will fetch its latest transactions with `get_latest_transactions()` and check them for open requests. With the pagination feature, we should now at least be able to see all 201 transactions although the page limit is just 200. Thus, the `get_latest_transactions()` function needs to be extended to either start fetching from the very first 'genesis' transaction of the account, or from a (locally-stored) paging token. And it needs to return all `201` transactions in one go. The caller of `get_latest_transactions()` should not have to call it multiple times to see all `201` transactions. At least that's what I would expect the function to do; iteratively fetch all transactions of an account until no more transactions can be fetched. The algorithm would be:
- Fetch the accounts transactions without a paging token
- loop ( If the response has exactly 200 transactions  )
  - add the fetched transactions to a vec
   - fetch the accounts transactions with the paging token of the last transaction out of the 200
- return the transactions

Now, to the problem with storing the paging token. We cannot just save the paging token of the last transaction we fetched. Because, if the vault client suddenly crashes after storing the paging token of the last transaction but the last transaction (and the preceding ones) _were not yet handled,_ (ie. they were not yet used to fulfill issue requests) the next time the vault goes online it would ignore them because it only fetches from the cached paging token. 
It seems that this caching is more complex and probably needs a separate issue ticket. I will create that tomorrow together with an idea on how to approach it, but maybe it's better to postpone the caching until then. The changes you added so far are not in vain but we should move them somewhere else and limit this PR to the pagination probably.

ebma:
@b-yap does my comment answer your question on how to fetch all transactions or should I be more specific? If something is unclear don't hesitate to ask

b-yap:
@ebma  

I actually have started last night. I cannot return a vec of transactions because the size is too big. The best would be something like an  iterator. 

```rust
pub struct TransactionsResponse<'a> {
  pub(crate) records:Vec<TransactionResponse>,
  pub(crate) order_ascending: bool,
  pub(crate) wallet: &'a StellarWallet
}

impl <'a> TransactionsResponse<'a> {
  pub async fn next(&mut self) -> Option<TransactionResponse> {
    match self.records.pop() {
	Some(x) => {
          if let Err(e) = self.wallet.save_cursor(x.paging_token) {
	          tracing::warn!("failed to save paging token {} as cursor: {e:?}", x.paging_token);
          }
          println!("------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR:{}", x.paging_token);
          Some(x)
	}
	None => {
          println!("------------- XXXXXXXXXXXXXX -------- NO MORE DATA, SO LET'S GET SOME!");
          if let Ok(result) = self.wallet.get_latest_transactions(self.order_ascending).await {
	          self.records = result;
	          return self.records.pop();
          }
          None
	}
    }
  }
}
```
This is the initial structure, returned by a different method `get_latest_transactions_iter(...)` called inside `async fn execute_open_requests`:
```rust
let wallet = wallet.read().await;
let transactions_result: Result<TransactionsResponse,Error> = wallet.get_latest_transactions_iter(false).await;
//drop(wallet);

// Check if some of the requests that are open already have a corresponding payment on Stellar
// and are just waiting to be executed on the parachain
match transactions_result {
  Ok(mut transactions) => { 
    while let Some(transaction) =  transactions.next().await {
        if rate_limiter.check().is_ok() {
	        // give the outer `select` a chance to check the shutdown signal
	        tokio::task::yield_now().await;
        }
...				
```

However the test case `test_execute_open_requests_succeeds` fails:
```
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983162494234624
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983166789201920
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983256983511040
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983274163388416
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983506091622400
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983609170829312
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983690775216128
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983768084623360
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983926998405120
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1983995717894144
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1984073027305472
------------- XXXXXXXXXXXXXX -------- THERE IS STILL DATA, SO RETURN X. NEW CURSOR: 1984081617240064
------------- XXXXXXXXXXXXXX -------- NO MORE DATA, SO LET'S GET SOME!
PRINT THE URL: https://horizon-testnet.stellar.org/accounts/GA6ZDMRVBTHIISPVD7ZRCVX6TWDXBOH2TE5FAADJXZ52YL4GCFI4HOHU/transactions?limit=200&cursor=1984081617240064&order=desc
Apr 12 19:59:07.467  INFO substrate: 💤 Idle (0 peers), best: #11 (0x7f3c…551e), finalized #11 (0x7f3c…551e), ⬇ 0 ⬆ 0    
thread 'test_execute_open_requests_succeeds' panicked at 'could not find event: Redeem::ExecuteRedeem', /spacewalk/clients/runtime/src/integration/mod.rs:165:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Apr 12 19:59:08.091  WARN parity-db: Shutdown error: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })    
test test_execute_open_requests_succeeds ... FAILED

failures:

failures:
    test_execute_open_requests_succeeds

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 11 filtered out; finished in 91.28s
``` 

Seems like a never-ending loop, because i don't think I should rely on the transaction's paging token; but instead the href links. Hence my _todo_ to update the **_`HorizonTransactionsResponse`_**.
I'll let you know if I got something that's somehow working.


-----


About saving the pagination, you're right, it's best on a separate ticket (at least the methods are already there for use). I'll rename this draft once it's actually ready for review.

clients/wallet/src/stellar_wallet.rs
b-yap: removed

clients/wallet/src/horizon.rs
```diff
@@ -131,6 +132,33 @@ where
...
+	_links: HorizonLinks,
+}
+
+#[allow(dead_code)]
+impl HorizonTransactionsResponse {
+	fn previous_page(&self) -> String {
+		self._links.prev.href.clone()
+	}
+
+	pub(crate) fn next_page(&self) -> String {
```
ebma, line 144: Somehow I would assume this function to return the new `HrefPage` struct instead of a string.

clients/wallet/src/horizon.rs
```diff
@@ -270,8 +305,53 @@ pub const fn horizon_url(is_public_network: bool) -> &'static str {
 	}
 }
 
+/// An iter structure equivalent to a list of TransactionResponse
+pub struct TransactionsResponseIter<C> {
+	/// holds a maximum of 200 items
+	pub(crate) records: Vec<TransactionResponse>,
+	/// the url for the next page
+	pub(crate) next_page: String,
+	/// a client capaable to do get operation
```
ebma, line 314: Nit: typo in `capable`

ebma: Maybe capitalize to "GET operation"

clients/wallet/src/horizon.rs
```diff
@@ -374,79 +458,71 @@ pub(crate) struct HorizonFetcher<C: HorizonClient> {
 	client: C,
 	is_public_network: bool,
 	vault_account_public_key: PublicKey,
+	/// keeps track of the latest sequence number
+	last_sequence: SequenceNumber,
```
ebma, line 462: What's the sequence number for?

clients/wallet/src/horizon.rs
```diff
@@ -374,79 +458,71 @@ pub(crate) struct HorizonFetcher<C: HorizonClient> {
...
-			transactions.iter().map(|tx| tx.paging_token).max().unwrap_or(last_paging_token);
+	) -> Result<(), Error> {
+		let mut response_iter = self.fetch_transactions_iter().await?;
 
 		let (issue_map, memos_to_issue_ids) =
 			future::join(issue_map.read(), memos_to_issue_ids.read()).await;
-		for transaction in transactions {
-			let tx = transaction.clone();
 
+		while let Some(tx) = response_iter.next().await {
```
ebma, line 525: Before calling `next()` we also have to handle the 'current' elements that the iterator was initialized with right? If we just handle the items returned by the call to `.next()` we would miss the `records` that you initialized the struct with in line 503.

ebma: Nevermind I was misunderstanding the logic in the iterator. Looks good then 👍 

But can we maybe rename `response_iter` to `transactions_iter`?

ebma (commented):
Using the page href provided in the horizon response for fetching the next page seems like a very good idea. Did this fix your issues @b-yap or did this cause other problems?

clients/wallet/src/horizon.rs
b-yap: Currently, it has no use. I figured this will come handy in the future.

The problem with cursors (esp. a transaction's paging token) is that it returns a different result depending on: the **_limit_** and the **_order_** and the time of day. 

a [next cursor of 23892903079936 with limit 1 and order asc](https://horizon-testnet.stellar.org/accounts/GCXQXD3EDC7YILNAHRJLYR53MGQU4V5RXB7JHGPVS3HQHLBJDWCJHVIB/transactions?&limit=1&order=asc) actually returns 
```
{
  "type": "https://stellar.org/horizon-errors/bad_request",
  "title": "Bad Request",
  "status": 400,
  "detail": "The request you sent was invalid in some way.",
  "extras": {
    "invalid_field": "cursor",
    "reason": "cursor: invalid value"
  }
}
```
however if you set the order to desc, the result is:
```
{
  "_links": {
    "self": {
      "href": "https://horizon-testnet.stellar.org/accounts/GCXQXD3EDC7YILNAHRJLYR53MGQU4V5RXB7JHGPVS3HQHLBJDWCJHVIB/transactions?cursor=23892903079936&limit=200&order=desc"
    },
    "next": {
      "href": "https://horizon-testnet.stellar.org/accounts/GCXQXD3EDC7YILNAHRJLYR53MGQU4V5RXB7JHGPVS3HQHLBJDWCJHVIB/transactions?cursor=23892903079936&limit=200&order=desc"
    },
    "prev": {
      "href": "https://horizon-testnet.stellar.org/accounts/GCXQXD3EDC7YILNAHRJLYR53MGQU4V5RXB7JHGPVS3HQHLBJDWCJHVIB/transactions?cursor=23892903079936&limit=200&order=asc"
    }
  },
  "_embedded": {
    "records": []
  }
}
```

In other words, saving the cursor alone is not enough. 
However, if we save the last known [sequence number](https://developers.stellar.org/docs/glossary#sequence-number) before a sudden shutdown, we can actually figure out how far away the vault is from the latest transaction. 

We could compare against the `https://horizon-testnet.stellar.org/accounts/<pub_account>`'s _**`sequence`**_ field, and know whether or not the last known transaction was actually executed or not.

Anyway, I'm not yet sure how/when/where to compare it, but i know it's sort of helpful. I hope. 
Say the _`execute_open_request`_: there's no "time" to identify when to stop iterating over a list of transactions. If we have the last known sequence number, we can stop calling `.next()` in the loop by checking whether or not the transaction's sequence number happened too long ago; and that it shouldn't be bothered with (hence breaking the loop entirely because the next transactions are useless. Why fetch 10 hours old transactions, when the vault halted only 9 hours ago?)

b-yap: actually, not sure if it should be in HorizonFetcher or the wallet. 
But i'm not sure whether it's best to share the wallet to `listen_for_new_transactions()`.

clients/vault/tests/vault_integration_tests.rs
```diff
@@ -1149,6 +1159,9 @@ async fn test_execute_open_requests_succeeds() {
 				Duration::from_secs(0),
 			)
 			.map(Result::unwrap),
+			// Sleep 5 second to give other thread some time to receive the RequestIssue event and
+			// add it to the set
+
```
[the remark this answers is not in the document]
b-yap: done

b-yap:
@ebma yep, you have a point about the cursor. I adjusted the function `listen_for_new_transactions()`, removing the `last_cursor` param. 

I only removed the extra fields from the fetcher, and left the wallet as is. We can decide on what to do in the next issue tackling the saving/adding pagination.

RustNinja (approved):

clients/vault/tests/vault_integration_tests.rs
```diff
@@ -878,6 +873,10 @@ async fn test_automatic_issue_execution_succeeds() {
 
 			tracing::warn!("Sent payment to address. Ledger is {:?}", result.unwrap().ledger);
 
+			// Sleep 3 seconds to give other thread some time to receive the RequestIssue event and
```
RustNinja, line 876: @b-yap 

> sleep 5 seconds.

clients/vault/src/oracle/mod.rs
```diff
@@ -14,23 +14,29 @@ mod errors;
 pub mod storage;
 pub mod types;
 
-#[cfg(test)]
-pub fn test_stellar_relay_config() -> stellar_relay_lib::StellarOverlayConfig {
+#[cfg(any(test, feature = "integration"))]
```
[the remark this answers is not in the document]
b-yap: created a file `testing_utils.rs`

clients/vault/src/oracle/testing_utils.rs
```diff
@@ -0,0 +1,24 @@
+pub fn test_stellar_relay_config(is_mainnet: bool) -> stellar_relay_lib::StellarOverlayConfig {
```
ebma, line 1: Nit:
```suggestion
pub fn get_test_stellar_relay_config(is_mainnet: bool) -> stellar_relay_lib::StellarOverlayConfig {
```

clients/vault/src/oracle/testing_utils.rs
```diff
@@ -0,0 +1,24 @@
...
+	let res = stellar_node_points
+		.choose(&mut rand::thread_rng())
+		.expect("should return a value");
+	let path_string = format!("./resources/config/{dir}/stellar_relay_config_{res}.json");
+
+	stellar_relay_lib::StellarOverlayConfig::try_from_path(path_string.as_str())
+		.expect("should be able to extract config")
+}
+
+pub fn test_secret_key(is_mainnet: bool) -> String {
```
ebma, line 20: Nit:
```suggestion
pub fn get_test_secret_key(is_mainnet: bool) -> String {
```

ebma (approved):
Added two nit comments but LGTM overall 👍

clients/vault/src/oracle/testing_utils.rs
b-yap: [6f02be3](https://github.com/pendulum-chain/spacewalk/pull/335/commits/6f02be36eb9e79dc4bacf29081c350cceabc67a0)

clients/vault/src/oracle/testing_utils.rs
b-yap: [6f02be3](https://github.com/pendulum-chain/spacewalk/pull/335/commits/6f02be36eb9e79dc4bacf29081c350cceabc67a0)

Issue #325, opened 2023-03-27
Fix xcm currency conversion for test chain


ebma:
Right now we have an issue with the XCM currency conversion for the test chain. 
The XCM currency conversion that was introduced in #284 is working fine with the integration tests but it is problematic when running the test chain locally. This is because for the test chain, we need to give DIA a proper conversion for the `XCM(0)` currency while in the integration tests we feed that manually. 

A quick-fix for this would be to define the XCMCurrencyConversion like this in `testchain/runtime/src/lib.rs`:
```
impl XCMCurrencyConversion for SpacewalkNativeCurrencyKey {
	fn convert_to_dia_currency_id(token_symbol: u8) -> Option<(Vec<u8>, Vec<u8>)> {
		if token_symbol == 0 {
			return Some((b"Kusama".to_vec(), b"KSM".to_vec()))
		}

		// We assume that the blockchain is always 0 and the symbol represents the token symbol
		let blockchain = vec![0u8];
		let symbol = vec![token_symbol];

		Some((blockchain, symbol))
	}

	fn convert_from_dia_currency_id(blockchain: Vec<u8>, symbol: Vec<u8>) -> Option<u8> {
		// We assume that the blockchain is always 0 and the symbol represents the token symbol
		if blockchain.len() != 1 && blockchain[0] != 0 || symbol.len() != 1 {
			return None
		}
		return Some(symbol[0])
	}
}
```
but this would break the integration tests again.

Maybe we need to change something about our `cfg_if` configuration in the testchain runtime.

PR #336, opened 2023-04-12
Use different XCMCurrencyConversion impl depends on features flag

RustNinja:
PR related to issue https://github.com/pendulum-chain/spacewalk/issues/325
- updated testchain runtime
- use different implementation of `XCMCurrencyConversion` trait depends on enabled feature flag.
- add new implementation for non `testing-utils` feature in spacewalk testnet
- keep the prev implementation of `XCMCurrencyConversion` trait for `SpacewalkNativeCurrencyKey` to keep integration working as expected.

### DIFF
--- a/testchain/runtime/src/lib.rs
+++ b/testchain/runtime/src/lib.rs
@@ -460,20 +460,47 @@ impl NativeCurrencyKey for SpacewalkNativeCurrencyKey {
 
 // It's important that this is implemented the same way as the MockOracleKeyConvertor
 // because this is used in the benchmark_utils::DataCollector when feeding prices
-impl XCMCurrencyConversion for SpacewalkNativeCurrencyKey {
-	fn convert_to_dia_currency_id(token_symbol: u8) -> Option<(Vec<u8>, Vec<u8>)> {
-		// We assume that the blockchain is always 0 and the symbol represents the token symbol
-		let blockchain = vec![0u8];
-		let symbol = vec![token_symbol];
-		Some((blockchain, symbol))
-	}
+cfg_if::cfg_if! {
+	if #[cfg(feature = "testing-utils")] {
+		impl XCMCurrencyConversion for SpacewalkNativeCurrencyKey {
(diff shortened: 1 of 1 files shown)


testchain/runtime/src/lib.rs
```diff
@@ -460,20 +460,47 @@ impl NativeCurrencyKey for SpacewalkNativeCurrencyKey {
...
+		}
 	}
-
-	fn convert_from_dia_currency_id(blockchain: Vec<u8>, symbol: Vec<u8>) -> Option<u8> {
-		// We assume that the blockchain is always 0 and the symbol represents the token symbol
-		if blockchain.len() != 1 && blockchain[0] != 0 || symbol.len() != 1 {
-			return None
+	else{
+		impl XCMCurrencyConversion for SpacewalkNativeCurrencyKey {
+			fn convert_to_dia_currency_id(token_symbol: u8) -> Option<(Vec<u8>, Vec<u8>)> {
```
b-yap, line 484: I'm not familiar with cfg_if.... but since the only difference between them is the `fn convert_to_dia_currency_id(token_symbol: u8)`, is it possible to actually do:

```rust
impl XCMCurrencyConversion for SpacewalkNativeCurrencyKey {
  
  cfg_if::cfg_if! {
    if #[cfg(feature = "testing-utils")] {
      fn convert_to_dia_currency_id(token_symbol: u8) -> Option<(Vec<u8>, Vec<u8>)> { ... }
    } 
    else {
      fn convert_to_dia_currency_id(token_symbol: u8) -> Option<(Vec<u8>, Vec<u8>)> { ... }
    }
  }

  fn convert_from_dia_currency_id(blockchain: Vec<u8>, symbol: Vec<u8>) -> Option<u8> {
		...
   }
  
}
```

RustNinja: Thanks. Done

ebma (approved):
Great idea to change it like that. Somehow I did not think of that myself haha, I expected a solution that is more complex. Good job @RustNinja 👍

PR #338, opened 2023-04-17
Use dependency from creates io for orml packages.

RustNinja:
- add `[patch.crates-io]` for:
```
orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37" }
orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37" }
orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37" }
```

- use 

`orml-... = { version = "0.4.1-dev" }` 

instead of 

`orml-... = {git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false}`

### DIFF
--- a/pallets/currency/Cargo.toml
+++ b/pallets/currency/Cargo.toml
@@ -22,9 +22,9 @@ pallet-transaction-payment = { git = "https://github.com/paritytech/substrate", 
 primitives = { package = "spacewalk-primitives", path = "../../primitives", default-features = false }
 
 # Orml dependencies
-orml-tokens = { version = "0.4.1-dev", default-features = false }
-orml-currencies = { version = "0.4.1-dev", default-features = false }
-orml-traits = { version = "0.4.1-dev", default-features = false }
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
+orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
 
 # for other pallets wanting to mock functions
 mocktopus = { version = "0.8.0", optional = true }
--- a/pallets/currency/Cargo.toml
+++ b/pallets/currency/Cargo.toml
@@ -22,9 +22,9 @@ pallet-transaction-payment = { git = "https://github.com/paritytech/substrate", 
 primitives = { package = "spacewalk-primitives", path = "../../primitives", default-features = false }
 
 # Orml dependencies
-orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
-orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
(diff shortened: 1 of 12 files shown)


pallets/currency/Cargo.toml
```diff
@@ -22,9 +22,9 @@ pallet-transaction-payment = { git = "https://github.com/paritytech/substrate",
 primitives = { package = "spacewalk-primitives", path = "../../primitives", default-features = false }
 
 # Orml dependencies
-orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
-orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
-orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
+orml-tokens = { version = "0.4.1-dev", default-features = false }
+orml-currencies = { version = "0.4.1-dev", default-features = false }
+orml-traits = { version = "0.4.1-dev", default-features = false }
```
ebma, line 27: Is it really necessary to change all these dependencies to the higher version just to patch them down again? Why don't we just leave the declarations we have already? Is there an error if we leave the version declaration as it was?

RustNinja: @ebma 
yes. It is an error if we leave them as before.
There error looks like that "you do not use patch 0.4.1-dev in your packages" and it failed to compile.

ebma: Aha okay hmm. Can you try the trick with the double slash? You might have noticed that in the existing `patch` statements many URL declarations have an 'unnecessary' double slash. Maybe if you use that, ie. 
```
- orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
- orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }
- orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.37", default-features = false }

+ orml-traits = { git = "https://github.com/open-web3-stack//open-runtime-module-library", branch = "polkadot-v0.9.37" }
+ orml-currencies = { git = "https://github.com/open-web3-stack//open-runtime-module-library", branch = "polkadot-v0.9.37" }
+ orml-tokens = { git = "https://github.com/open-web3-stack//open-runtime-module-library", branch = "polkadot-v0.9.37" }
```
the compilation error disappears and we can get rid of all the extra changes?

RustNinja: Yes. Thanks. It works. Even more. Now it works with one slash to.))

RustNinja: @ebma 
Done.

RustNinja: @ebma but anyway
you wrote that we also need use dependency from crates for orml packages. so i added dependency to crates instead of git.

RustNinja:
> @RustNinja do we actually need to patch the orml dependencies in spacewalk though? Or would this only be required for pendulum/amplitude/foucoco? We are not using any bifrost dependency in spacewalk itself do we?

Oh. yes. you 100% right. 
We do not need any `[patch.crates-io]` in spacewalk.
We need it in the pendulum.

Will close this PR at all. 
If we need to use crates dependency instead of git for orml packages please, let me know.

ebma:
Okay perfect, got it.

PR #339, opened 2023-04-19
Indefinitely resubmit transactions with recoverable error

### LINKED ISSUE #315: Retry Stellar transactions indefinitely

b-yap:
closes #315 

Transactions with timeout issues will be retried again and again, without any limit.

1. Remove the limited "maxretries" logic; 
2. Add method for wallet `Error` to determine if an error is a timeout or server error.
3. Add a fallback horizon url in case of server errors
4. Implement a loop inside `submit_transaction()`
5. Update how the `fn resubmit_transactions_from_cache()` handles errors.
6.  The handling of errors during resubmission will NOT be covered in this PR. (it's a `todo`)

------

Things left to do:

- [x] what is the fallback horizon url?

----

### How to begin the review:
1. Go to task.rs: 
   - the _`ReachedMaxRetries`_ variant of `SlotTaskStatus` is removed.
   - the _`retries_remaining`_ field of `SlotTask` is removed.
   - no more _`max_retries_exhausted()`_ test case.
2. Go to error.rs:
   - a method `fn is_recoverable(&self) -> bool` is added for the error
   - a method `is_server_error(&self) -> bool` is added for the error
3. Go to horizon.rs:
   - check the loop inside `fn submit_transaction(...`:
        - ```rust
           let need_fallback = if server_error_count == 0 {
				      // reset the count, in case the fallback fails but the original url is back running.
				      server_error_count = 3;
				      true
			      } else {
				      false
			      };
           ```  
           3 retries for internal server errors before using a fallback horizon url. However, the counter resets back to 3 once the fallback url is used.
       -  ```rust
          let response =
				match self.post(url).form(&params).send().await.map_err(Error::HttpFetchingError) {
					Ok(response) => interpret_response::<TransactionResponse>(response).await,
					Err(e) => {
						if e.is_recoverable() || e.is_server_error() {
							if e.is_server_error() {
								server_error_count -= 1;
							}
						...
						continue	

           ```    
           continue with the loop if it's either: a recoverable error or a server_error
      - check the error again after interpreting the response.
4. Go to stellar_wallet.rs:
   -  new return type:
      ```rust 
       pub async fn resubmit_transactions_from_cache(
          &self,
       ) -> Vec<oneshot::Receiver<Result<TransactionResponse, Error>>> {
      ``` 
      with 1 channel PER transaction. The channel returns a result of the transaction, whether it succeeded or failed.
      An example of how the return type is handled can be seen in the **_`fn resubmit_transactions_works()`_** test case.
5. Go to system.rs:
    - as mentioned in the overview, handling of errors is a "todo" and will be addressed in another issue or PR.

### DIFF
--- a/clients/vault/src/oracle/mod.rs
+++ b/clients/vault/src/oracle/mod.rs
@@ -17,4 +17,5 @@ pub mod types;
 #[cfg(any(test, feature = "integration"))]
 mod testing_utils;
 
+#[cfg(any(test, feature = "integration"))]
 pub use testing_utils::*;


clients/vault/src/system.rs
```diff
@@ -452,11 +452,8 @@ impl VaultService {
 		let is_public_network = wallet.is_public_network();
 
 		// re-submit transactions in the cache
-		let (_, errors) = wallet.resubmit_transactions_from_cache().await;
-		if !errors.is_empty() {
-			// todo: handle timeouts
-			tracing::error!("Failed to resubmit: {:?}", errors);
-		}
+		let _receivers = wallet.resubmit_transactions_from_cache().await;
+		//todo: handle errors from the receivers
```
RustNinja, line 456: I think that we already able to remove this TODO comment because we have a pretty good logging mechanism inside `resubmit_transactions_from_cache` function

clients/wallet/src/horizon.rs
```diff
@@ -427,30 +435,77 @@ impl HorizonClient for reqwest::Client {
...
-					status,
-					reason,
-					envelope_xdr: Some(transaction_xdr.to_string()),
-				}
-			},
-			other => other,
-		})
+
+		let mut server_error_count = 3;
+		loop {
```
RustNinja, line 445: My opinion that It is better to move value '3' to configuration instead of hardcoding.

clients/vault/src/system.rs
b-yap: it's for https://github.com/pendulum-chain/spacewalk/issues/340

clients/wallet/src/horizon.rs
b-yap: @RustNinja as I do not know what our specific num of attempts should be, i added a method of StellarWallet:
**_`pub fn with_max_retry_attempt(mut self, max_retries: u8) -> Self {...`_**

and we can configure it in system.rs.

clients/wallet/src/error.rs
```diff
@@ -31,6 +31,36 @@ pub enum Error {
...
+	}
+
+	pub fn is_server_error(&self) -> bool {
+		match self {
+			Error::HttpFetchingError(e)
+				if e.status() == Some(reqwest::StatusCode::INTERNAL_SERVER_ERROR) =>
+				true,
+			Error::HorizonSubmissionError { title: _, status, reason: _, envelope_xdr: _ }
+				if *status == 500 =>
+				true,
```
ebma, line 59: Should we add a range to this so that all 5xx errors are declared as server errors? Probably not that relevant but per definition of HTTP all 5xx errors would be server errors IIRC.

clients/wallet/src/horizon.rs
```diff
@@ -426,31 +439,69 @@ impl HorizonClient for reqwest::Client {
...
-					status,
-					reason,
-					envelope_xdr: Some(transaction_xdr.to_string()),
-				}
-			},
-			other => other,
-		})
+
+		let mut server_error_count = 0;
+		let mut counter = 1;
```
ebma, line 451: Let's indicate with a name what this counter counts.

clients/wallet/src/horizon.rs
```diff
@@ -426,31 +439,69 @@ impl HorizonClient for reqwest::Client {
...
+					if e.is_server_error() {
+						server_error_count += 1;
+					}
+
+					// let's wait awhile before resubmitting.
+					tracing::warn!(
+						"submission failed for transaction with sequence number {seq_no:?}: {e:?}"
+					);
+					// exponentially sleep before retrying again
+					sleep(Duration::from_secs(2u64.pow(counter))).await;
```
ebma, line 483: Since with the exponential growth, this sleep duration might get huge with enough failures it might make sense to let it saturate at a max value. 
Maybe we can use a min function here. eg. `std::cmp::min(2u64pow(counter), 600)`. We could create another constant for that other value and call it `MAX_BACKOFF_DELAY` or similar.

clients/wallet/src/horizon.rs
```diff
@@ -297,9 +301,17 @@ pub struct Claimant {
 	// pub predicate: serde_json::Value,
 }
 
-pub const fn horizon_url(is_public_network: bool) -> &'static str {
+pub const fn horizon_url(is_public_network: bool, is_need_fallback: bool) -> &'static str {
 	if is_public_network {
+		if is_need_fallback {
+			//todo: what's the fallback address?
+			return "fallback domain name"
```
ebma, line 308: `https://horizon.stellarx.com`
or 
`https://horizon.stellar.lobstr.co`

clients/wallet/src/horizon.rs
```diff
@@ -297,9 +301,17 @@ pub struct Claimant {
...
+		}
+
 		"https://horizon.stellar.org"
+	} else if is_need_fallback {
+		//todo: what's the fallback address?
+		"fallback domain for testnet"
```
ebma, line 314: There are none for testnet that I know of

clients/wallet/src/stellar_wallet.rs
```diff
@@ -35,23 +35,30 @@ pub struct StellarWallet {
 	/// Used for caching Stellar transactions before they get submitted.
 	/// Also used for caching the latest cursor to page through Stellar transactions in horizon
 	cache: WalletStateStorage,
+
+	/// maximum retry attempts for submitting a transaction before switching to a fallback url
+	max_retry_attempt: u8,
```
ebma, line 40:
```suggestion
	max_retry_attempts_before_fallback: u8,
```

clients/wallet/src/stellar_wallet.rs
```diff
@@ -35,23 +35,30 @@ pub struct StellarWallet {
...
 }
 
 impl StellarWallet {
+	/// if the user doesn't define the maximum number of retry attempts for 500 internal server
+	/// error, this will be the default.
+	const MAX_RETRY_ATTEMPT: u8 = 3;
```
ebma, line 46:
```suggestion
	const MAX_RETRY_ATTEMPTS_BEFORE_FALLBACK: u8 = 3;
```

clients/wallet/src/stellar_wallet.rs
```diff
@@ -141,9 +148,16 @@ impl StellarWallet {
 			is_public_network,
 			transaction_submission_lock: Arc::new(Mutex::new(())),
 			cache,
+			max_retry_attempt: Self::MAX_RETRY_ATTEMPT,
 		})
 	}
 
+	pub fn with_max_retry_attempt(mut self, max_retries: u8) -> Self {
```
ebma, line 155: Nit: I'd also rename this to match the other name I suggested

ebma (commented):
I gave some feedback already but have to check the details again

RustNinja (approved):

clients/wallet/src/error.rs
b-yap: done.

clients/wallet/src/horizon.rs
b-yap: renamed to `exponent_counter`.

clients/wallet/src/horizon.rs
b-yap: instead of 600, i've set the `MAX_BACKOFF_DELAY` to the power of **_9_**; 2^9 is the nearest to 600.

clients/wallet/src/horizon.rs
b-yap: done

clients/wallet/src/cache.rs
```diff
@@ -158,7 +158,7 @@ impl WalletStateStorage {
 	/// Removes the directory itself.
 	/// User should not be able to do this in production.
 	pub fn remove_dir(&self) -> bool {
-		std::fs::remove_dir_all(&self.path).is_ok()
+		std::fs::remove_dir_all(&self.path).is_ok() || !Path::new(&self.path).is_dir()
```
ebma, line 161: Is this just used to rturn a proper boolean in case the directory did not exist or am I misreading this?

clients/wallet/src/error.rs
```diff
@@ -31,6 +31,47 @@ pub enum Error {
...
+		let check_status =
+			|status: u16| status >= internal_server_error && status <= net_auth_required;
+
+		match self {
+			Error::HttpFetchingError(e) => {
+				if let Some(status) = e.status() {
+					return check_status(status.as_u16())
+				}
+				false
+			},
```
ebma, line 67: Maybe we can simplify this as follows? I guess you wanted to limit it to the status codes that actually exist but I guess we don't even need to do this limitation. As the server error responses range from [500-599](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses), we could just increase the range I suppose. WDYT @b-yap?
```suggestion
		match self {
			Error::HttpFetchingError(e) => {
				if let Some(status) = e.status() {
					let server_errors = 500..599;
					return server_errors.contains(&status)
				}
				false
			},
```

clients/wallet/src/horizon.rs
```diff
@@ -426,31 +442,74 @@ impl HorizonClient for reqwest::Client {
...
+					tracing::warn!(
+						"submission failed for transaction with sequence number {seq_no:?}: {e:?}"
+					);
+					// exponentially sleep before retrying again
+					sleep(Duration::from_secs(2u64.pow(exponent_counter))).await;
+					tracing::debug!("resubmitting transaction with sequence number {seq_no:?}...");
+
+					// retry/resubmit again
+					if exponent_counter > u32::from(max_backoff_delay) {
+						exponent_counter = 1;
```
ebma, line 492: Instead of re-setting the increment and starting from `1` again, as a caller of this function I would naturally assume that once it reaches the maximum, it just stays at the maximum (and does not flip back to a lower delay again). So I would suggest instead of the `if-else` we just use an `if` statement where we check if the exponent counter is less than the limit and increment it if that's the case. And we do nothing in the other case.

ebma: Also, the naming is a bit confusing. We have the `max_backoff_delay` here, which without knowing the implementation would make you think that this is some value in seconds, which is the maxium backoff delay, meaning that the transaction submission will not be delayed more than this value. _But_ what this actually is, is an exponent for the exponential backoff calculation. So we'd at least have to rename it so we don't cause confusion about it.

However, I think I'd prefer if we rename the variable to `max_backoff_delay_seconds`. Then in this check here, we'd do (also including the changes from my previous comment):
```
if 2u64pow(exponent_counter) < max_backoff_delay_seconds {
	exponent_counter += 1;
}
```


And for the sleep statement above, we'd do:
```
let sleep_duration = std::cmp::min(2u64.pow(exponent_counter, max_backoff_delay_seconds)));
sleep(sleep_duration);
```


WDYT @b-yap?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -35,23 +35,39 @@ pub struct StellarWallet {
...
+
+	max_backoff_delay: u8,
 }
 
 impl StellarWallet {
+	/// if the user doesn't define the maximum number of retry attempts for 500 internal server
+	/// error, this will be the default.
+	const MAX_RETRY_ATTEMPTS_BEFORE_FALLBACK: u8 = 3;
+
+	const MAX_BACKOFF_DELAY: u8 = 9;
```
ebma, line 50: Nit: Should we prepend a 'DEFAULT_' to the names to make it clearer? Not sure

clients/wallet/src/stellar_wallet.rs
```diff
@@ -141,9 +157,26 @@ impl StellarWallet {
 			is_public_network,
 			transaction_submission_lock: Arc::new(Mutex::new(())),
 			cache,
+			max_retry_attempts_before_fallback: Self::MAX_RETRY_ATTEMPTS_BEFORE_FALLBACK,
+			max_backoff_delay: Self::MAX_BACKOFF_DELAY,
 		})
 	}
 
+	pub fn with_max_retry_attempt_before_fallback(mut self, max_retries: u8) -> Self {
```
ebma, line 165:
```suggestion
	pub fn with_max_retry_attempts_before_fallback(mut self, max_retries: u8) -> Self {
```

clients/wallet/src/stellar_wallet.rs
```diff
@@ -141,9 +157,26 @@ impl StellarWallet {
...
+		self.max_retry_attempts_before_fallback = max_retries;
+
+		self
+	}
+
+	pub fn with_max_backoff_delay(mut self, max_backoff_delay: u8) -> Self {
+		// a number more than the default max would be too large
+		if max_backoff_delay < Self::MAX_BACKOFF_DELAY {
+			self.max_backoff_delay = max_backoff_delay;
+		}
```
ebma, line 175: I think this check can be removed if we change the max_backoff_delay the way I described. Just saying, so we don't forget

clients/wallet/src/stellar_wallet.rs
```diff
@@ -191,25 +224,52 @@ impl StellarWallet {
...
 	/// Returns a tuple: (list of txs successfully submitted, list of errors of txs not resubmitted)
-	pub async fn resubmit_transactions_from_cache(&self) -> (Vec<TransactionResponse>, Vec<Error>) {
+	pub async fn resubmit_transactions_from_cache(
+		&self,
+	) -> Vec<oneshot::Receiver<Result<TransactionResponse, Error>>> {
 		let _ = self.transaction_submission_lock.lock().await;
 		let horizon_client = Client::new();
 
-		let mut passed = vec![];
+		let mut return_val = vec![];
```
ebma, line 233: Let's rename this. The doc comment above also needs to be updated I think.

Maybe to (?)
```suggestion
		let mut error_receivers = vec![];
```

clients/wallet/src/stellar_wallet.rs
```diff
@@ -191,25 +224,52 @@ impl StellarWallet {
...
+
```
ebma, line 234: Let's add a comment here. Maybe

```suggestion
// Iterates over all errors and creates channels that are used to send errors back to the caller of this function. 
```

clients/wallet/src/horizon.rs
```diff
@@ -1,9 +1,14 @@
...
-use futures::future;
+use futures::{
+	future::{self, ready},
+	TryFutureExt,
+};
+
 use parity_scale_codec::{Decode, Encode};
-use primitives::TextMemo;
+use primitives::{TextMemo, TransactionEnvelopeExt};
+use rand::seq::SliceRandom;
```
ebma, line 11: It seems like this is not used, so we can remove this and the crate from Cargo.toml again I suppose

ebma (commented):
Added more comments. Core logic looks good to me, my comments are just some suggestions for improvements that came to my mind

clients/wallet/src/cache.rs
b-yap: updated the method to return nothing, and perform the `remove_dir_all()` only if the path exists.

clients/wallet/src/error.rs
b-yap: done

clients/wallet/src/horizon.rs
b-yap: done

clients/wallet/src/stellar_wallet.rs
b-yap: hhm, user could set it to `u16::MAX`.  Maybe it's best to let the user allow it,  only if it's not too big.

And the const `DEFAULT_MAX_BACKOFF_DELAY_IN_SECS` is found in the `stellar_wallet`, not in horizon.

clients/wallet/src/stellar_wallet.rs
b-yap: done

clients/wallet/src/horizon.rs
b-yap: the rand? i used it to determine the horizon url fallback

ebma: Ah oh, I thought I checked the code and didn't see any references. But maybe I just missed the function that used that haha

ebma (approved):
Thanks for adding the requested changes. LGTM now. Good job @b-yap 👍

PR #342, opened 2023-04-25
replace VecDequeue with a struct _containing_ VecDequeue

### LINKED ISSUE #309: Fix increasing resource consumption of vault client
Apparently running the vault client for some period of time on our infrastructure makes it break because it starts consuming too many resources. This happened on t3.micro machines with 1GB of RAM. We should check which values are not properly pruned and cause the memory consumption to increase linearly. 

One thing that probably causes this is an incomplete implementation of the maps for storing transaction sets and scp envelopes. The `EnvelopesMap` and `TxSetMap` are defined as [VecDeques](https://github.com/pendulum-chain/spacewalk/blob/0d9e1f4fa7bf300d1c5007bdc0439f491e9e979b/clients/vault/src/oracle/types.rs#L25-L26) and a custom `LifoMap` trait was implemented for easier getters/setters. The idea was to limit these VecDeques in their size so that the last/oldest item is dropped when a new item is added that exceeds the defined limit. This is not implemented at the moment though and might be the cause of the ever-increasing resource consumption. There is a constant defined [here](https://github.com/pendulum-chain/spacewalk/blob/0d9e1f4fa7bf300d1c5007bdc0439f491e9e979b/clients/vault/src/oracle/constants.rs#L10) but it's not used where it needs to be.

We should: 
- [ ] Rename `LifoMap` to `LimitedFifoMap` because we want to do 'first in first out' actually
- [ ] Turn `LimitedFifoMap` from a trait into a struct. The problem is that we want to define the maximum number of items when instantiating it and we need to access that value for our helper methods so we cannot use a trait. 
- [ ] Instantiate these maps with reasonable limits. Should be at least 200 because when falling back to an archive these contain 200 SCP messages and choosing a smaller limit would drop some potentially important SCP messages from archive files.

The struct can look roughly as follows
```
pub struct LimitedFifoMap<T,K> {
  limit: usize
  queue: VecDeque<T,K>
}

impl<T, K> LimitedFifoMap<(K, T)>
where
	K: PartialEq,
{
     fn new() {
         let default_limit = ...
         Self { }
     }

     fn new_with_limit(){}

      // ...  + the existing helpers on `LifoMap`
     // but accommodating for the limit parameter we have
      fn set_with_key(&mut self, key: K, value: T) {
		// If the key already exists, remove it first
		self.queue.remove_with_key(&key);

                if self.queue.length == limit {
                   // remove first element in queue
                   self.queue.pop_front();
                }

		self.queue.push_back((key, value));
	}
}
```

### Checking the resource consumption
To be able to judge whether the resource consumption was 'fixed' we should check how much RAM the vault client consumes when being run for some time. This should be checked before adding these changes and after. We would expect that the RAM usage of the vault client now saturates at a maximum (when the SCP maps are filled). Thus, when checking the 'after'-implementation, we first have to wait until the maps are filled and can then judge if the RAM usage continues growing.
I would suggest doing the following for the before/after comparison:
1) Run the vault client for two hours
2) Check the resource consumption after one and two hours. (For this, simply check the RAM usage of the process in your system monitor)

Repeat steps 1) and 2) with the new implementation. We would expect that the RAM usage after one hour is very close to the RAM usage after two hours. If that's the case we can consider this ticket completed. Consider adding your measured values to this ticket so they can be used for reference.

b-yap:
potentially closes #309 

Introduce a size limit for maps that stores the transactionevenlope and transactionset.

1.  Create the `LimitedFifoMap` struct.
2. Changed the datatype for `EnvelopesMap` and `TxSetMap` to a `LimitedFifoMap`
3. Create a test case to for all the methods of `LimitedFifoMap`
4. There were problems connecting to sdftest1, so I've opted to use sdftest2 instead.
------

Things left to do:
- [ ] compare the memory size of the running vault  with the code changes vs the code in main branch 

----

### How to begin the review:
1. Go to vault/src/oracle/types.rs:
   - ```rust
      #[derive(Debug, Clone)]
      pub struct LimitedFifoMap<K, T> {
	      limit: usize,
	      queue: VecDeque<(K, T)>,
      }
      ```
      - check all methods of struct `LimitedFifoMap`, _mimicking_ the method names of the map, except for `first()` and `append()`
      - check the test case `test_LimitedFifoMap` 
2. Go to vault/src/oracle/storage/impl.rs
    - instead of calling 2 methods`push_back()` and  `pop_front()`, the logic is now found in 1 method: `insert()`
3. For other files: 
    - `get_with_key()` is already `get()`   
    - `set_with_key()` is `insert()`
    -  `remove_with_key()` is `remove()`
    -  `contains_key()` is `contains()`

### DIFF
--- a/clients/stellar-relay-lib/resources/config/testnet/stellar_relay_config_sdftest2.json
+++ b/clients/stellar-relay-lib/resources/config/testnet/stellar_relay_config_sdftest2.json
@@ -1,6 +1,6 @@
 {
   "connection_info": {
-    "address": "34.235.168.98",
+    "address": "3.82.209.201",
     "port": 11625
   },
   "node_info": {
--- a/clients/vault/src/oracle/types.rs
+++ b/clients/vault/src/oracle/types.rs
@@ -46,7 +46,9 @@ where
 	}
 
 	pub fn with_limit(mut self, limit: usize) -> Self {
-		// cannot set a number smaller than the default.
+		// cannot set a number smaller than the default, because if the limit is too small,
+		// restoring SPC messages from an archive might have issues
+		// since the archived data is inserted in large batches.
 		if limit < DEFAULT_MAX_ITEMS_IN_QUEUE {
 			self.limit = DEFAULT_MAX_ITEMS_IN_QUEUE
 		} else {
@@ -91,7 +93,7 @@ where
 		// remove the oldest entry if the queue reached its limit
 		if self.queue.len() == self.limit {
 			if let Some(oldest_entry) = self.queue.pop_front() {
-				tracing::debug!("removing old entry with key: {:?}", oldest_entry.0);
+				tracing::info!("removing old entry with key: {:?}", oldest_entry.0);
 			}
 		}
 
@@ -106,17 +108,19 @@ where
 		let allowable_size = self.limit - self.len();
 
 		let mut other_queue = other.queue;
+		let other_queue_len = other_queue.len();
 
-		let last_partition = if allowable_size < other_queue.len() {
+		let mut last_partition = if other_queue_len > allowable_size {
+			let split_index = other_queue_len - allowable_size;
 			// split off the 'other' map, since it's too big to append all of its elements
-			other_queue.split_off(allowable_size)
+			other_queue.split_off(split_index)
 		} else {
 			VecDeque::new()
 		};
 
-		self.queue.append(&mut other_queue);
+		self.queue.append(&mut last_partition);
 
-		last_partition
+		other_queue
 	}
 }
 
@@ -198,6 +202,7 @@ mod test {
 		// --------- test insert and len methods ---------
 		let fill_size = sample_map.limit();
 
+		// inserts a char value of the index.
 		for x in 0..fill_size {
 			let key = u32::try_from(x).expect("should return ok");
 			let value = char::from_u32(key).unwrap_or('x');
@@ -257,12 +262,22 @@ mod test {
 		let second_len = second_map.len();
 		let remaining_space = limit - second_len;
 
+		// the remainder should be the old elements of `sample_map`.
 		let remainder = second_map.append(sample_map.clone());
+		let remainder_len = remainder.len();
 		assert_eq!(second_map.len(), limit);
+		assert_eq!(remainder_len, sample_map.len() - remaining_space);
+
+		// check that the elements of the remainder list, is the same with the elements
+		// of the original sample map.
 
 		let first_remainder = remainder[0];
-		let expected = sample_map.queue[remaining_space];
+		let expected = sample_map.queue[0];
 		assert_eq!(first_remainder, expected);
+
+		let last_remainder = remainder[remainder_len - 1];
+		let expected = sample_map.queue[remainder_len - 1];
+		assert_eq!(last_remainder, expected);
 	}
 
 	#[test]
--- a/clients/vault/src/oracle/types.rs
+++ b/clients/vault/src/oracle/types.rs
@@ -110,17 +110,18 @@ where
 		let mut other_queue = other.queue;
 		let other_queue_len = other_queue.len();
 
-		let mut last_partition = if other_queue_len > allowable_size {
+		if other_queue_len > allowable_size {
 			let split_index = other_queue_len - allowable_size;
 			// split off the 'other' map, since it's too big to append all of its elements
-			other_queue.split_off(split_index)
+			let mut last_partition = other_queue.split_off(split_index);
+
+			self.queue.append(&mut last_partition);
+
+			other_queue
 		} else {
+			self.queue.append(&mut other_queue);
 			VecDeque::new()
-		};
-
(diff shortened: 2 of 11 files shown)


b-yap:
testing with the standalone chain in this branch, the memory jump (between 2-3 hours) was from 44mb to 52.1mb.

clients/stellar-relay-lib/resources/config/testnet/stellar_relay_config_sdftest2.json
```diff
@@ -1,6 +1,6 @@
 {
   "connection_info": {
-    "address": "34.235.168.98",
+    "address": "3.82.209.201",
```
ebma, line 3: Even if there are issues with sdftest1 at the moment, maybe it would make more sense not to _replace_ sdftest1 with sdftest2 but to create a new file next to it?

clients/vault/src/oracle/types.rs
```diff
@@ -22,47 +26,103 @@ pub(crate) type SlotEncodedMap = BTreeMap<Slot, SerializedData>;
...
-	fn get_with_key(&self, key: &K) -> Option<&T> {
-		self.iter().find(|(k, _)| k == key).map(|(_, v)| v)
+	pub fn new() -> Self {
+		LimitedFifoMap { limit: DEFAULT_MAX_ITEMS_IN_QUEUE, queue: VecDeque::new() }
+	}
+
+	pub fn with_limit(mut self, limit: usize) -> Self {
+		// cannot set a number smaller than the default.
+		if limit < DEFAULT_MAX_ITEMS_IN_QUEUE {
+			self.limit = DEFAULT_MAX_ITEMS_IN_QUEUE
```
ebma, line 51: Nice catch. Maybe we should also mention why. We can add a comment that says
```
// because if the limit is too small, restoring SPC messages from an archive might have issues since the archived data is inserted in large batches.
```

clients/vault/src/oracle/types.rs
```diff
@@ -22,47 +26,103 @@ pub(crate) type SlotEncodedMap = BTreeMap<Slot, SerializedData>;
...
+	/// Consumes the other and returns the excess of it. The limit will be based on self.
+	pub fn append(&mut self, other: Self) -> VecDeque<(K, T)> {
+		// check the remaining size available for this map.
+		let allowable_size = self.limit - self.len();
+
+		let mut other_queue = other.queue;
+
+		let last_partition = if allowable_size < other_queue.len() {
+			// split off the 'other' map, since it's too big to append all of its elements
+			other_queue.split_off(allowable_size)
```
ebma, line 112: Interesting approach. Now something that comes to my mind is, whether we want to have the 'younger' or 'older' split of elements when we cut this off. We probably want to have the 'younger' elements, and if I understand correctly your implementation already does exactly that :+1:

clients/vault/src/oracle/types.rs
```diff
@@ -114,7 +174,96 @@ impl TxSetHashAndSlotMap {
...
+		// change limit is less than minimum
+		let expected_limit = 199;
+		let mut sample_map = new_sample.with_limit(expected_limit);
+		assert_ne!(sample_map.limit(), expected_limit);
+		assert_eq!(sample_map.limit(), DEFAULT_MAX_ITEMS_IN_QUEUE);
+
+		// --------- test insert and len methods ---------
+		let fill_size = sample_map.limit();
+
+		for x in 0..fill_size {
```
ebma, line 201: Maybe add a comment above that you insert values equal to their index

ebma (approved):
I love it, great job @b-yap :+1:

The memory jump from 44mb to 52.1mb is better than before obviously. But (out of curiosity) did you also test if it increases even more after that?

b-yap:
@ebma 
Screen Shot 2023-05-03 at 5 22 05 PM
The terminal started at around 39~40MB. 
Been running for more than 4 hours now, and the highest it jumped (so far) is 70MB. But then it goes back to 58/59 MB.

The screenshot on the right side you will see 2 logs with entry deletion of the same slot.
1 is from the `EnvelopesMap`, the other is `TxSetMap`. I've reverted the log back to `debug` (from info)

b-yap:
canceled the CI, because there's a [running job](https://github.com/pendulum-chain/spacewalk/actions/runs/4870266462/jobs/8685804858?pr=343).

will execute once the prev. job is done.

clients/stellar-relay-lib/resources/config/testnet/stellar_relay_config_sdftest2.json
b-yap: done. this sdftest2 is added

b-yap: done

clients/vault/src/oracle/types.rs
b-yap: @ebma  can you review again? 

you can check the test case at lines 274 to 281

ebma (approved):
LGTM, ready for merge I'd say

PR #343, opened 2023-05-02
use the "saturated" amount as the reimbursed amount

### LINKED ISSUE #281: Fix untracked amount transferred
There is a scenario in which the vault is able to gain more funds with the `mint_tokens_for_reimbursed_redeem` function than it should be able to. The reason is that when cancelling a redeem, we use the `vault_registry::transfer_funds_saturated()` function to reimburse the user. Since it is 'saturated' this function will try to transfer the target amount, or less if not enough funds are available. The problem is that we don't store information about the amount that is actually transferred. In case that this amount is less than originally intended, the vault operator might be able to gain more funds than he lost when he reimburses the redeem.

To solve it we need to store information about the actually transferred amount, ie. the return value of the `vault_registry::transfer_funds_saturated()` function call and use it in the `mint_tokens_for_reimbursed_redeem` function.

The related audit ticket is `LI7-04`

b-yap:
closes #281 

use the amount derived from `transfer_funds_saturated()` as the amount to reimburse.

1.  Create a storage map called `CancelledRedeemAmount` which will be the amount to reimburse
2. An item is inserted to `CancelledRedeemAmount` when calling `cancel_redeem()`.
3. An item is removed in `CancelledRedeemAmount` when calling `mint_tokens_for_reimbursed_redeem()`.

Things left to do:
- [ ] add a test case


pallets/redeem/src/lib.rs
```diff
@@ -917,7 +931,12 @@ impl<T: Config> Pallet<T> {
 
 		ensure!(redeem.vault == vault_id, Error::<T>::UnauthorizedVault);
 
-		let reimbursed_amount = redeem.amount().checked_add(&redeem.transfer_fee())?;
+		//let reimbursed_amount = redeem.amount().checked_add(&redeem.transfer_fee())?;
```
ebma, line 934: Let's remove the comment

pallets/redeem/src/lib.rs
```diff
@@ -822,6 +826,11 @@ impl<T: Config> Pallet<T> {
 				&confiscated_collateral,
 			)?;
 
+			CancelledRedeemAmount::<T>::insert(
+				redeem_id,
+				(confiscated_collateral.amount(), confiscated_collateral.currency()),
+			);
+
```
ebma, line 833: Somehow I fear that this is not accurate and we should not do this. 
The audit ticket was about issues with `transfer_funds_saturated()` and not about the `transfer_funds()` case. I am not 100% sure though, but I'd prefer if we remove this and add an additional check below.

pallets/redeem/src/lib.rs
```diff
@@ -917,7 +931,12 @@ impl<T: Config> Pallet<T> {
 
 		ensure!(redeem.vault == vault_id, Error::<T>::UnauthorizedVault);
 
-		let reimbursed_amount = redeem.amount().checked_add(&redeem.transfer_fee())?;
+		//let reimbursed_amount = redeem.amount().checked_add(&redeem.transfer_fee())?;
+
+		let Some((amount,currency_id)) = CancelledRedeemAmount::<T>::take(redeem_id) else {
+			return Err(DispatchError::from(Error::<T>::RedeemIdNotFound));
+		};
+		let reimbursed_amount = Amount::new(amount, currency_id);
```
ebma, line 939: To follow up on the other ticket, here I would add a check to see if an item for the `redeem_id` exists in the map or not. And based on that we determine the reimbursed amount.

```suggestion
                 // Here, we check if the `transfer_funds_saturated()` function was used during cancellation and if that's not the case, calculate the reimbursed_amount
		let reimbursed_amount = match CancelledRedeemAmount::<T>::take(redeem_id) {
		Some(saturated_amount) => saturated_amount,
		None => {
                      redeem.amount().checked_add(&redeem.transfer_fee())?
		};
```

WDYT @b-yap?

ebma (changes requested):
I suggested changes because I'm scared that this brings implications that we did not consider. @TorstenStueber do you have an opinion on this? These changes could be critical security wise.

pallets/redeem/src/lib.rs
ebma: @TorstenStueber my main concern is that we remove this line. 
@b-yap added changes such that the transferred amount is stored in both cases, 1) when the vault is liquidated and 2) when it's not. The audit ticket only raised concerns about the `transfer_funds_saturated()` function which is used in case 2). 

The existing code calculated the reimbursed_amount with 
```
let reimbursed_amount = redeem.amount().checked_add(&redeem.transfer_fee())?;
```

but in case 1), the amount transferred, (which @b-yap decided to store in the `CancelledRedeemAmount` storage item) is however calculated as
```
let confiscated_collateral = ext::vault_registry::calculate_collateral::<T>(
				&ext::vault_registry::get_liquidated_collateral::<T>(&redeem.vault)?,
				&vault_to_be_burned_tokens,
				&vault_to_be_redeemed_tokens, /* note: this is the value read prior to making
				                               * changes */
			)?;
```

Maybe that does not matter, maybe it does. Anyways, with the changes I suggested in my other comments, we would keep using the existing logic and only store the transferred amount in case 2).

pallets/redeem/src/lib.rs
b-yap: changed it

TorstenStueber:
Now @b-yap removed the line from the case the vault is liquidated. 

What happens now if the vault is liquidated, below a secure threshold when the redeem is cancelled and the user wants to reimburse? Is this a case that can happen?

I unfortunately need more time to understand these cases.

ebma:
@b-yap @TorstenStueber how should we continue with this?

TorstenStueber:
I will add my comment once done with the fee model. Stay tuned, almost there.

ebma:
@TorstenStueber this is the PR that we talked about in today's sync meeting, blocking Spacewalk's public launch. We still need to find some action here.

TorstenStueber:
Ah understood, I give this now highest priority.

pallets/redeem/src/lib.rs
TorstenStueber: The main confusion here arises because the collateral sent from the vault account to the redeemer (`amount_to_slash`) is a collateral amount, however, the `reimbursed_amount` is a wrapped token amount.

The implementation right now is already wrong because `CancelledRedeemAmount` stores the actually transferred collateral amount (the saturated `amount_to_slash`) but uses that as the reimbursement amount in `_mint_tokens_for_reimbursed_redeem` – this is wrong independent of the liquidated/non-liquidated situation we discussed above.

A more fundamental problem already arises in `_cancel_redeem` and not just in `_mint_tokens_for_reimbursed_redeem`: if the redeemer wishes a reimburse and the vault is not undercollateralized (see [this code](https://github.com/pendulum-chain/spacewalk/pull/343/files/f4f92705e0a716f60c924caba5c8e756f5e9a041#diff-066389da98714f72bf9e0f00533ef0829d9181607572e15e866b3285c08841a6R885-R892)), then all the wrapped tokens of the redeemer are immediately transferred to the vault. Also here one could argue that this is unfair to the user if the collateral asset dropped sharply in price when `_cancel_redeem` is executed (i.e., the same argument as in [the audit ticket](https://skyharbor.certik.com/report/1388e1a1-81d0-4c89-812e-c6a3e209bb0c?findingIndex=LI7-04)).

So in order to be completely fair to the user, we should immediately calculate the equivalent wrapped token amount from the actually transferred amount [after the `transfer_funds_saturated`](https://github.com/pendulum-chain/spacewalk/pull/343/files/f4f92705e0a716f60c924caba5c8e756f5e9a041#diff-066389da98714f72bf9e0f00533ef0829d9181607572e15e866b3285c08841a6R849):

```
let transferred_amount = ext::vault_registry::transfer_funds_saturated::<T>(...)
let transferred_amount_in_wrapped =
transferred_amount.saturating_sub(&punishment_fee_in_collateral).convert_to(vault_id.wrapped_currency())?;
```

Then we should:
* for the undercollateralized case: store that value to be used in the `_mint_tokens_for_reimbursed_redeem` function 
* for the non-undercollateralized case: only [send that amount to the vault](https://github.com/pendulum-chain/spacewalk/pull/343/files/f4f92705e0a716f60c924caba5c8e756f5e9a041#diff-066389da98714f72bf9e0f00533ef0829d9181607572e15e866b3285c08841a6R887) account after unlocking

## Liquidated case
This situation is quite different as the [collateral amount sent to the user](https://github.com/pendulum-chain/spacewalk/pull/343/files/f4f92705e0a716f60c924caba5c8e756f5e9a041#diff-066389da98714f72bf9e0f00533ef0829d9181607572e15e866b3285c08841a6R823-R827) (`confiscated_collateral`) is calculated using values that are fixed in the moment the vault turned into the liquidated state. This whole situation assumes that as long as the liquidated state last, the prices don't change. So for that reason I would use the simple solution to set `transferred_amount_in_wrapped` to `vault_to_be_burned_tokens` – exactly as it already is.

TorstenStueber (changes requested):

pallets/redeem/src/lib.rs
ebma: Are you sure that it should be 
```
let transferred_amount_in_wrapped = transferred_amount.saturating_sub(&punishment_fee_in_collateral).convert_to(vault_id.wrapped_currency())?;
```
and not 
```
let transferred_amount_in_wrapped = transferred_amount.convert_to(vault_id.wrapped_currency())?;
```

Why do we need to subtract the `punishment_fee_in_collateral` from the transferred amount?

ebma: @TorstenStueber pinging you again on this.

TorstenStueber: I wrote the above answer immediately but saw that it was still pending and not shown to you because I had to submit a review first 🙄.

TorstenStueber: (@ebma)

ebma: Okay thanks, I think I got it now @TorstenStueber. I think the naming of the variable was throwing me off a bit. 
Maybe we should not call it `transferred_amount_in_wrapped` but `retrievable_amount_in_wrapped`? Or similar, I don't know. But it's not necessarily the transferred amount transformed to wrapped tokens, but less than that because it's what the vault can retrieve later on.

Or maybe it's best to add an extra variable to make it perfectly clear. Something like:
```
let transferred_amount = ext::vault_registry::transfer_funds_saturated::<T>(...)
// Subtract the punishment fee to avoid overpaying the vault when reclaiming tokens
let retrievable_amount_in_collateral = transferred_amount.saturating_sub(&punishment_fee_in_collateral);
let retrievable_amount_in_wrapped = retriveable_amount_in_collateral.convert_to(vault_id.wrapped_currency())?;
```

TorstenStueber: Absolutely, this code is annoyingly complex so I appreciate any effort to make it cleaner and easier to grasp.

ebma: Perfect. Now that we agreed on the final approach, @b-yap would you be able to implement @TorstenStueber's suggested changes in this PR?

b-yap: Okay so I lost the links after I rebased. I should not have done that prior to reading the comment.

If I understand it right @TorstenStueber, I'll try to rephrase: 
### For the liquidated case:
> So for that reason I would use the simple solution to set transferred_amount_in_wrapped to vault_to_be_burned_tokens – exactly as it already is.

instead of transferring the `confiscated_collateral`, it should be the `vault_to_be_burned_tokens` ?
So we are removing the `confiscated_collateral` in general.
```rust
let vault_to_be_burned_tokens = redeem.amount().checked_add(&redeem.transfer_fee())?;
...
ext::vault_registry::transfer_funds::<T>(
				CurrencySource::LiquidatedCollateral(vault_id.clone()),
				slashing_destination,
				- &confiscated_collateral,
				+ &vault_to_be_burned_tokens  
			)?;
```
^ is this correct? 

When it comes to `fn _mint_tokens_for_reimbursed_redeem()`, it'll be the same amount, correct? 

### For NOT liquidated case:
@ebma Marcel's `retrievable_amount_in_wrapped` is the `amount_to_slash` , AND the amount to use in `fn _mint_tokens_for_reimbursed_redeem()`

pallets/redeem/src/lib.rs
```diff
@@ -204,6 +204,10 @@ pub mod pallet {
 	#[pallet::storage]
 	pub(super) type LastIntervalIndex<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
 
+	#[pallet::storage]
+	pub(super) type CancelledRedeemAmount<T: Config> =
+		StorageMap<_, Blake2_128Concat, H256, (BalanceOf<T>, T::CurrencyId), OptionQuery>;
```
ebma, line 209: Can we store this as `Amount<T>` or would that not work?

pallets/redeem/src/lib.rs
```diff
@@ -819,10 +823,10 @@ impl<T: Config> Pallet<T> {
 			ext::vault_registry::transfer_funds::<T>(
 				CurrencySource::LiquidatedCollateral(vault_id.clone()),
 				slashing_destination,
-				&confiscated_collateral,
+				&vault_to_be_burned_tokens,
 			)?;
 
-			confiscated_collateral
+			vault_to_be_burned_tokens.clone()
```
ebma, line 829: I don't think we need to change this.

ebma (commented):
It's really hard to follow now because the old commits are lost after the re-basing 😅 

@TorstenStueber can you check the changes again and see if it addresses what you mentioned in your other review?

pallets/redeem/src/lib.rs
TorstenStueber: Yes, I think so, too. The amount given to the function `decrease_liquidated_collateral` should be equal to the one given to `transfer_funds`, otherwise some invariant is violated. This [invariant](https://www.notion.so/satoshipay/Vault-Registry-Pallet-b4a52755953e4db8bdac937be96a7406?pvs=4#a67ed18e9dc549919d05764a71d4e47a) is not mentioned explicitly but must be kept true in my opinion for the vault registry to work correctly.

TorstenStueber (changes requested):
I request a change in one comment. The rest looks good to me.

pallets/redeem/src/lib.rs
b-yap: No, that won't work
```rust
 the trait `WrapperTypeEncode` is not implemented for `currency::Amount<T>`

```

ebma (approved):

TorstenStueber (approved):

PR #345, opened 2023-05-03
Pad CurrencyId slices with space characters instead of '0' bytes

### LINKED ISSUE #344: Fix 'auto-register' not working properly for Stellar assets with less than 4 characters

ebma:
The code field of Stellar currency IDs needs to have exactly 4 or 12 bytes. So far, we were padding with the `0` byte, ie. 'null' which makes sense. However, it's not easy to put in this 0 byte when using the UI to enter codes in eg. polkadot.js that are less than 4 or 12 characters. That's why, now instead of prefilling the 4 or 12 bytes with `0` bytes, we now fill them with the space character.

Closes #344.


ebma:
Closing because we are planning to use another approach.

PR #347, opened 2023-05-08
remove trailing non-alphanumeric characters for AlphaNum4

### LINKED ISSUE #346: Add integration test for bridging XLM and a Stellar asset with a 3 character code
So far we only bridged USDC from Stellar to spacewalk chains. We want to make sure that besides USDC, we are also able to bridge XLM (ie `StellarNative` internally) and a Stellar asset that only has a code of 3 character length (+ a \0 byte, so 4 bytes in total).

We should consider adding tests for these two assets to the integration tests. Because we should add an asset that DIA has prices for, I would suggest we use 'MXN' as the three-character asset, issued by the same testnet Stellar account that also issues our USDC assets. 

---
For this to work, we need to:
- [ ] Add the following to the [oracle_keys](https://github.com/pendulum-chain/spacewalk/blob/e76f4d46ce59811c00f557e74593f22e3a304d66/testchain/node/src/chain_spec.rs#L322)
```
Key::ExchangeRate(CurrencyId::AlphaNum4(
					*b"MXN\0",
					[
						20, 209, 150, 49, 176, 55, 23, 217, 171, 154, 54, 110, 16, 50, 30, 226,
						102, 231, 46, 199, 108, 171, 97, 144, 240, 161, 51, 109, 72, 34, 159, 139,
					],
				)),
```
- [ ] Add the following to [supported_currencies](https://github.com/pendulum-chain/spacewalk/blob/e76f4d46ce59811c00f557e74593f22e3a304d66/testchain/node/src/chain_spec.rs#L387)
```
AssetId::new(b"Stellar".to_vec(), b"XLM".to_vec()),
```
- [ ] (Probably) we have to change the conversion function of our DIA converter. I assume that we need to remove the `\0` bytes from the vector [here](https://github.com/pendulum-chain/spacewalk/blob/e76f4d46ce59811c00f557e74593f22e3a304d66/pallets/oracle/src/dia.rs#L24). Otherwise the created symbol might be off and the conversion not working properly. Maybe also rename the function
- [ ] In the integration tests, we then need to feed prices also for the new MXN and XLM asset. They should be appended [here](https://github.com/pendulum-chain/spacewalk/blob/e76f4d46ce59811c00f557e74593f22e3a304d66/clients/vault/tests/vault_integration_tests.rs#L139-L152). Using the same price as for the wrapped USDC asset should be fine (ie. 1/10)

b-yap:
closes #346 

test 3-letter codes for alphanum4


primitives/src/lib.rs
```diff
@@ -408,6 +408,15 @@ pub trait CurrencyInfo {
 	fn decimals(&self) -> u8;
 }
 
+pub fn remove_trailing_zeroes(input:&[u8]) -> &[u8] {
```
ebma, line 411: Would rename to 'remove_trailing_null_bytes' because it's more accurate

primitives/src/tests.rs
```diff
@@ -7,6 +7,20 @@ use super::{CurrencyId, *};
...
+#[test]
+fn test_trailing_non_alphanum() {
+	let test = b"ZerO";
+	assert_eq!(remove_trailing_non_alphanum_bytes(test).len(), test.len());
+
+	let test = b"Zer0";
+	assert_eq!(remove_trailing_non_alphanum_bytes(test).len(), test.len());
+
+	let test = b"Zer\0";
+	assert_ne!(remove_trailing_non_alphanum_bytes(test).len(), test.len());
```
ebma, line 19: Nit: I was confused at first because the 'assert_**ne**' is easy to miss so you could have used the same structure as in l.22. 
But it really does not matter too much 😅

primitives/src/lib.rs
```diff
@@ -408,6 +408,15 @@ pub trait CurrencyInfo {
 	fn decimals(&self) -> u8;
 }
 
+pub fn remove_trailing_non_alphanum_bytes(input: &[u8]) -> &[u8] {
+	for (idx, elem) in input.iter().enumerate().rev() {
+		if elem.is_ascii_alphanumeric() {
+			return &input[..idx + 1]
```
TorstenStueber, line 414: Alternatively, `[..=idx]`

TorstenStueber (approved):

primitives/src/tests.rs
b-yap: changed

primitives/src/lib.rs
b-yap: :open_mouth:

PR #349, opened 2023-05-11
Waive fee for successful execution of `execute_issue()` extrinsic

### LINKED ISSUE #348: Don't charge fees for successful execution of `execute_{issue/redeem}` extrinsics
# Problem
We identified that the proofs for fulfilling issue and redeem requests can be quite large. 
To give an example, we examined the attached extrinsic and found that:
- the extrinsic is 349KB
- it's main payload is the TransactionSet with 337 KB
- the three arguments that the function takes are all base64 encoded strings instead of the actual binary data; we could save 33% of space if we use raw binary data; inside the extrinsic it is base64 decoded anyway so that this step would not be necessary
- this way the actual TransactionSet is "only" 253KB
- Stellar has no size limits for blocks (ledgers), only operation limits, namely 1000 operations at most
- this TransactionSet almost reaches that limit: it contains 453 transactions and they contain 999 operations in total
some transactions have more than 80 operations, all of them are ManageSellOffer or ManageBuyOffer operations, probably some trading bots that insert these massive transactions
- many transactions also have 2 signatures, that alone is 80 bytes

# Solution
Since we will not be able to reduce the size of this extrinsic by large margin soon, we decided to waive the gas fees upon successful execution of the `execute_issue()` and `execute_redeem()` extrinsic. This is the supposed outcome of this ticket.

ebma:
This changes the `execute_issue()` extrinsic to not charge a fee upon successful execution. 
The `execute_redeem()` function was not changed because it actually already implements this fee waiving mechanism. 

Closes #348.

### DIFF
--- a/pallets/issue/src/lib.rs
+++ b/pallets/issue/src/lib.rs
@@ -282,7 +282,10 @@ pub mod pallet {
 				transaction_set_encoded,
 			)?;
 
-			Ok(().into())
+			// Don't take tx fees on success. If the vault had to pay for this function, it would
+			// have been vulnerable to a griefing attack where users would issue amounts just
+			// above the minimum transfer value.
+			Ok(Pays::No.into())
 		}
 
 		/// Cancel the issuance of tokens if expired


TorstenStueber (commented):
@ebma can we ensure that for every issue there can always be one successful call to `execute_issue`?

PR #350, opened 2023-05-25
upgrade to polkadot-v0.9.40

b-yap:
partially closes [pendulum issue 219](https://github.com/pendulum-chain/pendulum/issues/219).

I have updated the weights as well, since a lot of the pallets are `#![deny(warnings)]`. 

## summary of changes: 
1. [rename grandpa traits](https://github.com/paritytech/substrate/commit/7c8b89b3a6b603902709fc33fdb10a864b8b4204)
2. [deprecated trait store](https://github.com/paritytech/substrate/pull/13535)
3. [hash instead of BlockId](https://github.com/paritytech/substrate/pull/13255)
4. [deprecated `from_ref...`, changed to `from_parts()` function](https://github.com/paritytech/substrate/pull/13475)

Specific to clients directory:
1. [Rename *-private-ipv4 to *-private-ip CLI args](https://github.com/paritytech/substrate/pull/13208)

Specific to testchain directory:
1. update batching api in chain spec
2. [rename `warp_sync` to `warp_sync_params`](https://github.com/paritytech/substrate/pull/12761)
3. [Extract syncing protocol from `sc-network`](https://github.com/paritytech/substrate/pull/12828)
4. [Add `WeightToFee` and `LengthToFee` impls to transaction-payment Runtime API](https://github.com/paritytech/substrate/pull/13110)
5. [`KeyOwnerProofSystem` of grandpa removed](https://github.com/paritytech/substrate/pull/13425)

extra:
1. cleanup based on [this suggestion](https://github.com/actions/runner-images/issues/2840#issuecomment-790492173).
2. fix failing `test_check_slot_position` test case after updating the `MAX_SLOTS_TO_REMEMBER` value.


b-yap:
@ebma the weights have increased. You can check each extrinsic's comments from the benchmark. 
I can always replace `from_ref_time` with `from_parts`, if you find the numbers too big/unreasonable.

pallets/fee/src/default_weights.rs
```diff
@@ -1,33 +1,37 @@
...
 // Executed Command:
 // ./target/release/spacewalk-standalone
 // benchmark
 // pallet
-// --chain=dev
-// --pallet=fee
-// --extrinsic=*
+// --chain
+// dev
+// --execution=wasm
```
ebma, line 16: Wow, since when does it work with `wasm`? That's great. I remember when I created all the benchmarks a while ago this parameter was not working and unfortunately I was not able to figure out why. Great that it works now

pallets/issue/src/benchmarking.rs
```diff
@@ -64,6 +64,8 @@ benchmarks! {
 	request_issue {
 		let origin: T::AccountId = account("Origin", 0, 0);
 		let vault_id = get_vault_id::<T>();
+		// this amount is when running the benchmark with the testchain
+		// let amount = 1000_0000u32.into();
```
ebma, line 68: Why this comment? I need to change the amount when running benchmarks with the testchain but use the `1000` otherwise? Can we change it such that it always uses `100_00000u32` then so that it works in both cases?

testchain/node/src/chain_spec.rs
```diff
@@ -451,7 +451,7 @@ fn testnet_genesis(
 				AssetId::new(b"FIAT".to_vec(), b"MXN-USD".to_vec()),
 				AssetId::new(b"Stellar".to_vec(), b"XLM".to_vec()),
 			],
-			batching_api: b"http://localhost:8070/currencies".to_vec(),
+			batching_api: b"http://dia-00.pendulumchain.tech:8070/currencies".to_vec(),
```
ebma, line 454:
```suggestion
			batching_api: b"https://dia-00.pendulumchain.tech/currencies".to_vec(),
```

testchain/runtime/src/lib.rs
```diff
@@ -176,16 +176,17 @@ impl pallet_aura::Config for Runtime {
...
-		GrandpaId,
-	)>>::IdentificationTuple;
-	type KeyOwnerProofSystem = ();
-	type HandleEquivocation = ();
+	// type KeyOwnerProof =
+	// 	<Self::KeyOwnerProofSystem as KeyOwnerProofSystem<(KeyTypeId, GrandpaId)>>::Proof;
+	// type KeyOwnerIdentification = <Self::KeyOwnerProofSystem as KeyOwnerProofSystem<(
+	// 	KeyTypeId,
+	// 	GrandpaId,
+	// )>>::IdentificationTuple;
```
ebma, line 184: Can we remove the comment?

ebma (approved):
Thanks for adding all these changes @b-yap 🙏 

I see that the weights changed indeed. Sometimes the execution time varies just a little and sometimes the weights doubled (which might look confusing at first since the weight comments now state everything in pico seconds, but since pico is 1e-3 smaller than nano and the values mostly have three additional 0s now, they are pretty similar.) But it could be that the fact that you were now using `--execution=wasm` also amplified this variation. It's great though that you managed to use that parameter as it's preferred and should be more accurate AFAIK. I only did not use it before because it did not work for some reason.

Maybe we can fix some of the CI warnings but besides that it's good to go 👍

TorstenStueber (approved):

pallets/issue/src/benchmarking.rs
b-yap: changed amount to the fixed `1000_0000u32.into()`

PR #352, opened 2023-05-26
filter archive for `ScpStExternalize` only

### LINKED ISSUE #351: Execution of issue requests failing with `InvalidScpPledge` error
Sometimes the execution of issue requests fails with the following error:
```
vault::issue: Task with slot 46445303 has failed: "SubxtRuntimeError(Runtime(Module(ModuleError { pallet: \"StellarRelay\", error: \"InvalidScpPledge\", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [5, 0, 0, 0] } })))"
```

image

This was encountered for the XLM vault. Restarting the vault also does not fix the issue because it encounters the same error again and again. 

We need to investigate the issue further. Maybe the vault is not sending enough SCP messages. Maybe the SCP messages from the archive are wrong?

b-yap:
closes #351

### DIFF
--- a/clients/stellar-relay-lib/examples/connect.rs
+++ b/clients/stellar-relay-lib/examples/connect.rs
@@ -16,7 +16,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 	let (cfg_file_path, sk_file_path) = if arg_network == "mainnet" {
 		(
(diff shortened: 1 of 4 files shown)


ebma:
We are probably not gonna need this (or at least not only this but way more) once we change the `stellar-relay` pallet to also be able to use `Confirm` SCP messages for proving transaction validity. So I'm closing this for now.

PR #355, opened 2023-05-30
increase value of `MAX_SLOTS_TO_REMEMBER` to 14400

### LINKED ISSUE #354: Increase time window for fetching SCP messages from overlay
To increase the time window where our vault clients can fetch SCP messages from the overlay instead of from archive nodes, we will change the [MAX_SLOTS_TO_REMEMBER](https://github.com/stellar/stellar-core/blob/ca2fb06059c15442cb4c9a8c89de1a8fc3579a39/docs/stellar-core_example.cfg#L478) to **`14400`**.
By changing it to `14400` we increase the time window to 24 hours.

We need to update [this](https://github.com/pendulum-chain/spacewalk/blob/befd6fe8de903407e212902956398df25b50e8f3/clients/vault/src/oracle/constants.rs#L19) constant to that same value once done.

b-yap:
closes #354

### DIFF
--- a/clients/vault/src/oracle/constants.rs
+++ b/clients/vault/src/oracle/constants.rs
@@ -16,7 +16,9 @@ pub const MAX_DISTANCE_FROM_CURRENT_SLOT: Slot = 3;
 pub const VALIDATOR_COUNT_TEST_NETWORK: usize = 3;
 pub const VALIDATOR_COUNT_PUBLIC_NETWORK: usize = 23;
 
-/// increase the time window to 24 hours.
+/// Set the _expected_ `MAX_SLOTS_TO_REMEMBER` parameter to 14400 Slots on Stellar,
+/// ie. 24 hours. This parameter is only correct when connecting to SatoshiPay Stellar validators
+/// as the configuration of these nodes deviates from the default configuration.
 pub const MAX_SLOTS_TO_REMEMBER: Slot = 14400;
 
 pub const ARCHIVE_NODE_LEDGER_BATCH: Slot = 64;


ebma (changes requested):
We'll need to increment this value even further, just not yet sure to what exactly.

TorstenStueber:
As a reference: https://github.com/pendulum-chain/spacewalk/issues/354

ebma:
I updated the original ticket. We want to increase it to `14400` now. But we'll first have to check if the archive nodes break down because of that.

clients/vault/src/oracle/constants.rs
```diff
@@ -16,7 +16,8 @@ pub const MAX_DISTANCE_FROM_CURRENT_SLOT: Slot = 3;
 pub const VALIDATOR_COUNT_TEST_NETWORK: usize = 3;
 pub const VALIDATOR_COUNT_PUBLIC_NETWORK: usize = 23;
 
-pub const MAX_SLOT_TO_REMEMBER: Slot = 12;
+/// increase the time window to 24 hours.
+pub const MAX_SLOTS_TO_REMEMBER: Slot = 14400;
```
ebma, line 20: Nice. Let's maybe rewrite the comment though to give a bit more context.
```suggestion
/// Set the _expected_ `MAX_SLOTS_TO_REMEMBER` parameter to 14400 Slots on Stellar, 
/// ie. 24 hours. This parameter is only correct when connecting to SatoshiPay Stellar validators
/// as the configuration of these nodes deviates from the default configuration.
pub const MAX_SLOTS_TO_REMEMBER: Slot = 14400;
```

ebma (approved):
Looks good. We could already merge this even though the parameter is not accurate (ie. not matched by our deployed nodes). But the archive logic is broken anyways, so there is probably no harm done.

clients/vault/src/oracle/constants.rs
b-yap: https://github.com/pendulum-chain/spacewalk/pull/355/commits/d7e87245e8192f6599e87e334ce650a8187f1e01
done

TorstenStueber (commented):
👍

Issue #353, opened 2023-05-26
Extend the `CurrencyId` enum with a LP token type


ebma:
In order to support Zenlink on our parachains, we need to add an LP token to our `CurrencyId`. While this is unrelated to Spacewalk itself, adding this type here is necessary because we share the `CurrencyId` definition in both projects.

## What to do
Add a type `ZenlinkLPToken` to the `CurrencyId` enum. This can be copied either from [here](https://github.com/zenlinkpro/Zenlink-DEX-Module/blob/7662acbff71cea3b35db918ceb9370e8929be991/example/primitives/src/lib.rs#L83) or [here](https://github.com/bifrost-finance/bifrost/blob/0eb697cc517daf9103982a6be80d6e7ea8f7e458/node/primitives/src/currency.rs#L354).

PR #359, opened 2023-06-09
Add ZenlinkLPToken to CurrencyId enum

adelarja:
I'm implementing a draft for [issue1](https://github.com/pendulum-chain/spacewalk/issues/353) and [issue2](https://github.com/pendulum-chain/pendulum/issues/241) tickets.

I've started implementing the ZenlinkLPToken enum value like this:
```rust
pub enum CurrencyId {
	Native = 0_u8,
	XCM(u8),
	Stellar(Asset),
	ZenlinkLPToken(u8, u8),
}
```

Bifrost is using this enum value:
```rust
// [currency1 Tokensymbol, currency1 TokenType, currency2 TokenSymbol, currency2 TokenType]
LPToken(TokenSymbol, u8, TokenSymbol, u8),
```

~I'm not entirely sure about why Bifrost is using this enum value. But it seems that they use a TokenType that could be:~

~For that they are defining the enum value like this (or at least I think so, I will keep researching a little bit). They take the token type into account when calculating the asset_index [here](https://github.com/bifrost-finance/bifrost/blob/0eb697cc517daf9103982a6be80d6e7ea8f7e458/node/primitives/src/currency.rs#LL85C11-L85C11). I think we don't need this (?)~

Maybe Bifrost is using that enum value for something related to what I mentioned [here](https://github.com/pendulum-chain/pendulum/issues/201#issuecomment-1589883661).

After adding the enum, I had to perform the following tasks:
* I made some changes to the following files:
   * pallets/oracle/src/dia.rs
   * pallets/oracle/src/oracle_mock.rs
* Additionally, I had to modify some match statements in primitives/lib.rs

I have implemented TryFrom<ZenlinkAssetId> for CurrencyId. It is currently a draft, and I need to work on improving its readability and qualitv.


primitives/src/lib.rs
```diff
@@ -553,6 +559,74 @@ impl TryFrom<(&str, AssetIssuer)> for CurrencyId {
 	}
 }
 
+
+pub type ZenlinkAssetId = zenlink_protocol::AssetId;
+const LP_DISCRIMINANT: u64 = 6u64;
+const PARA_CHAIN_ID: u32 = 1000;
+
+impl TryFrom<CurrencyId> for ZenlinkAssetId {
```
adelarja, line 567: This needs to be changed to something like [this](https://github.com/bifrost-finance/bifrost/blob/0eb697cc517daf9103982a6be80d6e7ea8f7e458/node/primitives/src/currency.rs#LL78C8-L78C22)?

primitives/src/lib.rs
```diff
@@ -553,6 +559,74 @@ impl TryFrom<(&str, AssetIssuer)> for CurrencyId {
...
+						()
+					})?;
+
+					return Ok(CurrencyId::XCM(foreign_id));
+				}
+				else {
+					let discriminant = (asset_id.asset_index & 0x0000_0000_0000_ff00) >> 8;
+					return if discriminant == LP_DISCRIMINANT {
+						let token0_id = ((asset_id.asset_index & 0x0000_0000_ffff_0000) >> 16) as u8;
+						let token1_id = ((asset_id.asset_index & 0x0000_ffff_0000_0000) >> 32) as u8; // Should it be >> 32? It was a 16 bits right shift in the example, but i think it was wrong.
```
ebma: I guess it should actually be `16` but would have to think about the details again. But since they are using that value on their chain in practice I guess we could be quite confident that it works as intended.

ebma (commented):
For the `TryFrom` implementation and everything that is closely related to Zenlink needs to move to the 'pendulum' repository and would then be part of solving [this](https://github.com/pendulum-chain/pendulum/issues/241) ticket. 

We shouldn't even need any Zenlink dependencies here. We just need to add the `ZenlinkLPToken(u8, u8)` to the enum (and fix the resulting compiler errors) as you did.

So I would recommend you move parts of the changes to a new PR in the 'pendulum' repo that is about solving the other task :)

primitives/src/lib.rs
adelarja: Thanks for the comment!

I referenced these lines from the [example code](https://github.com/zenlinkpro/Zenlink-DEX-Module/blob/7662acbff71cea3b35db918ceb9370e8929be991/example/primitives/src/lib.rs):

When implementing the [TryFrom trait for ZenlinkAssetId](https://github.com/zenlinkpro/Zenlink-DEX-Module/blob/7662acbff71cea3b35db918ceb9370e8929be991/example/primitives/src/lib.rs#LL107C4-L113C5), they shift the values of the discriminant by 8, symbol0 by 16, and symbol1 by 32. The sum of these values represents the asset_index.

However, during the implementation of the [TryFrom trait for CurrencyId](https://github.com/zenlinkpro/Zenlink-DEX-Module/blob/7662acbff71cea3b35db918ceb9370e8929be991/example/primitives/src/lib.rs#LL128C4-L129C81), they apply the following masks and shifts:

```rust
let token0_id = ((asset_id.asset_index & 0x0000_0000_ffff_0000) >> 16) as u8;
let token1_id = ((asset_id.asset_index & 0x0000_ffff_0000_0000) >> 16) as u8;
```
Based on this code, it appears that token1_id would still be shifted by 16 bits. Additionally, if I'm not mistaken, the result of `(asset_id.asset_index & 0x0000_ffff_0000_0000) >> 16` could potentially exceed the range for u8.

ebma: Hmm, maybe it's easiest to check which implementation is correct if we just do the 'calculations' based on example values. Could you maybe try both approaches (ie. the one with `>> 16` and the one with `>> 32`)? I mention some example IDs (like an example value for a LP token based on the default generation function) in [this](https://satoshipay.slack.com/archives/C01V1F56RMJ/p1684949005110329) message on slack. So maybe you could try how this turns out when we use `(2124,0,0)`, `(2124,0,1)` for base tokens and assume `(2124,0,4294968832)` is the generated LP token.

ashneverdawn: In terms of the calculation,
asset_id.asset_index is for sure at least a u64 (or this wouldn't compile because the second mask is 64bit)

This should illustrate the behavior this has:
```
    let asset_index = 0x1234_5678_9012_3456_u64;
    let token0_id = (asset_index & 0x0000_0000_ffff_0000) >> 16;
    let token1_id = (asset_index & 0x0000_ffff_0000_0000) >> 16;
    println!("token0_64: {:#x}", token0_id);
    println!("token1_64: {:#x}", token1_id);
    println!("token0_u8: {:#x}", token0_id as u8);
    println!("token1_u8: {:#x}", token1_id as u8);
```
output:
```
token0_64: 0x9012
token1_64: 0x56780000
token0_u8: 0x12
token1_u8: 0x0
```

So, token1_id will always be zero in the end.

So it would make sense if token1 has to be shifted by 32, not 16:
```
    let asset_index = 0x1234_5678_9012_3456_u64;
    let token0_id = (asset_index & 0x0000_0000_ffff_0000) >> 16;
    let token1_id = (asset_index & 0x0000_ffff_0000_0000) >> 32;
    println!("token0_64: {:#x}", token0_id);
    println!("token1_64: {:#x}", token1_id);
    println!("token0_u8: {:#x}", token0_id as u8);
    println!("token1_u8: {:#x}", token1_id as u8);
```
output:
```
token0_64: 0x9012
token1_64: 0x5678
token0_u8: 0x12
token1_u8: 0x78
```

basically, this code is fetching specific bytes inside the asset_index.

clients/README.md
```diff
@@ -86,6 +88,18 @@ subxt metadata -f bytes > runtime/metadata-{your-chain-name}.scale
...
+After making the changes, the first step is to run the local node.
+
+```bash
+cargo run --bin spacewalk-standalone --release -- --dev
+```
+
+Once the local node is running, run this command:
+
+```bash
+subxt metadata -f bytes > runtime/metadata-{your-chain-name}.scale
```
ebma, line 100: It will always be the standalone one with the local node, so we can make it explicit in this command.
```suggestion
subxt metadata -f bytes > runtime/metadata-standalone.scale
```

adelarja:
After analyzing how to generate ZenlinkLPTokens, we have decided to utilize the following enum value

```rust
pub enum CurrencyId {
	Native = 0_u8,
	XCM(u8),
	Stellar(Asset),
	ZenlinkLPToken(u8, u8, u8, u8),
}
```
[Here](https://github.com/pendulum-chain/pendulum/issues/201#issuecomment-1589883661) and [here](https://github.com/pendulum-chain/pendulum/issues/201#issuecomment-1591840337) you can find explanations on why we opted for this enum value.

adelarja:
For some reason, something is breaking in the CI pipeline. [Here](https://github.com/pendulum-chain/spacewalk/actions/runs/5313734815/jobs/9620028334#step:10:2552) is one of the tests that is failing.

It appears that a request is failing and we are receiving a `HorizonSubmissionError`. Do you have any ideas about what might be happening here?

ebma:
@adelarja I noticed this morning that there was a [Stellar testnet reset](https://docs.blockdaemon.com/docs/stellar-quarterly-testnet-reset) recently. I spent some time to re-do the setup, so I will now re-run your CI job and we'll see if it's fixed now.

primitives/src/lib.rs
```diff
@@ -482,7 +483,7 @@ impl CurrencyId {
 		match self {
 			CurrencyId::Stellar(asset) => asset.decimals(),
 			// We assume that all other assets have 12 decimals
-			CurrencyId::Native | CurrencyId::XCM(_) => 12,
+			CurrencyId::Native | CurrencyId::XCM(_) | CurrencyId::ZenlinkLPToken(_, _, _, _) => 12,
```
TorstenStueber, line 486: Strictly speaking this is not true anymore as USDT has now only 6 decimals. (shall we change the XCM definition to extend to 12 decimals instead @ebma?)

ebma: I think we should, yes. We'd require a migration since the current USDT we have are already bridged using only 6 decimals.
So I guess we can leave this as is.

ebma:
@adelarja great that you managed to fix the CI job.

ebma (approved):
Ready for merge IMO

PR #362, opened 2023-06-13
360 support confirm scp messages in the stellar relay when proving transaction validity

### LINKED ISSUE #360: Support 'confirm' SCP messages in the stellar-relay when proving transaction validity
# Problem
With the current implementation, the `stellar-relay` pallet only considers `ScpStatementExternalize` consensus messages to prove the validity of a Stellar transaction. However, we identified that Stellar archive entries usually only contain just **one** externalized message but many other 'ScpStatementConfim` or `ScpStatementPrepare` messages. This means that we currently cannot use archive entries to submit 'execute...' extrinsics, since this one externalized message is not enough.

# Solution
We want to change the validation logic used in the `stellar-relay` pallet to also consider `ScpStatementConfirm` messages. If we combine all 'confirm' and 'externalize' messages contained in archive entries, we should be build valid proofs even when missing the relevant consensus messages when they are emitted on Stellar's overlay network.

The reasoning why we can do this can be found on [this](https://www.notion.so/satoshipay/Archived-SCP-Messages-6801530d58d44f1fba1208db27ddee03#b052bb35b2fa416ea69034c5a2b34ae9) Notion page. For our implementation, we will follow the 'simplified approach' described [here](https://www.notion.so/satoshipay/Archived-SCP-Messages-6801530d58d44f1fba1208db27ddee03?pvs=4#b052bb35b2fa416ea69034c5a2b34ae9). 

### Steps
Note: The structure of `ScpStatements` can be found [here](https://stellar-sdk.readthedocs.io/en/stable/api.html#scpstatement). This is relevant for the next steps.  

Extend the logic of [this](https://github.com/pendulum-chain/spacewalk/blob/06873f16c4a7ab884e4759d98b7b86c712e0c0ae/pallets/stellar-relay/src/lib.rs#L577-L586) `match` statement:
- [ ] Add a new branch for `ScpStConfirm` pledges. Add logic that is very similar to the check for externalized messages, ie. check that the transaction set hash contained in the `ScpBallot` of the pledge matches the one we expect. **Note** that you might have to slightly adapt the `get_tx_set_hash` function because the names for the ballots differ between 'externalize' and 'confirm' messages (ie. it's either `commit` or `ballot`). 
- [ ] Ensure that all statements contained in the envelopes refer to the same `nH` value. 

By checking if the transaction set hash contained in the externalized/confirmed ballot matches the expected, we make sure that the consensus messages are related to the same value and the value we are looking for simultaneously.

ebma:
I tried to adhere to the simplified approach described in [this](https://www.notion.so/satoshipay/Archived-SCP-Messages-6801530d58d44f1fba1208db27ddee03#66776f6d3f8241bc9be31937e09b2404) document.

The implementation deviates from the described approach as follows:
- Instead of checking if the `n_h` value lies within a certain range, the implementation takes the assumption that this value will be the same for all SCP messages.
- Instead of using the quorum hash and comparing it to hashes of existing quorum sets, we use the existing implementation that uses the stored in the on-chain storage to determine if the SCP messages originate from nodes that build a valid quorum set. 

I had to change some of the test helper functions so that it's easier to build externalized and confirmed messages in one go, without having to duplicate code. 

I recommend reading the Notion page and looking at the test cases to see if everything is covered.

Closes #360.


ebma:
> Instead of checking if the n_h value lies within a certain range, the implementation takes the assumption that this value will be the same for all SCP messages.

@TorstenStueber can I take this assumption or do I need to implement the range like you describe in the Notion page (and implemented [here](https://github.com/TorstenStueber/scp-messages/blob/b953b4ba74ac474ae725abda2d63940af5b68e31/src/checkConsensus.ts#L199-L205)).

TorstenStueber:
Turns out that it is not possible. Ledger 46647565 of the Stellar archive contains the following messages:

```
  Externalize message from FranklinTempleton3
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from Lobster4
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from BlockDaemon1
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from Coinqvest2
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from FranklinTempleton1
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from Coinqvest3
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from PublicNode1
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from Lobster1
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from SDF1
    quorumSet: quorumHash1
    c = (1, value1), h.n = 4294967295
  Externalize message from PublicNode3
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from SDF2
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from FranklinTempleton2
    quorumSet: quorumHash1
    c = (1, value1), h.n = 1
  Externalize message from Lobster3
    quorumSet: quorumHash1
```

Here `n_h` is `4294967295` for one of the nodes.

Luckily the same ledger looks different for the SatoshiPay archive:

```
  Confirm message from BlockDaemon1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SDF3
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Coinqvest2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SatoshiPay3
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from FranklinTempleton1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Coinqvest3
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from PublicNode1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SatoshiPay1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Lobster1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SDF2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from FranklinTempleton2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from PublicNode2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Coinqvest1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Lobster2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Externalize message from GDGOUEQK5G6QINBH7277DC2MHVI4KXLAKEIB2OHN572AAONKLE5IMRRE
    quorumSet: quorumHash2
    c = (1, value1), h.n = 1
```

So we could fallback to another archive in this case again. Please check, though, whether the `h.n` messages really agree and then fallback.

ebma:
@TorstenStueber thanks for investigating. Since the fallback you mentioned would have to happen in the vault client though, we don't need to consider it here probably and can live with taking the assumption of having the same n.h for all messages. It's very important though to keep it in mind for the vault client.

TorstenStueber:
Okay, but the runtime should still check equality of the `nH`, not?

pallets/stellar-relay/src/lib.rs
```diff
@@ -565,27 +584,60 @@ pub mod pallet {
...
+				} else {
+					externalized_value = Some(value);
+				}
+
+				// Check if the externalized nh is the same for all envelopes
+				if let Some(externalized_nh) = &externalized_n_h {
+					ensure!(externalized_nh == &n_h, Error::<T>::ExternalizedNHMismatch);
+				} else {
+					externalized_n_h = Some(n_h);
 				}
```
ebma, line 635: @TorstenStueber I added a check for this here, or are we talking about different things?

pallets/stellar-relay/src/lib.rs
```diff
@@ -565,27 +584,60 @@ pub mod pallet {
 
 				ensure!(node_id_found, Error::<T>::EnvelopeSignedByUnknownValidator);
```
TorstenStueber, line 585: We need to be careful here, for some strange reason the SatoshiPay archive contains messages where one of the official SatoshiPay nodes is replaced with a completely different node. I am not sure why this is happening, maybe some misconfiguration in the archive or some historical deviation that might go back many years when we setup the archives for the first time?

See this example for ledger 46647587 from one of the SatoshiPay archives (all ledger entries look like this from that archive):

```
LEDGER 46647587
ScpQuorumSets
  Quorum Set
    threshold: 5
    inner quorum set:
        threshold: 2
        validators: BlockDaemon1 BlockDaemon2 BlockDaemon3 
    inner quorum set:
        threshold: 2
        validators: SDF3 SDF1 SDF2 
    inner quorum set:
        threshold: 2
        validators: Coinqvest2 Coinqvest3 Coinqvest1 
    inner quorum set:
        threshold: 2
        validators: SatoshiPay3 SatoshiPay2 SatoshiPay1 
    inner quorum set:
        threshold: 2
        validators: FranklinTempleton1 FranklinTempleton3 FranklinTempleton2 
    inner quorum set:
        threshold: 3
        validators: Lobster5 Lobster4 Lobster1 Lobster2 Lobster3 
    inner quorum set:
        threshold: 2
        validators: PublicNode1 PublicNode3 PublicNode2 
  Quorum Set
    threshold: 5
    inner quorum set:
        threshold: 2
        validators: BlockDaemon1 BlockDaemon2 BlockDaemon3 
    inner quorum set:
        threshold: 2
        validators: SDF3 SDF1 SDF2 
    inner quorum set:
        threshold: 2
        validators: Coinqvest2 Coinqvest3 Coinqvest1 
    inner quorum set:
        threshold: 2
        validators: FranklinTempleton1 FranklinTempleton3 FranklinTempleton2 
    inner quorum set:
        threshold: 3
        validators: Lobster5 Lobster4 Lobster1 Lobster2 Lobster3 
    inner quorum set:
        threshold: 2
        validators: SatoshiPay2 SatoshiPay1 GDGOUEQK5G6QINBH7277DC2MHVI4KXLAKEIB2OHN572AAONKLE5IMRRE 
    inner quorum set:
        threshold: 2
        validators: PublicNode1 PublicNode3 PublicNode2 
LedgerScpMessages (24)
  Prepare message from Lobster5
    quorumSet: quorumHash1
    b = (1, value1), p = (0, ⊥), p' = (0, ⊥), c.n = 0, h.n = 0
  Prepare message from BlockDaemon3
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 0, h.n = 0
  Prepare message from PublicNode3
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 0, h.n = 0
  Prepare message from FranklinTempleton3
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 1, h.n = 1
  Prepare message from Lobster4
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 1, h.n = 1
  Prepare message from Coinqvest2
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 1, h.n = 1
  Prepare message from BlockDaemon2
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 1, h.n = 1
  Prepare message from SatoshiPay2
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 1, h.n = 1
  Prepare message from Lobster3
    quorumSet: quorumHash1
    b = (1, value1), p = (1, value1), p' = (0, ⊥), c.n = 1, h.n = 1
  Confirm message from BlockDaemon1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SDF3
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SatoshiPay3
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from FranklinTempleton1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Coinqvest3
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from PublicNode1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SatoshiPay1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Lobster1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SDF1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from SDF2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from FranklinTempleton2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from PublicNode2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Coinqvest1
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Confirm message from Lobster2
    quorumSet: quorumHash1
    b = (1, value1), p.n = 1, c.n = 1, h.n = 1
  Externalize message from GDGOUEQK5G6QINBH7277DC2MHVI4KXLAKEIB2OHN572AAONKLE5IMRRE
    quorumSet: quorumHash2
    c = (1, value1), h.n = 1
```

pallets/stellar-relay/src/lib.rs
```diff
@@ -565,27 +584,60 @@ pub mod pallet {
...
-							tx_set_hash == expected_tx_set_hash,
-							Error::<T>::TransactionSetHashMismatch
-						);
-					},
+				let (value, n_h) = match envelope.clone().statement.pledges {
+					ScpStatementPledges::ScpStExternalize(externalized_statement) =>
+						(externalized_statement.commit.value, externalized_statement.n_h),
+					ScpStatementPledges::ScpStConfirm(confirmed_statement) =>
+						(confirmed_statement.ballot.value, confirmed_statement.n_h),
 					_ => return Err(Error::<T>::InvalidScpPledge),
```
TorstenStueber, line 613: The archive also contains Prepare messages (see my comment above). Is the requirement here that only the Externalize and Confirm messages are allowed to be fed into this extrinsic and Prepare needs to be filtered before by the vault client?

pallets/stellar-relay/src/lib.rs
```diff
@@ -565,27 +584,60 @@ pub mod pallet {
...
 
 			// Check if transaction set matches tx_set_hash included in the ScpEnvelopes
 			let expected_tx_set_hash = compute_non_generic_tx_set_content_hash(transaction_set)
-				.ok_or(Error::<T>::FailedToComputenonGenericTxSetContentHash)?;
+				.ok_or(Error::<T>::FailedToComputeNonGenericTxSetContentHash)?;
+
+			// We store the externalized value in a variable so that we can check if it's the same
+			// for all envelopes. We don't distinguish between externalized and confirmed values as
+			// it should be the same value regardless.
+			let mut externalized_value: Option<Value> = None;
```
TorstenStueber, line 604: We can also initialize these two values from `externalized_envelope` so that they don't need to be optional – simplifies the equality check logic below a little.

TorstenStueber (commented):
Looks good in principle but I am concerned about the requirement that all messages in the archive stem from nodes in the Tier 1 set. We need to discuss what we can do here.

pallets/stellar-relay/src/default_weights.rs
```diff
@@ -1,41 +1,40 @@
...
 
 #![cfg_attr(rustfmt, rustfmt_skip)]
 #![allow(unused_parens)]
 #![allow(unused_imports)]
+#![allow(missing_docs)]
 
-use frame_support::{traits::Get, weights::Weight};
-use sp_std::marker::PhantomData;
+use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
+use core::marker::PhantomData;
```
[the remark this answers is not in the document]
ebma: Do you think this matters? It changed because the latest [template](https://github.com/paritytech/substrate/blob/2cc1667545f5e715fb94d492a2a2385ec8e00c7c/.maintain/frame-weight-template.hbs#L21) file uses `core::`

pallets/stellar-relay/src/lib.rs
ebma: >  I am not sure why this is happening, maybe some misconfiguration in the archive or some historical deviation that might go back many years when we set up the archives for the first time?

But the slot `46647587` [happened](https://stellar.expert/explorer/public/ledger/46647587) just about one week ago, right?
That's super weird. But I am honestly not too sure what we would do about it. We shouldn't really allow SCP messages signed by validators that are unknown to us and not in the tier 1 set. 

The 'unknown' node that externalized the message belongs to `quorumSet: quorumHash2` while all others belong to `quorumHash1`, what does it tell us? That it was not part of the tier 1 validator set? 

Oh but maybe it's only possible that the externalized message comes from a different node that is not in the tier 1 set. It was the same in the second snippet of your comment [here](https://github.com/pendulum-chain/spacewalk/pull/362#issuecomment-1591635924). It seems that in both cases, there are enough `Confirm` messages (ie. `14`) to build a valid quorum without considering the unknown node. So maybe the trick is not to insist on the externalized message originating from a tier 1 validator node, but also not count it in when checking the quorum validity when it's not.

pallets/stellar-relay/src/lib.rs
ebma: Right, I'll change that

pallets/stellar-relay/src/lib.rs
TorstenStueber: That's correct.

pallets/stellar-relay/src/lib.rs
ebma: How would we loosen it up though? 

> It seems that in both cases, there are enough Confirm messages (ie. 14) to build a valid quorum without considering the unknown node. So maybe the trick is not to insist on the externalized message originating from a tier 1 validator node, but also not count it in when checking the quorum validity when it's not.

Do you think this could work? Or should we not even limit it to externalized messages coming from an 'unknown' validator but we basically allow any message to originate from an unknown validator but don't count it into the tier 1 quorum validity check?

TorstenStueber: What is critical is that it seems to be always the `Externalize` message is the one that is sent from this replacement node. And the externalize message is of course the most critical one. I see only three options:
* fix our archives
* add these three replacements nodes to the list of "well-known nodes" (but that is a new notion that does not occur in the runtime yet)
* deep dive to understand how archives are composed.

Stellar now [also wants to improve this situation](https://github.com/stellar/stellar-core/issues/3769) after I inquired about the archive problem last week. But this can take a long time to fix.

I think the best approach is to fix our archives. Seems to be some misconfiguration.

ebma: Okay, let's try to do option 1 then. Although I wouldn't really know how and where to start. Would we have to ask SDF devs about this again?

TorstenStueber: I will reach out to people at Stellar.

pallets/stellar-relay/src/lib.rs
ebma: I simplified that logic in [this](https://github.com/pendulum-chain/spacewalk/pull/362/commits/2ac84397528cc74eeac4c3f8c7dc0cbe173564a8) commit

ebma:
@b-yap @TorstenStueber what do you say, can this be merged?

ebma:
Ping @b-yap @TorstenStueber

pallets/stellar-relay/src/lib.rs
```diff
@@ -601,15 +601,21 @@ pub mod pallet {
 			// We store the externalized value in a variable so that we can check if it's the same
 			// for all envelopes. We don't distinguish between externalized and confirmed values as
 			// it should be the same value regardless.
-			let mut externalized_value: Option<Value> = None;
-			let mut externalized_n_h: Option<u32> = None;
+			let (externalized_value, externalized_n_h) =
+				match &externalized_envelope.statement.pledges {
+					ScpStatementPledges::ScpStExternalize(externalized_statement) =>
+						(&externalized_statement.commit.value, externalized_statement.n_h),
+					ScpStatementPledges::ScpStConfirm(confirmed_statement) =>
```
TorstenStueber, line 608: This case cannot happen as `externalized_envelope` is define in line 567 to be an `ScpStExternalize` message.

TorstenStueber (approved):
I approve of the changes. There is a small nit I mentioned in one comment but I don't find it important enough to slow down this PR any further.

PR #365, opened 2023-06-26
Introduce `CreateClaimableBalance` and `CreateNewAccount` operations

### LINKED ISSUE #357: Use `CreateClaimableBalance` operation when executing redeem requests
## Problem 
Accounts on Stellar are inactive until they get created by sending a payment of at least 1 XLM. Inactive accounts can not hold any assets (as they cannot create trustlines to them). When a user creates a new redeem request, they can choose whatever Stellar destination account they want. The problem is that, if the destination account is inactive, the vault cannot send a payment to that account and thus can never fulfill the redeem request. 

## Solution
We can use the '[CreateClaimableBalance](https://developers.stellar.org/docs/encyclopedia/claimable-balances#create-claimable-balance)' operation to 'send' assets to an account even if it is not yet active. Since this is slightly worse UX than using a regular payment operation we should only do this as a fallback when creating a payment would not work.

We should implement the following logic when the vault client is about to create a payment transaction on Stellar for fulfilling a redeem request:
- First fetch the Stellar network for the destination account.  
   - If the request returns 404, the account is inactive 
      - If the to-be-redeemed asset is XLM and the amount is >=1, do the transfer with createAccount operation 
      - Else, fallback to a claimable balance
   - If the request returns account data, check if a trustline to the to-be-redeemed asset exists, (XLM does not need a trustline of course). If that's not the case, fallback to claimable balance
   - else, use a normal payment operation 

### Creating the 'CreateClaimableBalance' operation
The CreateClaimableBalance operation has to match what is expected by [this](https://github.com/pendulum-chain/spacewalk/blob/06873f16c4a7ab884e4759d98b7b86c712e0c0ae/pallets/currency/src/lib.rs#L145) function. Thus, it has to have an '**unconditional**' claim predicate and **exactly one** claimant

b-yap:
closes #357 

Introduce `CreateClaimableBalance` operation for redeem request.

1. created operation.rs in `wallet` lib for creating stellar operations. 
2. Broke down the whole **_horizon.rs_** into multiple fies. I find this important because the more operations we support, the more different kinds of horizon responses must be supported.  
3. introduced `get_claimable_balance(ClaimableBalanceId)` that will return the ClaimableBalance. 
5. ~~added an extra param **_`extra_operations: Vec<Operation>`_** for _`fn send_payment_to_address(...)`_ method of `StellarWallet`.~~   

     - ~~this is when `CreateAccount` operation is required on top of `Payment` operation~~
6. introduced _`send_payment_to_address_for_redeem_request()`_ which will use any of the 3 operations:
    * Normal Payment Operation (similar to `send_payment_to_address`)
    * Create Account Operation
    * Create Claimable Balance Operation
3. to convert amounts of different types, I created a new function _`fn stellar_stroops_to_u128()`_ in `primitives lib`. The `wallet` lib does not have _`sp_runtime`_ dependency to use the _`IntoAmount`_ trait of _**`BalanceConversion`**_ struct.
7. added a helper trait _`StellarTypeToString`_ in `primitives` lib to easily convert Stellar types to string

----
### Things left to do

- [x] test for ClaimableBalance
- [x] test for CreateAccount

----

## How to begin the review:
1. **`execution.rs`** of `vault` lib
   * inside `fn transfer_stellar_asset(...)`, a checking for `is_claimable_balance_op_required` is called for request_type **`Redeem`**
      ```rust
      	let result = match self.request_type {
			// only for redeem requests, check if we need extra operations
			RequestType::Redeem => match is_claimable_balance_op_required(
                         ...     
      ```

2. **`responses.rs`** of `horizon` package, `wallet` lib
   * added a method `fn is_trustline_exist()` on `HorizonAccountResponse` struct
   * renamed `Balance` to **`HorizonBalance`** (because it gets confusing with `primitive`'s _`Balance`_.
       * added a method `fn get_asset()` on `HorizonBalance` struct
   * restructured `HorizonClaimableBalanceResponse` based on [this response](https://horizon-testnet.stellar.org/claimable_balances/000000000fde1848c9d51d192751fca9c6894bcdd3f4701121e5e104962fcff026184481).

3. **`traits.rs`** of `horizon` package `wallet` lib
    * to make the functions _flexible_, ids are now of generic type `A` where _`A: StellarTypeToString<(Stellartype), Error> + Send`:_
       ```rust
       async fn get_transactions<A: StellarTypeToString<PublicKey, Error> + Send>(
		 &self,
		 account_id: A,
		 ...
	   ) 
	    ...
	   
       async fn get_claimable_balance<A: StellarTypeToString<ClaimableBalanceId, Error> + Send>(
		 &self,
		 claimable_balance_id: A,
		 ...
	  )	
	  	
       ``` 
3. **`stellar_wallet.rs`** of `wallet` lib
   * introduced `fn send_to_address(..)` (and updated `fn create_envelope(..)`) as a multipurpose  method for submitting transactions with _different_ operations.
   * ~`fn send_payment_to_address_with_extra_operations(..)` method is an extension of the current `fn send_payment_to_address(..)`~
   *  `fn send_payment_to_address_for_redeem_request(..)` method for sending payments EXCLUSIVELY for redeem request.
   * moved `fn create_payment_envelope()` method inside test module, since it's useful to test cases ONLY.
   * moved `fn create_transaction(..)` to operations.rs
   * added new test cases:
     *  _`sending_payment_using_claimable_balance_works()`_
     * _`sending_payment_using_create_account_works()`_

9. **`operations.rs`** of `wallet` lib
   * logic specific to redeem is supported with the trait `RedeemOperationsExt`:
      ```rust
       pub trait RedeemOperationsExt: HorizonClient {
	         async fn create_payment_op_for_redeem_request(
		    &self,
		    source_address: PublicKey,
		    destination_address: PublicKey,
		    is_public_network: bool,
		    to_be_redeemed_asset: Asset,
		    to_be_redeemed_amount: StellarStroops,
	        ) -> Result<Operation, Error> {
	    ...
      ```
   * test cases :
      * **`create_payment_op_for_redeem_request()`** tests the ff. logic:
         * existing account with XLM asset, use normal op
         * existing account with non-XLM asset and has a trustline, use normal op
         * existing account with non-XLM asset BUT has NO trustline, use claimable balance
         * INactive account and XLM asset of value >=1, use create account op
         * INactive account but XLM asset of value < 1, use claimable balance
         * INactive account and asset is NOT XLM, use claimable balance
   * introduce 3 operations:
     * `create_claimable_balance_operation(..`
     *  `create_account_operation(..`
     *  `create_payment_operation(..`
     * `account_merge_operation(..` 
       * is accessible for testing purpose only, as this is a dangerous method to use.


clients/wallet/src/stellar_wallet.rs
```diff
@@ -223,6 +179,19 @@ impl StellarWallet {
 		Ok(TransactionsResponseIter { records, next_page, client: horizon_client })
 	}
 
+	///return balances for public key
```
ebma, line 182: The comment is a bit misleading. 
I'd rephrase to: "Returns the balances of this wallet's Stellar account"

ebma: We should also consider renaming the function to `get_balances` as the result is a vector

clients/wallet/src/stellar_wallet.rs
```diff
@@ -305,18 +276,59 @@ impl StellarWallet {
...
 
+	pub async fn send_payment_to_address_with_extra_operations(
+		&mut self,
+		destination_address: PublicKey,
+		asset: Asset,
+		stroop_amount: StellarStroops,
+		request_id: [u8; 32],
+		stroop_fee_per_operation: u32,
+		extra_operations: Vec<Operation>,
+	) -> Result<TransactionResponse, Error> {
```
ebma, line 287: Actually, the three operations we need, ie. 'payment' vs 'createaccount' vs 'createclaimablebalance' should be used mutually exclusive. Each of these operations is able to hold a specified amount (which will be our redeemed amount). Thus, we don't need to use a combination of them. 

This means, that we actually wouldn't need this function since we don't want to bundle these operations together in one transaction.

clients/wallet/src/operations.rs
```diff
@@ -0,0 +1,328 @@
...
+		destination_address: PublicKey,
+		to_be_redeemed_asset: Asset,
+		to_be_redeemed_amount: StellarStroops,
+	) -> Option<Operation> {
+		// create an account for the destination ONLY if redeeming >= 1 XLM.
+		let to_be_redeemed_amount_u128 = stellar_stroops_to_u128(to_be_redeemed_amount);
+
+		if to_be_redeemed_asset == Asset::AssetTypeNative &&
+			to_be_redeemed_amount_u128 >= primitives::CurrencyId::StellarNative.one()
+		{
```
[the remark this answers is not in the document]
ebma: Also, if we move the logic here, we should probably still rename the function to `create_account_ops_if_required`

ebma (changes requested):
I left some comments. 

It seems like there was some misconception about the needed combination of operations, sorry if the ticket #357 did not make that clear enough @b-yap. I think we can simplify the implemented logic if we take into account that the operations are exclusive. 

Just to be clear, the [createaccount](https://developers.stellar.org/docs/fundamentals-and-concepts/list-of-operations#create-account) operation is much like a payment operation. You can consider it as a 'payment' that always uses the native asset, with 'starting balance' being the amount you transfer/pay to the destination account. There is no extra payment necessary if we do this.

clients/wallet/src/stellar_wallet.rs
b-yap: method removed :bow:

clients/wallet/src/operations.rs
```diff
@@ -0,0 +1,328 @@
...
+
+	#[async_trait]
+	pub trait RedeemOps: HorizonClient {
+		async fn claimable_balance_ops_required(
+			&self,
+			destination_address: PublicKey,
+			is_public_network: bool,
+			to_be_redeemed_asset: Asset,
+			to_be_redeemed_amount: StellarStroops,
+		) -> Result<Operation, Vec<Operation>> {
```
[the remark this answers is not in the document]
b-yap: noted, and changed. see `create_payment_op_for_redeem_request(..)`

clients/wallet/src/stellar_wallet.rs
```diff
@@ -223,6 +179,19 @@ impl StellarWallet {
 		Ok(TransactionsResponseIter { records, next_page, client: horizon_client })
 	}
 
+	///return balances for public key
+	pub async fn get_balance(&self) -> Result<Vec<Balance>, Error> {
+		let horizon_client = reqwest::Client::new();
+		let public_key_encoded = self.get_public_key().to_encoding();
+		let account_id_string =
+			std::str::from_utf8(&public_key_encoded).map_err(Error::Utf8Error)?;
```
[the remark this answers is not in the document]
b-yap: addressed in primitives; see the `StellarTypeToString`

clients/wallet/src/stellar_wallet.rs
```diff
@@ -223,6 +179,19 @@ impl StellarWallet {
 		Ok(TransactionsResponseIter { records, next_page, client: horizon_client })
 	}
 
+	///return balances for public key
+	pub async fn get_balance(&self) -> Result<Vec<Balance>, Error> {
+		let horizon_client = reqwest::Client::new();
```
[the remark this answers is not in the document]
b-yap: added it in, as field `client`

clients/wallet/src/operations.rs
```diff
@@ -0,0 +1,332 @@
...
+	) -> Result<Operation, Error> {
+		change_error_type(Operation::new_account_merge(destination_address))?
+			.set_source_account(source_address)
+			.map_err(|e| {
+				Error::BuildTransactionError(format!("Setting source account failed: {e:?}"))
+			})
+	}
+
+	#[tokio::test]
+	async fn create_payment_op_for_redeem_request() {
```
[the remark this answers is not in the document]
ebma: The test cases itself look good overall.

clients/wallet/src/stellar_wallet.rs
```diff
@@ -41,6 +47,10 @@ pub struct StellarWallet {
 
 	/// the waiting time (in seconds) for retrying.
 	max_backoff_delay: u16,
+
+	/// a client to connect to Horizon
+	// todo: it's probably better that this is of generic type C, to remove dependency from reqwest
```
ebma, line 52: Can we do what the 'todo' says already? Or is there any blocker?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -305,21 +287,58 @@ impl StellarWallet {
...
 
+	pub async fn send_payment_to_address_for_redeem_request(
+		&mut self,
+		destination_address: PublicKey,
+		asset: Asset,
+		stroop_amount: StellarStroops,
+		request_id: [u8; 32],
+		stroop_fee_per_operation: u32,
+	) -> Result<TransactionResponse, Error> {
+		let op = self
```
ebma, line 298: Nit: Maybe rename `op` to `payment_like_operation` to indicate that this will be an operation that works similar to a payment? Or we add a comment above that states it will be any of the three types (payment, createaccount, createclaimablebalance)

clients/wallet/src/stellar_wallet.rs
```diff
@@ -387,10 +408,65 @@ mod test {
...
+			self.create_envelope(
+				request_id,
+				stroop_fee_per_operation,
+				next_sequence_number,
+				vec![payment_op],
+			)
+		}
+	}
+
 	fn wallet(storage: &str) -> Arc<RwLock<StellarWallet>> {
```
ebma, line 462: Nit: I'd rename to `wallet_with_secret_key_for_storage(storage: &str)` or similar. Seeing just `wallet("")` being called in the tests is confusing.

clients/wallet/src/stellar_wallet.rs
```diff
@@ -494,6 +582,152 @@ mod test {
...
+				.expect("should return a public key");
+
+		let asset = CurrencyId::try_from((
+			"USDC",
+			"GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC",
+		))
+		.expect("should convert ok");
+		let asset: Asset = asset.try_into().expect("should convert to Asset");
+
+		// let amount = 1;  // in the response, value is 0.0000001.
```
ebma, line 605: Nit: I guess we can delete this commment

clients/wallet/src/stellar_wallet.rs
```diff
@@ -494,6 +582,152 @@ mod test {
...
+		// let's cleanup, just to make sure.
+		wallet.cache.remove_all_tx_envelopes();
+
+		let destination =
+			PublicKey::from_encoding("GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN")
+				.expect("should return a public key");
+
+		let asset = CurrencyId::try_from((
+			"USDC",
+			"GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC",
```
ebma, line 600: Don't we have a constant for this exact USDC issuer already somewhere? Maybe we can re-use it somehow?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -494,6 +582,152 @@ mod test {
...
+
+						assert_eq!(claimant.destination, destination.to_encoding());
+					},
+					other => {
+						panic!("wrong result: {other:?}");
+					},
+				}
+			},
+			Err(e) => {
+				panic!("failed. {e:?}");
```
ebma, line 655: Nit: Maybe be more specific and say what failed

ebma (commented):
I left some nitpicky comments and suggestions. Overall, I think the changes are good and you did a great job @b-yap. Especially the tests are super valuable.

clients/wallet/src/stellar_wallet.rs
b-yap: :sob: it's a bit difficult, what with the wallet being cloned and shared in different tokio spawns; an example would be the code in `fn resubmit_transactions_from_cache()` in stellar_wallet.rs.

Thinking about it, I guess it doesn't matter right now, since the client is private anyway. And it is instantiated inside this crate, so external code won't even need to bother. 

I'll just remove the `todo`.

clients/wallet/src/stellar_wallet.rs
b-yap: is `payment_like_op` ok?

clients/wallet/src/operations.rs
```diff
@@ -0,0 +1,332 @@
...
+	source_address: PublicKey,
+) -> Result<Operation, Error> {
+	let stroop_amount = StroopAmount(stroop_amount);
+
+	change_error_type(Operation::new_payment(destination_address, asset, stroop_amount))?
+		.set_source_account(source_address)
+		.map_err(|e| Error::BuildTransactionError(format!("Setting source account failed: {e:?}")))
+}
+
+pub fn create_basic_transaction(
```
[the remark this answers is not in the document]
b-yap: renamed to `create_basic_stellar_transaction()`

clients/wallet/src/operations.rs
```diff
@@ -0,0 +1,332 @@
...
+			},
+			Err(e) => Err(e),
+		}
+	}
+}
+
+#[async_trait]
+impl RedeemOperationsExt for reqwest::Client {}
+
+fn change_error_type<T>(item: Result<T, StellarSdkError>) -> Result<T, Error> {
```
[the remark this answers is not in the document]
b-yap: changed to a more specific one: `map_err_as_build_tx_error()`

clients/wallet/src/operations.rs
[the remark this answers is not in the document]
b-yap: broke down to 6 test cases

clients/wallet/src/stellar_wallet.rs
b-yap: renamed

b-yap:
i'll rerun the CI after the [other CI job finishes.](https://github.com/pendulum-chain/spacewalk/actions/runs/5540603942?pr=366)

clients/wallet/src/stellar_wallet.rs
```diff
@@ -287,15 +287,28 @@ impl StellarWallet {
 		Ok(envelope)
 	}
 
+	/// Sends a 'Payment' Operation transaction specifically for redeem requests.
```
ebma, line 290: I'd rephrase to the following because otherwise it's technically wrong
```suggestion
	/// Sends a 'Payment' transaction specifically for redeem requests.
```

ebma (approved):
Thanks for the improvements, looks good 👍 I just left few minor comments.

clients/wallet/src/operations.rs
```diff
@@ -66,43 +79,48 @@ pub trait RedeemOperationsExt: HorizonClient {
...
 	create_claimable_balance_operation(to_be_redeemed_asset, to_be_redeemed_amount, vec![claimant])
 }
 
+/// Converts external error into wallet's `BuildTransactionError`.
+trait BuildTransactionErrorExt<T> {
+	fn map_err_as_build_tx_error(self) -> Result<T, Error>;
+}
+
+impl<T> BuildTransactionErrorExt<T> for Result<T, StellarSdkError> {
+	fn map_err_as_build_tx_error(self) -> Result<T, Error> {
```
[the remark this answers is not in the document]
b-yap: added new one: `fn map_err_as_build_tx_error_with_text(text:&str)`

ebma:
Nice, looks good and ready for merge 👍 By the way, thanks for creating such a detailed PR description again @b-yap 👌

PR #366, opened 2023-06-26
Make vault client submit proof using 'confirm' messages and add fallback to other archive nodes on error

### LINKED ISSUE #361: Make vault client submit proof using 'confirm' messages and add fallback to other archive nodes on error
Once we changed the stellar-relay pallet to also consider 'Confim' SCP messages as part of a proof in #360, we need to add some changes to the vault client so that he is able to build such a proof using the entries from a Stellar archive. 

We should:
- [ ] Change the logic about building a proof from an archived entry such that the vault client uses all 'externalized' and 'confirm' statements that are available in that archive entry. Note that a valid proof always has to contain at least 1 externalized message, if that's not the case we should fall back to a different archive node. 
- [ ] Add some error handling that makes the vault client fall back to a different Stellar archive node in case the `execute...` extrinsic fails with a selection of error messages. Basically, all errors thrown particularly by the `stellar-relay` pallet can be considered worth a retry because we should always be able to build a proper proof using an archive entry. An error thrown by the `stellar-relay` would indicate that the entry is faulty and we should try again with the entry of another archive node.

ebma:
- [x] Change logic about proof building from archive in accordance to #361.
- [x] Adds fallback mechanism for archive fetching. Instead of storing just one archive URL, the overlay config files now contain an array.
- [x] Changes the URL creation to not assume a trailing '/' at the end of the URL. 

Closes #361.


clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -182,43 +182,90 @@ impl ScpMessageCollector {
...
+						let relevant_envelopes = vec_scp
+							.into_iter()
+							.filter(|scp| match scp.statement.pledges {
+								ScpStatementPledges::ScpStExternalize(_) |
+								ScpStatementPledges::ScpStConfirm(_) => true,
+								_ => false,
+							})
+							.collect::<Vec<_>>();
+
+						let externalized_envelopes = relevant_envelopes
```
b-yap, line 236: Just a suggestion:
`externalized_envelopes` is not actually used; might be better if we use `filter(...).count()` instead? That way we're not collecting duplicate envelopes

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -182,43 +182,90 @@ impl ScpMessageCollector {
...
+							.iter()
+							.filter(|scp| match scp.statement.pledges {
+								ScpStatementPledges::ScpStExternalize(_) => true,
+								_ => false,
+							})
+							.collect::<Vec<_>>();
+
+						// Ensure that at least one envelope is externalized
+						if externalized_envelopes.len() == 0 {
```
b-yap, line 245: so in this part, it'll be `if count == 0`.

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -231,29 +278,33 @@ impl ScpMessageCollector {
...
-			let value = transactions_archive
-				.get_vec()
-				.iter()
-				.find(|&entry| Slot::from(entry.ledger_seq) == slot);
-
-			if let Some(target_history_entry) = value {
-				tracing::debug!("Adding archived tx set for slot {}", slot);
-				let mut tx_set_map = txset_map_arc.write();
-				tx_set_map.insert(slot, target_history_entry.tx_set.clone());
+					return
```
b-yap, line 292: Correct me if I'm wrong... 
let's say we have 10 archive urls. 

If at **_first_**  or **_fifth_** iteration it errors out, is this going to stop the loop and return immediately? Is there no need to check until the last url?
^ if that is the expected behaviour, please disregard this comment, :sweat:

ebma: Ahh nice catch. It should actually be a `continue` and not a `return` since we of course want to continue with the other archives in case we receive an error

clients/vault/src/oracle/collector/proof_builder.rs
ebma: Good point 👍

ebma:
@b-yap I addressed your feedback and added some tests specifically for the fallback. I noticed some issues while doing so and also addressed them now.

clients/stellar-relay-lib/src/config.rs
```diff
@@ -8,9 +8,9 @@ use substrate_stellar_sdk::SecretKey;
 /// It configures both the ConnectionInfo and the NodeInfo.
 #[derive(Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct StellarOverlayConfig {
-	stellar_history_base_url: String,
-	connection_info: ConnectionInfoCfg,
-	node_info: NodeInfoCfg,
+	pub stellar_history_archive_urls: Vec<String>,
```
b-yap, line 11: possibly `pub(crate)` ? I assumed this was made pubilc for testing reasons.
Not sure if we want the fields exposed outside of `stellar-relay-lib`

b-yap:
once the test is fixed, I  :+1:

clients/stellar-relay-lib/src/config.rs
ebma: We actually can't use `pub(crate)` here because this config lives in the `stellar-relay-lib` and the tests I added live in the `vault/oracle` crate 😅

b-yap: oh, my bad. alrighty then, 👍

ebma:
@b-yap CI passed, can you approve?

b-yap (approved):

PR #377, opened 2023-07-19
added error logs and debug

### LINKED ISSUE #356: Improve info and usage of log messages
We should make some effort to improve the info in some of our log messages.  Also, we should remove unnecessary log messages that clutter the output of our vault clients.

I'd recommend we:
- Remove the log messages about removing an old entry in `vault::oracle::types`
- Change the log level or reword the log messages in `stellar_relay_lib::connection::connector::message_handler`
- Add log messages that describe the current status of issue/redeem request execution. E.g. I noticed that the client does not log a statement once it tries to execute (and successfully executed) a new issue request

Think of it from the perspective of an external vault operator. When running the vault client you want to be able to choose between log levels that suit your needs. For example, you'd expect that:
- the 'info' level shows you log messages about the general operation of the vault client, ie. it informs you about current execution steps (receiving/finding a new issue/redeem event on-chain, executing that issue, etc). 
- the 'debug' level contains information that is useful for debugging the problem, but too detailed for general use
- the 'warn' level really only contains warnings, ie. messages that require actions from the operator or similar.

b-yap:
closes #356 

added new logs, removed unnecessary ones, and downgrade/upgrade levels of existing logs


ebma (approved):
Don't mind all these comments, I started suggesting improvements but then realized that it's easier to just add them myself immediately, less overhead for everyone. I actually tried to cancel/delete those but apparently GitHub didn't want me to haha.

Anyways, thanks for your improvements @b-yap, I just tried to make the log messages of info or higher level more human-friendly. I think it's good now

ebma:
@b-yap I noticed I introduced some compilation errors 😅 Can you take care of those or should I?

PR #387, opened 2023-08-23
234 refactor dia implementation

### LINKED ISSUE #234: Refactor DIA implementation
There are some things that are not really clean about the code introduced when adding the DIA oracle yet. 

We should:
- [x] Deduplicate all the mock structures that were introduced and added to other pallets. 
Due to some issue with the `thread_local` implementation we had to copy the same mock code to multiple pallets. This is not ideal of course and we should try to get rid of it. Maybe an implementation similar to [this](https://github.com/pendulum-chain/spacewalk/blob/521b97187a271879d8e921b4d7d4da89493c8fbf/testchain/runtime/src/benchmark_utils.rs) one can help and make us able to declare all this code in just one place, sharing it throughout the repository.
- [x] ~~Improve and deduplicate the DiaOracleKeyConvertor. Right now, the DiaOracleKeyConvertor does not take the issuer of stellar assets into account. We should think about changing that. We should find a way to translate between Stellar assets and DIA identifiers bi-directionally meaning that a Stellar asset (issuer + code) is uniquely defined by a DIA asset and vice versa. For this to work, we might also have to touch the dia-batching-server implementation because this responds with the prices for DIA assets.~~ Will be addressed in #396 

Follow-up ticket to the changes made in #187.

ebma:
This PR creates a re-usable definition for the DIA oracle mock, which is moving the implementation from `pallets/oracle/src/mock.rs` to a new module `pallets/oracle/src/testing_utils` and adding some other trait definitions that make it reusable.
The duplicated definitions are then removed from the other pallets and testchain.

## Notes
- The implementation contained in the `oracle::testing_utils` module is now 'no_std'. This is important to prevent compilation issues when re-using this module in the other crates.
- The `spin` crate is used for no_std Mutex support
- The `once_cell` crate is used for thread-safe and synchronized static variables. Using `lazy_static` is not sufficient because it does not prevent the multi-threaded tests initialize the static variable at the same time, leading to race conditions. We have to use a struct of `once_cell`s 'race' module because only these are no_std, see the FAQ [here](https://docs.rs/once_cell/latest/once_cell/#faq).
- The static `LOCK` contained in the `oracle::testing_utils` is supposed to be used in tests that need to access the oracle mock data. It effectively makes all tests (that use this lock) execute sequentially instead of in parallel, similar to the `serial_test::serial` macro. But I did not want us to put the `#[serial]` macro to every test because it's ugly and a little more error-prone IMO.

Closes #234.


ebma:
Turning this into a draft again because apparently there are some more severe conflicts I need to solve.
Also sorry for actually putting you guys as 'assignees' and not as 'reviewers' haha @b-yap @adelarja.

ebma:
I now figured out why the vault registry tests were suddenly failing. The reason is that I replaced the [thread_local()](https://github.com/pendulum-chain/spacewalk/pull/387/files#diff-9ad170fb43c3d98addd1ed73b3faa45611cc114228d564c2ed5d4760af4cb15aL257-L259) implementation with a `lazy_static()` initialization. I didn't realize though that this introduces a problem because this `lazy_static` reference of the `COINS` is now shared across all tests, which then introduces race conditions because the exchange rates [are cleared ](https://github.com/pendulum-chain/spacewalk/pull/387/files#diff-d6cdd749af964b12aacc03febe5032c67c92157c7f9864bdaec757186d686165R378) before each test run. Since the tests are running in parallel by default, I had to introduce the `#[serial]` macro to make sure that the clearing of before one test does not clear the values that were prepared for the execution of another test, etc. 

I find this `#[serial]` annotation but I don't know of a better way to fix this right now. I had to make the code in the `oracle_mock` free of any `std` references so that it can be shared among all pallets without introducing compilation errors. Thus `thread_local` is not available. 

Does anyone have some ideas how to approach this without the `#[serial]` macro, @b-yap @adelarja @gianfra-t ?

gianfra-t:
> I now figured out why the vault registry tests were suddenly failing. The reason is that I replaced the [thread_local()](https://github.com/pendulum-chain/spacewalk/pull/387/files#diff-9ad170fb43c3d98addd1ed73b3faa45611cc114228d564c2ed5d4760af4cb15aL257-L259) implementation with a `lazy_static()` initialization. I didn't realize though that this introduces a problem because this `lazy_static` reference of the `COINS` is now shared across all tests, which then introduces race conditions because the exchange rates [are cleared ](https://github.com/pendulum-chain/spacewalk/pull/387/files#diff-d6cdd749af964b12aacc03febe5032c67c92157c7f9864bdaec757186d686165R378) before each test run. Since the tests are running in parallel by default, I had to introduce the `#[serial]` macro to make sure that the clearing of before one test does not clear the values that were prepared for the execution of another test, etc.
> 
> I find this `#[serial]` annotation but I don't know of a better way to fix this right now. I had to make the code in the `oracle_mock` free of any `std` references so that it can be shared among all pallets without introducing compilation errors. Thus `thread_local` is not available.
> 
> Does anyone have some ideas how to approach this without the `#[serial]` macro, @b-yap @adelarja @gianfra-t ?

If I understand correctly, the problem is that one test is clearing, then setting `COINS` to a test value and then in the middle comes another test and clears `COINS`.  If that is the case, I can think of some naive lock where you wrap `COINS` around a struct `struct NewCoins {actual_coins: COINS, using: bool, cvar: Condvar}` and then have the mock wait until it gets notified via cvar that using is false. Something like here: https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html

ebma:
Thanks for your comment @gianfra-t. I decided not to use `condvar` but your suggestion led me to implement a locking mechanism that works similarly to the `#[serial]` macro.

pallets/oracle/src/lib.rs
```diff
@@ -237,19 +242,33 @@ impl<T: Config> Pallet<T> {
...
+				oracle_keys.push(key);
 			}
 		}
 		<OracleKeys<T>>::put(oracle_keys.clone());
 		Ok(())
 	}
 
+	// public only for testing purposes
+	#[cfg(feature = "testing-utils")]
+	pub fn _clear_values() -> DispatchResult {
```
b-yap, line 260: Do we have a rule about when to use underscore prefix function names? 
Currently I had this impression of using such function/method _only_ internally... and I mean only on the _same lib/mod_. However, the `_clear_values()` is liberally called in other pallets like `issue` and `vault-registry`.

examples:
* the `fn _offchain_worker()` in `pallet/vault-registry` 
* the `fn _withdraw_collateral()` in `pallet/nomination`
*  the `fn _request_issue()` in `pallet/issue`

b-yap (approved):
aside from the conflicts, I have no complaints.

pallets/oracle/src/lib.rs
ebma: Good point. We did not discuss this yet and I just used the underscore prefix because it is used by the other 'testing-utils' functions. But you are right that it's used in other pallets as well. 

Soooo... @pendulum-chain/devs what's your opinion on `_function()` names? I would use them for functions that are supposed to be used 'module-only' or helper functions. Thus, I would need to change the naming to `clear_values()` in this case.

ebma: I now renamed all three functions because all of them are used in other pallets as well.

ebma:
I ran the tests and clippy locally and it worked fine so I'll merge this before waiting for CI to finish.

PR #388, opened 2023-08-28
257 FIx Informational Issues - Logic Should be Moved To an Separate Function - Refactoring

b-yap:
continuation of https://github.com/pendulum-chain/spacewalk/pull/385. Please review the base PR before continuing with this.

* ## 9B2-04 | Logic Should be Moved To an Separate Function - Refactoring (partial)
  * ## **from audit:**
    * Refactoring is recommended in the function `validate_stellar_transaction()` from stellar_relay/lib.rs.
    * `run_service()` from vault/src/system.rs
  * ## new file
    * validation.rs (covers the logic of `validate_stellar_transaction()`

### DIFF
--- a/pallets/stellar-relay/src/validation.rs
+++ b/pallets/stellar-relay/src/validation.rs
@@ -37,8 +37,7 @@ fn validator_count_per_org<T: Config>(
 fn targeted_organization_map<T: Config>(
 	envelopes: &UnlimitedVarArray<ScpEnvelope>,
 	validators: &BoundedVec<ValidatorOf<T>, T::ValidatorLimit>,
-	orgs_length: usize,
-) -> Result<BTreeMap<T::OrganizationId, u32>, Error<T>> {
+) -> BTreeMap<T::OrganizationId, u32> {
 	// Find the validators that are targeted by the SCP messages
 	let targeted_validators = validators
 		.iter()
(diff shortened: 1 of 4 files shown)


pallets/stellar-relay/src/validation.rs
```diff
@@ -0,0 +1,224 @@
...
+	let node_id_found = validators
+		.iter()
+		.any(|validator| validator.public_key.to_vec() == node_id.to_encoding());
+
+	ensure!(node_id_found, Error::<T>::EnvelopeSignedByUnknownValidator);
+
+	Ok(node_id)
+}
+
+pub fn check_targets<T: Config>(
```
ebma, line 105: Nit: Maybe we can give this a better name. Maybe `check_quorum_validity` or `check_for_valid_quorum_set`

pallets/stellar-relay/src/validation.rs
```diff
@@ -0,0 +1,224 @@
...
+	let targeted_organization_count = targeted_organization_map.len();
+
+	// Check that the distinct organizations occurring in the validator structs related to
+	// the externalized messages are more than 2/3 of the total amount of organizations in
+	// the tier 1 validator set.
+	// Use multiplication to avoid floating point numbers.
+	ensure!(
+		targeted_organization_count * 3 > orgs_length * 2,
+		Error::<T>::InvalidQuorumSetNotEnoughOrganizations
+	);
```
ebma, line 74: I would move this part also to the `check_targets()` function. It's not wrong here but it's better fitting to the other one, especially if we rename it to `check_for_valid_quorum_set` as this part of related to that.

ebma (approved):
These are really great changes, thanks for the great refactoring 🎉  I left two nitpicky comments but besides that, IMO it's ready to be merged into the other branch.

pallets/stellar-relay/src/validation.rs
b-yap: updated

pallets/stellar-relay/src/validation.rs
b-yap: updated

ebma (approved):

PR #391, opened 2023-09-01
[WIP] 340 add handling for non recoverable stellar submission errors

### LINKED ISSUE #340: Add handling for non-recoverable Stellar submission errors
This is related to #315. 

We need to define what we want to do in case we encounter a non-recoverable Stellar submission error. 
There are some options to approach this:

1. automatically try to recreate and resubmit the transaction without interrupting the flow. 
   - eg when we encounter a `tx_bad_seq` error, we could create a new transaction with the same details (operations, destination, etc) but use a new sequence number and automatically resubmit that. This would only work in some cases though, since if for example the transaction is rejected because the accounts balance is too low, automated handling of this is impossible (as it requires user actions). Also, this might lead to unwanted side effects because recreating a new creation automatically could lead to double payments if done wrong.
1. do not automatically try to create a 'valid' transaction but pass the non-recoverable error to the calling code and log an error message. This approach does not introduce any unwanted side effects but also does not fully solve the situation. For some operations like the account balance of the vault operators account being too low, an action by the vault operator would be required.

Both approaches have their disadvantages. Option 2) with not automatically treating 'non-recoverable' errors is preferrable to prevent double payments, but what do we want to do instead? How would we tell the vault operator what's wrong?

Maybe we should list all non-recoverable errors that can happen and that decide individually how we want to approach each of them.

adelarja:
The aim of this PR is to add error handling logic for failed submissions.

Closes #340

1. A few method renames;
1. Introduce new file: `resubmissions.rs`, dedicated for resubmission and handling of errors. Bulk of the code is here
2. Introduced new file: `mock.rs`, contains the function helpers for the tests cases
3. ~Also updated the `resubmit_transactions_from_cache()`, with a parameter that helps handle failed tasks~

## How to begin the review:
* clients/wallet/src/resubmissions.rs
  * ~type `FailedTasks` = essential for removing transaction envelopes in cache for failed resubmissions~
    * :star: instead of complicating things with channels, I just immediately call the `remove_tx_envelope_from_cache()` if the error is not viable for resubmission
  * fn `resubmit_transactions_from_cache()` = the only public method. Every 30 minutes it will read from cache and resubmit existing ones
  * fn `_resubmit_transactions_from_cache()`
     1.  Collect all envelopes from cache.
     2. Loops through all envelopes and resubmit them
     3.  ^ collects errors from the loop and tries to handle them
     4. ~if auto check is set to true, check `fn auto_check_running_tasks()`~
  * fn `handle_errors()`
    * loops through all the errors and handle them one by one
    * :star: for errors that cannot be resubmitted again, its corresponding envelope is removed from cache
  * fn `handle_error()` = resubmits transactions with the ff errors: 
    * `tx_bad_seq`
    * `tx_internal_error`
    * `CacheErrorKind::SequenceNumberAlreadyUsed`
 * fn `handle_tx_internal_error()` = resubmits a transaction with the error `tx_internal_error`
 * fn `handle_tx_bad_seq_error_with_xdr()` = resubmits a transaction with the error `tx_bad_seq`
 * fn `bump_sequence_number_and_submit()` = bump the sequence number of a transaction and resubmits it
 * fn `is_transaction_already_submitted()` = a costly checking if a transaction already exists. This is currently limited to just 10 pages
 * ~fn `auto_check_running_tasks()` = auto checks for failed resubmissions and forcefully remove transactions from the cache~

### DIFF
--- a/clients/wallet/src/resubmissions.rs
+++ b/clients/wallet/src/resubmissions.rs
@@ -23,7 +23,10 @@ const MAX_LOOK_BACK_PAGES: u8 = 10;
 
 #[cfg_attr(test, mockable)]
 impl StellarWallet {
-	pub async fn start_periodic_resubmission_of_transactions_from_cache(&self, interval_in_seconds: u64) {
+	pub async fn start_periodic_resubmission_of_transactions_from_cache(
+		&self,
+		interval_in_seconds: u64,
+	) {
 		// Perform the resubmission
 		self._resubmit_transactions_from_cache().await;
 
@@ -110,7 +113,6 @@ impl StellarWallet {
 		#[allow(unused_mut)] mut errors: Vec<(Error, TransactionEnvelope)>,
 	) {
 		while let Some((error, env)) = errors.pop() {
-
 			// handle the error
 			match self.handle_error(error).await {
 				// a new kind of error occurred. Process it on the next loop.
@@ -121,8 +123,8 @@ impl StellarWallet {
 					errors.push((e, env));
 				},
 
-				// Resubmission failed for this Transaction Envelope and it's a non-recoverable error
-				// Remove from cache
+				// Resubmission failed for this Transaction Envelope and it's a non-recoverable
+				// error Remove from cache
 				Ok(None) => self.remove_tx_envelope_from_cache(&env),
 
 				// Resubmission was successful
@@ -706,7 +708,8 @@ mod test {
 		// let's resubmit these 3 transactions
 		let _ = wallet.start_periodic_resubmission_of_transactions_from_cache(60).await;
 
-		// We wait until the whole cache is empty because eventually all transactions should be handled
+		// We wait until the whole cache is empty because eventually all transactions should be
+		// handled
 		pause_process_in_secs(10).await;
 
 		loop {
--- a/clients/wallet/src/resubmissions.rs
+++ b/clients/wallet/src/resubmissions.rs
@@ -35,7 +35,6 @@ impl StellarWallet {
 			let me_clone = Arc::clone(&me);
 			loop {
 				// Loops every 30 minutes or 1800 seconds
-				#[cfg(not(test))]
 				pause_process_in_secs(interval_in_seconds).await;
 
 				me_clone._resubmit_transactions_from_cache().await;
@@ -706,7 +705,7 @@ mod test {
 			.mock_safe(move |_, _| MockResult::Return(Box::pin(async move { false })));
 
 		// let's resubmit these 3 transactions
-		let _ = wallet.start_periodic_resubmission_of_transactions_from_cache(RESUBMISSION_INTERVAL_IN_SECS).await;
+		let _ = wallet.start_periodic_resubmission_of_transactions_from_cache(60).await;
 
 		// We wait until the whole cache is empty because eventually all transactions should be handled
 		pause_process_in_secs(10).await;
--- a/clients/wallet/src/resubmissions.rs
+++ b/clients/wallet/src/resubmissions.rs
@@ -358,7 +358,7 @@ async fn auto_check_running_tasks(running_tasks: FailedTasks, wallet: Arc<Stella
 	tokio::spawn(async move {
 		loop {
 			#[cfg(not(test))]
-			pause_process_in_secs(30).await;
+			pause_process_in_secs(1800).await;
 
 			#[cfg(test)]
 			pause_process_in_secs(5).await;
--- a/clients/wallet/src/resubmissions.rs
+++ b/clients/wallet/src/resubmissions.rs
@@ -70,9 +70,11 @@ impl StellarWallet {
 		//  Log those with errors.
 		let envelopes = match self.get_tx_envelopes_from_cache() {
 			Ok((envs, errors)) => {
-				tracing::warn!(
-					"_resubmit_transactions_from_cache(): errors from cache: {errors:?}"
-				);
+				if !errors.is_empty(){
+					tracing::warn!(
+						"_resubmit_transactions_from_cache(): errors from cache: {errors:?}"
+					);
+				}
 				envs
 			},
 			Err(errors) => {
@@ -92,7 +94,7 @@ impl StellarWallet {
 		if envelopes.is_empty() {
 			return
 		}
-		tracing::info!("resubmitting {:?} envelopes in cache...", envelopes.len());
+		tracing::info!("_resubmit_transactions_from_cache(): resubmitting {:?} envelopes in cache...", envelopes.len());
 
 		let mut error_collector = vec![];
 		// loop through the envelopes and resubmit each one
--- a/clients/wallet/src/resubmissions.rs
+++ b/clients/wallet/src/resubmissions.rs
@@ -5,7 +5,7 @@ use crate::{
 		CacheError, CacheErrorKind, Error,
 		Error::{DecodeError, ResubmissionError},
 	},
-	StellarWallet,
+	StellarWallet, TransactionResponse,
 };
 use primitives::{
 	stellar::{Memo, Transaction, TransactionEnvelope, XdrCodec},
@@ -18,16 +18,22 @@ use tokio::time::sleep;
 use mocktopus::macros::mockable;
 use tokio::sync::RwLock;
 
+pub type FailedTasks =
+	Arc<RwLock<Vec<tokio::sync::oneshot::Receiver<Option<TransactionEnvelope>>>>>;
+
 #[cfg_attr(test, mockable)]
 impl StellarWallet {
-	pub async fn resubmit_transactions_from_cache(
-		&self,
-	) -> Arc<RwLock<Vec<tokio::sync::oneshot::Receiver<bool>>>> {
+	/// Returns a list of receivers that will receive a `TransactionEnvelope`
+	/// if resubmission failed
+	///
+	/// # Arguments
+	/// * `auto_check` - creates another thread that does the checking for failed resubmissions
+	pub async fn resubmit_transactions_from_cache(&self, auto_check: bool) -> FailedTasks {
 		// a collection of all tasks currently running
 		let running_tasks = Arc::new(RwLock::new(vec![]));
 
 		// perform the resubmission
-		self._resubmit_transactions_from_cache(running_tasks.clone()).await;
+		self._resubmit_transactions_from_cache(running_tasks.clone(), auto_check).await;
 
 		// // clone the task, to share this in another thread
 		// let running_tasks_clone = running_tasks.clone();
@@ -38,10 +44,8 @@ impl StellarWallet {
 		// tokio::spawn(async move {
 		//     let me_clone = Arc::clone(&me);
 		//     loop {
-		//         println!("time for spawning");
 		//         pause_process_in_secs(300).await;
 		//
-		//         println!("let's gooo");
 		//         me_clone._resubmit_transactions_from_cache(running_tasks_clone.clone()).await;
 		//     }
 		// });
@@ -50,10 +54,17 @@ impl StellarWallet {
 		running_tasks
 	}
 
+	#[doc(hidden)]
 	/// Submits transactions found in the wallet's cache to Stellar.
+	///
+	/// # Arguments
+	/// * `running_tasks` - a list of receivers that will receive a `TransactionEnvelope`
+	/// if resubmission failed
+	/// * `auto_check` - creates another thread that does the checking for failed resubmissions
 	async fn _resubmit_transactions_from_cache(
 		&self,
-		running_tasks: Arc<RwLock<Vec<tokio::sync::oneshot::Receiver<bool>>>>,
+		running_tasks: FailedTasks,
+		auto_check: bool,
 	) {
 		let _ = self.transaction_submission_lock.lock().await;
 
@@ -84,36 +95,49 @@ impl StellarWallet {
 		}
 
 		let mut error_collector = vec![];
+		// loop through the envelopes and resubmit each one
 		for envelope in envelopes {
-			if let Err(e) = submit(envelope).await {
+			if let Err(e) = submit(envelope.clone()).await {
 				tracing::debug!("_resubmit_transactions_from_cache(): encountered error: {e:?}");
-				error_collector.push(e);
+				// save the kind of error and the envelope that failed
+				error_collector.push((e, envelope));
 			}
 		}
 
-		// clone self, to use in another thread
-		let me = Arc::new(self.clone());
+		// a few errors happened and must be handled.
+		if !error_collector.is_empty() {
+			// clone self, to use in another thread
+			let me = Arc::new(self.clone());
+			let running_tasks_clone = running_tasks.clone();
+			tokio::spawn(async move {
+				me.handle_errors(error_collector, running_tasks_clone).await;
+			});
+		}
 
-		tokio::spawn(async move {
-			// handle errors found after submission
-			if !error_collector.is_empty() {
-				me.handle_errors(error_collector, running_tasks).await;
-			}
-		});
+		// auto removes a receiver from the list, if a running task has finished
+		let me = Arc::new(self.clone());
+		if auto_check {
+			auto_check_running_tasks(running_tasks.clone(), me).await;
+		}
 	}
 
+	#[doc(hidden)]
 	/// Handle all errors
+	///
+	/// # Arguments
+	/// * `errors` - a list of a tuple containing the error and its transaction envelope
+	/// * `running_tasks` - a list of receivers that will receive a `TransactionEnvelope`
 	async fn handle_errors(
 		&self,
-		mut errors: Vec<Error>,
-		running_tasks: Arc<RwLock<Vec<tokio::sync::oneshot::Receiver<bool>>>>,
+		mut errors: Vec<(Error, TransactionEnvelope)>,
+		running_tasks: FailedTasks,
 	) {
-		while let Some(error) = errors.pop() {
+		while let Some((error, env)) = errors.pop() {
 			// pause process for 20 minutes
 			#[cfg(not(test))]
 			pause_process_in_secs(1200).await;
 
-			// for testig purpose, set to 5 seconds
+			// for testing purpose, set to 5 seconds
 			#[cfg(test)]
 			pause_process_in_secs(5).await;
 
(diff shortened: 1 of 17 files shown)


clients/wallet/src/stellar_wallet.rs
```diff
@@ -374,12 +378,113 @@ impl StellarWallet {
...
+			},
+			_ => Err(error),
+		}
+	}
+
+	fn get_transaction_from_envelope(&self, tx_envelope: TransactionEnvelope) -> Option<Tx> {
+		match tx_envelope {
+			TransactionEnvelope::EnvelopeTypeTxV0(transaction) => Some(Tx::V0(transaction.tx)),
+			TransactionEnvelope::EnvelopeTypeTx(transaction) => Some(Tx::V1(transaction.tx)),
+			TransactionEnvelope::EnvelopeTypeTxFeeBump(fee_bump_transaction) =>
```
adelarja, line 429: Should we also take `EnvelopeTypeTxFeeBump` into account?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -374,12 +378,113 @@ impl StellarWallet {
...
+						"tx_internal_error" => self.submit_transaction(transaction_envelope).await,
+						_ => Err(error),
+					}
+				} else {
+					Err(error)
+				}
+			},
+
+			Error::CacheError(CacheError {
+				kind: CacheErrorKind::SequenceNumberAlreadyUsed,
```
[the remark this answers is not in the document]
ebma: Yes, I think so :+1:

clients/wallet/src/stellar_wallet.rs
ebma: Hmm I don't think so. I doubt that we'll need that (soon or ever)

clients/wallet/src/stellar_wallet.rs
```diff
@@ -374,12 +378,113 @@ impl StellarWallet {
...
+	}
+
+	fn get_transaction_envelope(&self, tx: Tx) -> TransactionEnvelope {
+		match tx {
+			Tx::V0(transaction) => self.transaction_v0_into_transaction_envelope(transaction),
+			Tx::V1(transaction) => transaction.into_transaction_envelope(),
+		}
+	}
+
+	pub fn transaction_v0_into_transaction_envelope(
```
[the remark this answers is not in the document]
ebma: Maybe instead of duplicating this function here, we could transform the `v0` tx into a `v1` transaction? It seems like it implements the `From` conversion [here](https://github.com/pendulum-chain/substrate-stellar-sdk/blob/d1dd579cc27c614abe9eab7d0cdeae12ea735c0b/src/xdr/impls/transaction.rs#L52) so maybe a simple `into()` would do the trick.

clients/wallet/src/stellar_wallet.rs
```diff
@@ -374,12 +378,112 @@ impl StellarWallet {
...
+impl StellarWallet {
+	pub async fn handle_error(&self, error: Error) -> Result<TransactionResponse, Error> {
+		match &error {
+			Error::HorizonSubmissionError { title: _, status: _, reason, envelope_xdr } => {
+				if let Some(env_xdr) = envelope_xdr {
+					let transaction_envelope =
+						TransactionEnvelope::from_base64_xdr(env_xdr).expect("should convert ok");
+					match &reason[..] {
+						"tx_bad_seq" =>
+							self.bump_sequence_number_and_submit(transaction_envelope).await,
```
ebma, line 400: Here we would still need to add the check if the sequence number is greater or less than the accounts sequence number etc.

ebma (commented):
Looks good so far. We'd still need to add some very descriptive logging of course but the core logic seems on track

clients/wallet/src/stellar_wallet.rs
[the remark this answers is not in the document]
adelarja: Good idea @ebma, I've refactored the code.

clients/wallet/src/stellar_wallet.rs
```diff
@@ -374,12 +378,92 @@ impl StellarWallet {
...
+			_ => None,
+		}
+	}
+
+	async fn is_valid_seq_number(&self, tx_envelope: &TransactionEnvelope) -> bool {
+		let account = self
+			.client
+			.get_account(self.get_public_key(), self.is_public_network)
+			.await
+			.expect("Expected account");
```
b-yap, line 440: hhhm, it could happen, i guess? Probably.  Maybe just log `tracing::error!` the error and return false?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -374,12 +378,92 @@ impl StellarWallet {
...
+		if let Some(sequence_number) = tx_envelope.sequence_number() {
+			return sequence_number > account.sequence
+		}
+		false
+	}
+
+	async fn update_transaction_sequence_number(&self, tx: Transaction) -> Transaction {
```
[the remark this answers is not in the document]
adelarja: You're right, @b-yap, I agree. I used this name because perhaps we could implement a different update logic, one that doesn't involve simply adding 1 to the sequence number. But then again, maybe we won't.

clients/wallet/src/stellar_wallet.rs
ebma: > Right now, I've implemented the logic we've discussed on the Notion page: If the transaction sequence number is lower than the account sequence number, we discard the transaction. If the transaction sequence number is greater than the account sequence number, we wait until the transaction sequence number matches the correct one.

True, sorry I somehow missed that the logic here is kind of hidden in the preceding `!is_valid_seq_number` check. So it should indeed already work as proposed.

> Is it possible to have a sequence number collision? For example, if the account sequence number is 10, could two transactions with sequence number 14 exist? In that case, only one of them would be resubmitted.

In theory, yes. In practice, also yes but we try to prevent it with our [transaction_submission_lock](https://github.com/pendulum-chain/spacewalk/blob/7851e15a8f7c409347bd7808e6d05543589720c1/clients/wallet/src/stellar_wallet.rs#L42).
Or maybe this question has some more facets to it so let's dive a little deeper. 
If we consider the whole network level, ie. transactions created by possibly two or more different clients/wallets that happen to run in parallel: Let's say our account sequence number is 10, and we 'simultaneously' submit two different Stellar transactions from that account with sequence number 11. As both are valid, it's up to the Stellar horizon server to pick one of them and reject the other.
If we consider just our single vault client: It should never happen that we create two transactions with the same sequence number because we have this `transaction_submission_lock`. _But_ if we accidentally create two different instances of a `StellarWallet` that share the same secret key, then these instances have each their own `transaction_submission_lock` and this guarantee does not hold anymore. So we have to be careful with these instances.

> Instead of discarding the transaction, would it be a better idea to increase the sequence number when the transaction sequence number is lower than the account sequence number? We would not have any double-spending problem here (in theory) since once the transaction is submitted, it is removed from the cache. Am I right?"

If we did it this way, we'd risk double-payments because: Let's say our account sequence number is at 9 and we create a payment transaction with sequence number 10. Now we submit this transaction to the Stellar network with the `submit_transaction()` function but before reaching [this](https://github.com/pendulum-chain/spacewalk/blob/ad68c16f55141357d14df66d40717b42d089c96c/clients/wallet/src/stellar_wallet.rs#L84) line, our client executable suddenly dies for whatever reason. This means that we submitted the transaction to Stellar but it did not get removed from cache. Now, once the client restarts it will check the cache and see that this transaction is still contained. But when submitting it, it will receive a `tx_bad_seq` error because our account already reached a sequence number of 10. If we now naively decide to just bump the sequence number of this transaction to resubmit it, we do the same payment twice.

So...... maybe the only really foolproof solution here would be _not_ to discard the transaction but to first check the account's transactions on the network and check if this transaction was already executed or not. We can check the account's transactions and look for payments with exactly the same amount, asset and memo (since the payments our vault client does will _always_ have a memo attached to them).

ebma: I'd suggest the following:
For transactions with an invalid (because *outdated*) sequence number, we query all Stellar transactions of that account and go through each one of them trying to find a `payment` operation that contains the same `TEXT_MEMO` as the invalid transaction we are dealing with. If we don't find any `payment` operation with the same memo, we bump the transaction's sequence number to an up-to-date one and re-submit.

WDYT @adelarja @b-yap?

b-yap: Sure.  Use [get_transactions_iter()](https://github.com/pendulum-chain/spacewalk/blob/ad68c16f55141357d14df66d40717b42d089c96c/clients/wallet/src/stellar_wallet.rs#L179) as inspo,  since it's a function that iterates BACKWARDS (from the newest to the oldest). There must be a filter though, until how far to look back.

adelarja: @ebma  @b-yap I like the idea. The responses memos has Option<Vec<u8>> type as we can see [here](https://github.com/pendulum-chain/spacewalk/blob/ad68c16f55141357d14df66d40717b42d089c96c/clients/wallet/src/horizon/responses.rs#L152). And the [Transaction](https://github.com/pendulum-chain/substrate-stellar-sdk/blob/master/src/xdr/types.rs#L3802) struct has Memo types. We can obtain the [MemoHash](https://github.com/pendulum-chain/substrate-stellar-sdk/blob/d1dd579cc27c614abe9eab7d0cdeae12ea735c0b/src/xdr/types.rs#L7217) from there, perform the necessary conversion, and compare, WDYT?

ebma: Oh by the way, now that you mention it: We are not using `MemoHash` anymore but `MemoText` because unfortunately no Stellar wallet supports the `MemoHash` type somehow. To make it more convenient for users, we switched to using the `MemoText` instead (but are still storing some hashed data inside it).

adelarja: Great @ebma many thanks!

clients/wallet/src/stellar_wallet.rs
```diff
@@ -374,12 +377,108 @@ impl StellarWallet {
...
 
-		let _ = self.cache.save_tx_envelope(envelope.clone())?;
-
 		self.submit_transaction(envelope).await
 	}
 }
 
+// Error handling methods for StellarWallet.
+impl StellarWallet {
+	pub async fn handle_error(&self, error: Error) -> Result<TransactionResponse, Error> {
```
adelarja, line 386: Maybe we could change the return type to bool here (or simply don't return anything?)

ebma: You mean because we don't actually use the `TransactionResponse` that is returned sometimes?
I think it's still nice to return an error here. Maybe we could change it to `Result<(), Error>` if we are sure that the `TransactionResponse` will never be needed.

clients/wallet/src/stellar_wallet.rs
```diff
@@ -198,76 +178,115 @@ impl StellarWallet {
...
 
-			tokio::spawn(async move {
-				if let Err(e) = sender.send(me_clone.submit_transaction(env).await) {
-					tracing::error!(
-						"Failed to send message during transaction resubmission: {e:?}"
-					);
-				};
-			});
-		}
+	pub(crate) fn sign_and_create_envelope(
```
ebma, line 260: Nit: `create_and_sign_envelope` would make more sense due to the order haha

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,795 @@
...
+		let failed_tasks_clone = failed_tasks.clone();
+
+		// clone self, to be able to use this in another thread
+		let me = Arc::new(self.clone());
+		// spawn a thread to resubmit envelopes from cache
+		tokio::spawn(async move {
+			let me_clone = Arc::clone(&me);
+			loop {
+				// loops every 30 minutes or 1800 seconds
+				pause_process_in_secs(1800).await;
```
ebma, line 48: Nit: We should probably use a proper constant here instead of `1800` as a magic number.

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,795 @@
...
+		failed_tasks: FailedTasks,
+	) {
+		while let Some((error, env)) = errors.pop() {
+			// pause process for 20 minutes
+			#[cfg(not(test))]
+			pause_process_in_secs(1200).await;
+
+			// for testing purpose, set to 5 seconds
+			#[cfg(test)]
+			pause_process_in_secs(5).await;
```
ebma, line 145: Why do we need to pause it? We are already pausing in the `loop` of the outer function, no?

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,795 @@
...
+
+	async fn handle_tx_bad_seq_error_with_envelope(
+		&self,
+		tx_envelope: TransactionEnvelope,
+	) -> Result<TransactionResponse, Error> {
+		let tx = tx_envelope.get_transaction().ok_or(DecodeError)?;
+
+		// Remove original transaction.
+		// The same envelope will be saved again using a different sequence number
+		self.remove_tx_envelope_from_cache(tx.seq_num);
```
ebma, line 260: I think it's better to keep remove this here and keep it in line 270. In the (very rare/unlikely) case that our vault client shuts down when executing the logic between line 260 and 264, the actual resubmission might not have been happened (because the client dies) but the sequence number and envelope got removed from cache, so upon restart the vault client would not try that one again, which is bad.

ebma: Oh actually, I'm wrong. I missed that we might return from this function in line 264 already. Let's move this line into the if statement of lines 263 to 265.

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,795 @@
...
+
+		// Check if we already submitted this transaction
+		if !self.is_transaction_already_submitted(&tx).await {
+			return self.bump_sequence_number_and_submit(tx).await
+		}
```
ebma, line 265:
```suggestion
		// Check if we already submitted this transaction
		if !self.is_transaction_already_submitted(&tx).await {
			self.bump_sequence_number_and_submit(tx).await
			self.remove_tx_envelope_from_cache(tx.seq_num);
			return;
		}
```

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,795 @@
...
+				break
+			}
+
+			for response in transaction.records {
+				// Make sure that we are the sender and not the receiver because otherwise an
+				// attacker could send a transaction to us with the target memo and we'd wrongly
+				// assume that we already submitted this transaction.
+				let Ok(source_account) = response.source_account() else {
+                    continue
+                };
```
ebma, line 315: Indentation seems off here

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,795 @@
...
+#[cfg_attr(test, mockable)]
+async fn pause_process_in_secs(in_secs: u64) {
+	sleep(Duration::from_secs(in_secs)).await;
+}
+
+fn decode_to_envelope(
+	envelope_xdr_as_str_opt: &Option<String>,
+) -> Result<TransactionEnvelope, Error> {
+	let Some( envelope_xdr) = envelope_xdr_as_str_opt else {
+        tracing::warn!("handle_error(): no envelope_xdr found");
```
ebma, line 350: Maybe we should also add the `envelope_xdr_as_str_opt` to the log here

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,795 @@
...
+				dummy_envelope.get_transaction().expect("must return a transaction");
+
+			let _ = wallet
+				.bump_sequence_number_and_submit(dummy_transaction.clone())
+				.await
+				.expect("return ok");
+
+			dummy_transaction.seq_num = seq + 1;
+
+			assert!(wallet.is_transaction_already_submitted(&dummy_transaction).await);
```
ebma, line 495: Nit: Since `is_transaction_already_submitted` does not check the sequence number of the transaction we wouldn't even need to increment it beforehand.

ebma (commented):
Added some minor comments but I really like the new structure overall and I think it should do the trick. Good job @b-yap :+1:

clients/wallet/src/resubmissions.rs
b-yap: it's inside the `else{}` block, so `envelope_xdr_as_str_opt` is None, even if we log it

clients/wallet/src/resubmissions.rs
b-yap: removed

clients/wallet/src/resubmissions.rs
b-yap: i changed the code

clients/wallet/src/resubmissions.rs
b-yap: i called `remove_tx_envelope_from_cache()` first, before the `bump_sequence_number_and_submit()`

clients/wallet/src/resubmissions.rs
b-yap: moved

clients/wallet/src/resubmissions.rs
b-yap: removed

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,729 @@
...
+use mocktopus::macros::mockable;
+
+#[cfg(not(test))]
+const RESUBMISSION_INTERVAL_IN_SECS: u64 = 1800;
+
+const MAX_LOOK_BACK_PAGES: u8 = 10;
+
+#[cfg_attr(test, mockable)]
+impl StellarWallet {
+	pub async fn resubmit_transactions_from_cache(&self) {
```
ebma, line 27: Should we rename this function to indicate that this will actually start scheduling a periodic resubmission? Maybe we should also pass in the resubmission interval into this function call instead. Something like: `start_periodic_resubmission_of_transactions_from_cache(&self, interval_in_seconds: u64)`?

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,729 @@
...
+	/// # Arguments
+	/// * `errors` - a list of a tuple containing the error and its transaction envelope
+	async fn handle_errors(
+		&self,
+		#[allow(unused_mut)] mut errors: Vec<(Error, TransactionEnvelope)>,
+	) {
+		while let Some((error, env)) = errors.pop() {
+			// pause process for 30 minutes
+			#[cfg(not(test))]
+			pause_process_in_secs(RESUBMISSION_INTERVAL_IN_SECS).await;
```
ebma, line 117: Why are we pausing?

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,729 @@
...
+			match self.handle_error(error).await {
+				// a new kind of error occurred. Process it on the next loop.
+				Err(e) => {
+					tracing::error!("handle_errors(): new error occurred: {e:?}");
+
+					// push the transaction that failed, and the corresponding error
+					errors.push((e, env));
+				},
+
+				// Resubmission failed for this Transaction Envelope
```
ebma, line 129:
```suggestion
				// Resubmission failed for this Transaction Envelope and it's a non-recoverable error
```

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,729 @@
...
+				tracing::warn!("handle_error(): SequenceNumberAlreadyUsed error but no envelope");
+			},
+			_ => tracing::warn!("handle_error(): Unrecoverable error in Stellar wallet: {error:?}"),
+		}
+
+		// the error found is not recoverable, and cannot be resubmitted again.
+		Ok(None)
+	}
+
+	async fn handle_tx_internal_error(
```
ebma, line 187:
```suggestion

	// We encountered an unknown error and try submitting the transaction again as is
	async fn handle_tx_internal_error(
```

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,729 @@
...
+			let error = Error::HorizonSubmissionError {
+				title: "Transaction Failed".to_string(),
+				status: 400,
+				reason: "tx_bad_auth".to_string(),
+				result_code_op: vec![],
+				envelope_xdr,
+			};
+
+			match wallet.handle_error(error).await {
+				Ok(None) => assert!(true),
```
ebma, line 633:
```suggestion
				// `tx_bad_auth` is not recoverable so we expect `Ok(None)`
				Ok(None) => assert!(true),
```

clients/wallet/src/resubmissions.rs
```diff
@@ -0,0 +1,729 @@
...
+		StellarWallet::is_transaction_already_submitted
+			.mock_safe(move |_, _| MockResult::Return(Box::pin(async move { false })));
+
+		// let's resubmit these 3 transactions
+		let _ = wallet.resubmit_transactions_from_cache().await;
+
+		// let's pause for awhile, to catch up with the transactions that need resubmission
+		pause_process_in_secs(10).await;
+
+		loop {
```
ebma, line 715:
```suggestion
		// We wait until the whole cache is empty because eventually all transactions should be handled
		loop {
```

ebma (commented):
It looks good overall, I just added some minor suggestions

clients/wallet/src/resubmissions.rs
b-yap: removed

clients/wallet/src/resubmissions.rs
b-yap: renamed, and added the interval param

ebma (approved):
I think this is good to go now. At least I don't see any other immediate problems with this implementation and the error handling seems to be the way we need it. Good job @b-yap  and ready for merge (once the CI errors are gone).

PR #392, opened 2023-09-01
263 add optimizations audit recommendations

gianfra-t:
Some notable mentions:

- LIY-04/LIT-05 was already solved since audit.

- LIV-01 "Duplicated Helper Function Call" -> I left this unchanged. In my opinion cannot be optimized without modifying (splitting) `try_deposit_collateral` from pallet vault-registry. Although in `deposit_collateral` extrinsic it is duplicated the active vault check, this function is used also externally outside the pallet.

- IMP-02 "Empty Strings as prefixes" -> I also left this unchanged. I believe this could be useful in the future, since it is already coded and works, for extension purposes.

- HOR-01/HOI-01 Unnecessary Variable cloning -> the code seems to have changed since then. Couldn't find that specific reference.

- EXC-02/EXU-03 Double filter() calls can be reduced with filter_map() -> Is addressed along with RPC-01, where the use of closures filter and transform the into `Request` type inside both  `get_vault_redeem_requests` and `get_old_vault_replace_requests`.

### DIFF
--- a/clients/runtime/src/rpc.rs
+++ b/clients/runtime/src/rpc.rs
@@ -1118,7 +1118,6 @@ pub trait RedeemPallet {
 	async fn get_redeem_period(&self) -> Result<BlockNumber, Error>;
 }
 
-
 #[async_trait]
 impl RedeemPallet for SpacewalkParachain {
 	async fn request_redeem(
@@ -1172,46 +1171,40 @@ impl RedeemPallet for SpacewalkParachain {
 		account_id: AccountId,
 		filter: Box<dyn Fn((H256, SpacewalkRedeemRequest)) -> Option<T> + Send>,
 	) -> Result<Vec<T>, Error> {
-			let head = self.get_finalized_block_hash().await?;
-			let result: Vec<H256> = self
-				.api
-				.rpc()
-				.request("redeem_getVaultRedeemRequests", rpc_params![account_id, head])
-				.await?;
-		
-			let redeem_requests = join_all(
-				result.into_iter().map(|key| async move {
-					self.get_redeem_request(key).await.map(|value| (key, value))
-				})
-			)
-			.await;
+		let head = self.get_finalized_block_hash().await?;
+		let result: Vec<H256> = self
+			.api
+			.rpc()
+			.request("redeem_getVaultRedeemRequests", rpc_params![account_id, head])
+			.await?;
 
-			let mut some_error: Option<Error> = None;
-			let filtered_results: Vec<_> = redeem_requests
-				.into_iter()
-				.filter_map(|item_maybe| 
-					match item_maybe {
-						Ok(item) => filter(item),
-						Err(e) => {
-							some_error.get_or_insert(e);
-							None
-						},
-					}
-				)
-				.collect();
-			match some_error {
-				Some(err) => Err(err),
-				None => Ok(filtered_results),
-			}
+		let redeem_requests = join_all(result.into_iter().map(|key| async move {
+			self.get_redeem_request(key).await.map(|value| (key, value))
+		}))
+		.await;
+
+		let mut some_error: Option<Error> = None;
+		let filtered_results: Vec<_> = redeem_requests
+			.into_iter()
+			.filter_map(|item_maybe| match item_maybe {
+				Ok(item) => filter(item),
+				Err(e) => {
+					some_error.get_or_insert(e);
+					None
+				},
+			})
+			.collect();
+		match some_error {
+			Some(err) => Err(err),
+			None => Ok(filtered_results),
 		}
+	}
 
 	async fn get_redeem_period(&self) -> Result<BlockNumber, Error> {
 		self.query_finalized_or_error(metadata::storage().redeem().redeem_period())
 			.await
 	}
 }
-
-
 
 #[async_trait]
 pub trait ReplacePallet {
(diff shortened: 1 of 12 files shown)


pallets/vault-registry/src/lib.rs
```diff
@@ -911,6 +911,7 @@ impl<T: Config> Pallet<T> {
 		vault_id: &DefaultVaultId<T>,
 		amount: &Amount<T>,
 	) -> DispatchResult {
+
```
ebma, line 914: Nit: Remove newline
```suggestion
```

clients/vault/src/cancellation.rs
```diff
@@ -168,18 +168,9 @@ impl<P: IssuePallet + ReplacePallet + Send + Sync> Canceller<P> for ReplaceCance
...
-		if has_expired(&requests[i]) {
-			let req = requests.remove(i);
-			expired.push(req);
-		} else {
-			i += 1;
-		}
-	}
-	expired
+
+	requests.drain(..).filter(|active_req| current_height > active_req.parachain_deadline_height).collect()
```
ebma, line 172: I think we have to be very careful here.
As the [docs](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.drain) say: Removes the specified range from the vector in bulk, returning all removed elements as an iterator. If the iterator is dropped before being fully consumed, it drops the remaining removed elements.

This means that the passed `requests` vector will be empty after calling this function. Whereas in the previous implementation, elements were only 'moved' from `requests` to `expired` if needed, keeping the original `requests` vector mostly intact.

clients/runtime/src/rpc.rs
```diff
@@ -1352,22 +1370,35 @@ impl ReplacePallet for SpacewalkParachain {
...
+			.into_iter()
+			.collect();
+
+		match redeem_requests {
+			Ok(redeem_requests) => {
+				let filtered_results: Vec<_> =
+					redeem_requests.into_iter().filter_map(|item| filter(item)).collect();
+					Ok(filtered_results)
+				},
+			Err(e) => Err(e),
```
ebma, line 1399: Instead of collecting and turning it into an iter again, can't we somehow combine and simplify these statements?

clients/runtime/src/rpc.rs
```diff
@@ -1165,32 +1167,47 @@ impl RedeemPallet for SpacewalkParachain {
...
+			)
+			.await
+			.into_iter()
+			.collect();
+		
+			match redeem_requests {
+				Ok(redeem_requests) => {
+					let filtered_results: Vec<_> = redeem_requests
+						.into_iter()
+						.filter_map(|item| filter(item))
```
ebma, line 1195: Same here, can we prevent the need for a second `.into_iter()`?

ebma (changes requested):
I looked through the changes and think you did a great job @gianfra-t. I left some nitpicky comments but also one more severe issue related to changes for 'LBS-01 - Duplicated condition check'. At least I think it is, let me know your thoughts.

clients/vault/src/cancellation.rs
gianfra-t: Indeed!, I just commited a new impl for this, so that the functionality is the same as before. I think this is what may have caused the error in Integration test (still not 100% sure).

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -124,14 +124,14 @@ impl ScpMessageCollector {
 					"Proof Building for slot {slot}: not enough envelopes to build proof "
 				);
 			} else {
-				return UnlimitedVarArray::new(envelopes.clone()).unwrap_or(empty)
+				return Some(UnlimitedVarArray::new(envelopes.clone()).unwrap_or(empty))
```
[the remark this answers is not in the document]
gianfra-t: Yes! good catch.

primitives/src/lib.rs
```diff
@@ -844,41 +844,47 @@ impl TransactionEnvelopeExt for TransactionEnvelope {
...
 
+		let mut transferred_amount: StellarStroops = 0;
+
 		let tx_operations: Vec<Operation> = match self {
 			TransactionEnvelope::EnvelopeTypeTxV0(env) => env.tx.operations.get_vec().clone(),
 			TransactionEnvelope::EnvelopeTypeTx(env) => env.tx.operations.get_vec().clone(),
-			TransactionEnvelope::EnvelopeTypeTxFeeBump(_) => Vec::new(),
-			TransactionEnvelope::Default(_) => Vec::new(),
+			TransactionEnvelope::EnvelopeTypeTxFeeBump(_) => return BalanceConversion::unlookup(transferred_amount),
+			TransactionEnvelope::Default(_) => return BalanceConversion::unlookup(transferred_amount),
```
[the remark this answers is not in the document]
gianfra-t: Sure. In fact in the comments from the audit, the "extra computation" was just this check (if len>0). That's why I changed it. But I agree it is more readable like you say, and not really a performance issue. We can rollback on that one.

clients/runtime/src/rpc.rs
gianfra-t: Sure, you are right, no real need to collect redeem_requests in the first place.

gianfra-t:
> I looked through the changes and think you did a great job @gianfra-t. I left some nitpicky comments but also one more severe issue related to changes for 'LBS-01 - Duplicated condition check'. At least I think it is, let me know your thoughts.

Regarding LBS-01 (090c314ebae8398223c07dec77a0d2276d8fa476), I understood that `get_active_vault_from_id` will match if vault if active already, if not return an error (which I mapped to the desired one in the second check). Let me know if I am missing something.

gianfra-t:
> > I looked through the changes and think you did a great job @gianfra-t. I left some nitpicky comments but also one more severe issue related to changes for 'LBS-01 - Duplicated condition check'. At least I think it is, let me know your thoughts.
> 
> Regarding LBS-01 ([090c314](https://github.com/pendulum-chain/spacewalk/commit/090c314ebae8398223c07dec77a0d2276d8fa476)), I understood that `get_active_vault_from_id` will match if vault if active already, if not return an error (which I mapped to the desired one in the second check). Let me know if I am missing something.

Never Mind, it is matching `Active(_) not Active(true)`. Will rollback in that one too.

ebma:
> Never Mind, it is matching Active(_) not Active(true). Will rollback in that one too.

Exactly. A small but important difference. I thought I left a review comment on that specific line but maybe I accidentally discarded it again. At least I can't find it anymore haha. 
Anyways, it's actually an oversight of the auditors and not on you since you just did what they asked us to do.

ebma (commented):
Looks good to me overall. Please install the pre-commit hook and also run `cargo fmt --all` to make sure that the CI does not complain (which it will because of [this](https://github.com/pendulum-chain/spacewalk/blob/ad68c16f55141357d14df66d40717b42d089c96c/.github/workflows/ci-main.yml#L78). You can then turn the PR from 'draft' to 'ready for review' and once CI passed, I'll approve

ebma:
@gianfra-t the CI job threw some warning because clippy failed, can you please add the suggested change and re-run the job?

ebma (approved):
LGTM and ready for merge

Issue #257, opened 2023-02-13
Fix 'informational' issues (audit recommendations)


ebma:
There are some very minor things brought up in the audit that we might want to fix. The following checklist only includes the item ID; the description and details can be found by looking at the audit report.
- [x] TYP-01 Confusing Variable Naming
- [x] SYS-03 Unnecessary Variable
- [x] STA-02 Lack of Validation for `destination_adress` on `send_payment_to_address()`
- [x] STA-01  Code duplication
- [x] SRW-01 Usage of hard-coded string
- [x] SRC-01 Unused Methods and Storage
- [x] PRO-01 Unhandled Error
- [x] PAE-02 Usage of magic numbers
- [x] PAE-01 Unnecessary `Result<>` Return Type
- [x] LIT-04 Reduce using `unwrap()` and `expect()` in Production Codebase
- [x] LIT-03 Panic can happen between correlated storage modifications
- [x] LII-01 Inconsistent match expression
- [x] LIE-04 Mismatch in variable name and pallet
- [x] LIL-01 Dead Code in `execute_replace()`
- [x] IMP-01 Same Behavior Defined for different conditions
- [x] EXC-01 Missing information in logging message
- [x] CLI-03 Incorrect error type thrown
- [x] CLI-02 Typos
- [x] CLI-01 Confusion function naming
- [ ] #417
- [x] 4FC-04 Logic should be moved to a separate function - refactoring
- [x] 4FC-03 Unused errors
- [x] 4FC-02 Inconsistent comments

Since the changes are so minor, it's okay to include all of them in one branch and one Pull Request. But each of them should at least be addressed in a separate commit with a proper name, also mentioning the **ID** in the name so that it's easier to track the changes.

additional:

- [ ] GLOBAL-05 Unneceessary Off-Chain User Protection Mechanism
- [x] LI5-02 TryFrom `CurrencyId` implementations Contain Repeated Code
- [x] LIH-03 Values Length Not Validated in `feed_values` Function

vadaynujra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/257?planning-poker) with Zenhub @b-yap @ebma @TorstenStueber

b-yap:
EXC-01 -  Missing information in logging message
was completed by https://github.com/pendulum-chain/spacewalk/pull/377

[...]

PR #393, opened 2023-09-04
257 fix informational issues part 4

b-yap:
parent: https://github.com/pendulum-chain/spacewalk/pull/385
continues from: https://github.com/pendulum-chain/spacewalk/pull/390
 CI issues addressed by #414 

This PR addresses the ff. items:
* ## STL-02 | Code Duplication
    * ### **from audit:**
       * On stellar_wallet.rs, there is multiple code duplication to get the account_id. Every time public_key_encoded is used, it is only to calculate the account id.
       * It would be better to abstract this shared logic into an auxiliary function get_account_id() to avoid code duplication and improve the code's readability.
    * ### **files covered:** 
      * clients/wallet/src
        * cache.rs
        * horizon/
          * tests.rs
        * lib.rs
        * operations.rs
        * stellar_wallet.rs
      * as I can't seem to find where `public_key_encoded` is used, I believe it's fixed by a trait called  [StellarTypeToString](https://github.com/pendulum-chain/spacewalk/pull/365/files#diff-93cb951626774323817cabfcdf4c2db9940f66818168e287b6cb9fd725dbb0b0R110-R119)?

* ## SRC-01 | Unused Methods and Storage
    * ### **from audit:**
       * methods and storage are not used by the pallet or other pallets as they don't have any entry point.
    * ### **files covered:** 
      * pallet/vault-registry/src
        * types.rs
        * lib.rs
          * as I cannot find the commit hash posted, I based on https://github.com/pendulum-chain/spacewalk/commit/b5307e9c1e805bafe4546e661df056c96548a878 instead  

* ## 9B2-05 | Commented Out Code
    * ### **files covered:** 
      * clients/runtime/src/rpc.rs
      * wallet/src/horizon/responses.rs
         * [some of them do not exist anymore](https://github.com/pendulum-chain/spacewalk/issues/257#issuecomment-1706323568)

* ## LBC-03 | Inconsistent `match` Expression
    * ### **from audit:**
       * If the team wants to propagate the issueID when IssueRequestStatus==Cancelled, please provide the reason behind the choice, and no further action is needed.
    * ### **files covered:** 
      * pallets/issue/src/lib.rs
        * I renamed the function `get_issue_request_from_id()` to `get_executable_issue_request_from_id()` to indicate that even "Cancelled" issue request can be executed, depending on `ensure_accepting_new_issues()` which I moved to an earlier spot. 
        * created `fn validate_stellar_transaction()` to validate the transaction. A valid transaction should be able to return an amount.
       
* ## LIY-03 | Reduce Using `unwrap()` and `expect()` in Production Codebase
    * ### **from audit:**
       * We advise the team to consider using the pattern-match or ? operator to replace the unwrap() usage and further implement the error handling when panic is undesired.
    * ### **files covered:** 
      * pallets/stellar-relay/src/lib.rs
         * [some of them do not exist anymore](https://github.com/pendulum-chain/spacewalk/issues/257#issuecomment-1707960973) 
         * as we actually want it to panic, it is totally acceptable. But rather than `expect`, i used `assert!` instead 

* ## TYL-01 | Confusing Variable Naming
    * ### **from audit:**
       * We advise the team to rename the linked variables to more descriptive ones.
    * ### **files covered:** 
      * clients/vault/src/oracle
        * types.rs 
        * collector
           * collector.rs
           * proof_builder.rs 
              * renamed as best as I think (and open for modification) 

* ## LIL-01 | DEAD CODE IN  execute_replace()
    * ### **from audit:**
       * The  Completed  requests are filtered out at the beginning of the
function ( confirmed also by the comment in the code ). Therefore, that branch represents dead code and it can be removed
     * ### **files covered:** 
        * pallets/replace/src/lib.rs - we decided to keep it, just being on the safe side. We updated the comment.
        
* ## LIT-03 | PANIC CAN HAPPEN BETWEEN CORRELATED STORAGE MODIFICATIONS
    * ### **from audit:**
       * The conversion of  new_validator_vec
and  new_organization_vec  to  BoundedVec  should happen between different but related storage's  put  operations
(operations that save data to the storage).
    * ### **files covered:** 
        * pallets/stellar_relay/src/lib.rs - rearranged the order of the code


pallets/issue/src/lib.rs
```diff
@@ -855,4 +823,56 @@ impl<T: Config> Pallet<T> {
 		}
 		Ok(())
 	}
+
+	fn validate_stellar_transaction(
```
ebma, line 827: The refactoring is nice but maybe we should split this function even further. It does not only validate but also extract the amount which is confusing judging from the naming. 

So how about we remove lines 870-876 from here and put it back to the outer function? We would then change this function to then return the `transaction_envelope` so that it can be used in the outer function.

clients/vault/src/oracle/types.rs
```diff
@@ -134,50 +140,58 @@ impl<K: PartialEq + Debug, T> Default for LimitedFifoMap<K, T> {
...
-/// The slot is not found in the `StellarMessage::TxSet(...)`, therefore this map
-/// serves as a holder of the slot when we hash the txset.
 #[derive(Clone)]
-pub struct TxSetHashAndSlotMap {
-	hash_slot: HashMap<TxSetHash, Slot>,
-	slot_hash: HashMap<Slot, TxSetHash>,
+pub struct DoubleSidedHashMap<T1, T2> {
+	t1_to_t2_map: HashMap<T1, T2>,
+	t2_to_t1_map: HashMap<T2, T1>,
 }
```
ebma, line 147: I really like the idea, just the naming of `T1` and `T2` is a bit lacking. How about we just use some letters and no number?
Something like `<K, V>` and then we have `k_to_v_map` and `v_to_k_map`. K and V would loosely resemble key and value (which does not make perfect sense but at least has a familiar ring to it).

ebma (commented):
While not explicitly requested in LIY-03, should we try to reduce those usages also in the vault client side? The raised issue is more related to reducing the usage in our runtime, but maybe we should improve that also in the client side. WDYT @b-yap?

pallets/issue/src/lib.rs
b-yap: updated

clients/vault/src/oracle/types.rs
b-yap: updated

pallets/vault-registry/src/lib.rs
```diff
@@ -1133,9 +1132,10 @@ impl<T: Config> Pallet<T> {
 	///
 	/// # Arguments
 	/// * `issue_id` - secure id for generating deposit address
+	#[cfg(feature = "runtime-benchmarks")]
 	pub fn register_deposit_address(
```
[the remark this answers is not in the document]
b-yap: done

clients/wallet/src/stellar_wallet.rs
```diff
@@ -708,7 +708,8 @@ mod test {
 
 				// new wallet created, with the previous destination address acting as "SOURCE".
 				let temp_wallet =
-					wallet_with_secret_key_for_storage(storage_path, inactive_secret_key);
+					wallet_with_secret_key_for_storage(storage_path, inactive_secret_key)
+						.expect("should return an arc rwlock wallet");
```
ebma, line 712: Nit: Let's maybe rename all of them to 
```
"should return a wallet instance"
```
because I don't need to know that it's `Arc` and `RwLock` in this case.

ebma (commented):
Looks good 👍 we should probably wait until we found a way to fix the integration tests though because otherwise we would not know if these changes actually introduce other problems.

b-yap:
ubuntu passed; should be enough. 
But we have to take note: https://github.com/pendulum-chain/spacewalk/actions/runs/6480037206/job/17594746429#step:12:501

Issue #397, opened 2023-09-11
Add clients-info crate for auto-updating


ebma:
Our auto-updating mechanism needs to have a way to determine the location of the latest release. This information should be stored on-chain.

## TODO
For this, we can re-use the implementation of [this](https://github.com/interlay/interbtc/tree/master/crates/clients-info) crate. We can probably copy it as-is to our `spacewalk/pallets`. 
- [ ] Create `clients-info` crate
- [ ] Add and configure it for the spacewalk testchain

PR #399, opened 2023-09-12
Addition of pallet client-info to store data of current vault client

gianfra-t:
Stores URI and checksum of current vault client binary. 
Also stores the next client information to be utilized after upgrade, which will be set after upgrade hook is triggered.

Refer to [this](https://github.com/pendulum-chain/spacewalk/issues/397) issue.


pallets/client-info/README.md
```diff
@@ -0,0 +1,12 @@
+# Client info pallet
+This pallet contains storage items regarding the vault client info, and 2 extrinsics to set this value.
+
+## Testing
+
+To run the tests use:
+
+```bash
+cargo test --package redeem --features runtime-benchmarks
```
ebma, line 9:
```suggestion
cargo test --package client-info --features runtime-benchmarks
```

pallets/client-info/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "client-info"
```
ebma, line 2: Let's maybe stick to interlay's naming of 
```suggestion
name = "clients-info"
```
although for the time being we will only use it for the vault and no oracle or faucet client.

pallets/client-info/README.md
```diff
@@ -0,0 +1,12 @@
...
+```
+
+
```
ebma, line 12: Let's also add the instructions for benchmarking here, as we do in most other crates

```suggestion

## Benchmarking

Build the node with the `runtime-benchmarks` feature:

```bash
cargo build --package spacewalk-standalone --release --features runtime-benchmarks
```bash

```bash
# Show benchmarks for this pallet
./target/release/spacewalk-standalone benchmark pallet -p clients-info -e '*' --list
```bash

Run the benchmarking for a pallet:

```bash
./target/release/spacewalk-standalone benchmark pallet \
--chain=dev \
--pallet=clients-info \
--extrinsic='*' \
--steps=100 \
--repeat=10 \
--wasm-execution=compiled \
--output=pallets/clients-info/src/default_weights.rs \
--template=./.maintain/frame-weight-template.hbs
```bash

```

ebma (commented):
Looks good overall. @gianfra-t can you try to re-generate the `default_weights` file for our spacewalk-standalone chain by using the commands I added in one of my suggestions?

pallets/client-info/Cargo.toml
ebma: We'd also need to adjust the name of the directory and usages in other toml files.

gianfra-t:
Thanks @ebma for the suggestions!. I have renamed the pallet and regenerated the benchmarks

pallets/clients-info/src/default_weights.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct SubstrateWeight<T>(PhantomData<T>);
+impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
+	/// Storage: ClientsInfo CurrentClientReleases (r:0 w:1)
+	/// Proof: ClientsInfo CurrentClientReleases (max_values: None, max_size: Some(562), added: 3037, mode: MaxEncodedLen)
+	/// The range of component `n` is `[0, 255]`.
+	/// The range of component `u` is `[0, 255]`.
+	fn set_current_client_release(_n: u32, _u: u32, ) -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `0`
```
ebma, line 47: @gianfra-t any idea why these seem to be ignored in our benchmarks? I can see in interlay's weights that the `n` and `u` parameters are actually used for calculating the weights. 🤔

gianfra-t: I don't know! I will try running the command interlay used to see if it has to do with it.

gianfra-t: Actually, by re running the command, every new run seems to be random in how it tests (or not) each variable 🤔

ebma: So you are saying that sometimes the variables are included and sometimes they are not? 😶

gianfra-t: Okay, I believe that what is going on is that in some scenarios the time influence of the parameter is near 0 so it gets ignored.

ebma: Ah I see, thanks for looking into it again 👍

ebma (approved):
Looking good, nice job @gianfra-t 👍

PR #400, opened 2023-09-14
398 client runner

gianfra-t:
Addition of an implementation of the [runner client](https://github.com/interlay/interbtc-clients/tree/master/runner) taken from interlay clients.

In summary, the runner fetches periodically for the on-chain information about the vault client, both download URI and checksum. It will then proceed to download and execute it, also checking for new releases and handling them. 

The only difference on this implementation is:
- The runner client will now verify checksum of the vault client once it is downloaded for the first time, rejecting the download if it does not match.
- The runner will retry the download if it fails (for instance, if the checksum is incorrect)


runner/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "0.1.0"
+edition = "2021"
+
+
+[dependencies]
+clap = { version = "4.0.17", features = ["derive"]}
+hex = "0.4.3"
+tokio = { version = "1.8", features = ["rt-multi-thread", "macros", "time"] }
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "full", "bit-vec"] }
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
```
ebma, line 12: Let's keep this in sync with the other dependencies and use `polkadot-v0.9.40`

clients/README.md
```diff
@@ -106,10 +106,10 @@ After making the changes, the first step is to run the local node.
 cargo run --bin spacewalk-standalone --release -- --dev
 ```
 
-Once the local node is running, run this command:
+Once the local node is running, run this command from the clients directory:
```
ebma, line 109: Nice, can we change the command in line 92 to specify installation of a fixed subxt version? Something like `cargo install subxt-cli@0.31.0`

runner/Cargo.toml
ebma, line 1: I would say we move the whole `runner` directory into the `clients` directory instead of placing it in root.

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		let s = match self {
+			ClientType::Vault => "vault",
+		};
+		write!(f, "{s}")
+	}
+}
+
+/// Pallet in the parachain where the client release is assumed to be stored
+pub const PARACHAIN_MODULE: &str = "ClientInfo";
```
ebma, line 59: I assume this matters.
```suggestion
pub const PARACHAIN_MODULE: &str = "ClientsInfo";
```

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+
+/// Storage item in the Pallet where the client release is assumed to be stored
+pub const CURRENT_RELEASES_STORAGE_ITEM: &str = "CurrentClientReleases";
+
+/// Parachain block time
+pub const BLOCK_TIME: Duration = Duration::from_secs(6);
```
ebma, line 65: Should be 12 seconds actually (?)

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+
+/// Timeout used by the retry utilities: One minute
+pub const RETRY_TIMEOUT: Duration = Duration::from_millis(60_000);
+
+/// Waiting interval used by the retry utilities: One minute
```
ebma, line 70: Comment does not match the actual value of `1_000` milliseconds.

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+
+/// Multiplier for the interval in retry utilities: Constant interval retry
+pub const RETRY_MULTIPLIER: f64 = 1.0;
+
+/// Data type assumed to be used by the parachain to store the client release.
+/// If this type is different from the on-chain one, decoding will fail.
+#[derive(Decode, Default, Eq, PartialEq, Debug, Clone)]
+pub struct ClientRelease {
+	/// A link to the Releases page of the `interbtc-clients` repo.
+	/// Example: https://github.com/interlay/interbtc-clients/releases/download/1.16.0/vault-parachain-metadata-interlay
```
ebma, line 81: Let's adjust the comment and replace the link with this one. 
`https://downloads.pendulumchain.tech/index.html`

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+	let mut hasher = Sha256::default();
+	hasher.input(bytes);
+	hasher.result().as_slice().to_vec()
+}
+
+/// Per-network manager of the client executable
+pub struct Runner {
+	/// `subxt` api to the parachain
+	subxt_api: OnlineClient<PolkadotConfig>,
+	/// The child process (interbtc client) spawned by this runner
```
ebma, line 108:
```suggestion
	/// The child process (Spacewalk client) spawned by this runner
```

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+		let _ = retry_with_log_async(
+			|| {
+				Runner::do_download_binary(bin_path.clone(), release.clone())
+					.into_future()
+					.boxed()
+			},
+			"Error fetching executable".to_string(),
+		)
+		.await?;
+		//Runner::do_download_binary(runner ,release.clone()).await?;
```
ebma, line 171: Can we remove this line?

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+		// Read the content of the file
+		let mut file_content = Vec::new();
+		file.read_to_end(&mut file_content)?;
+
+		let checksum = H256::from_slice(&sha256sum(&file_content));
+
+		if checksum != release.checksum {
+			return Err(Error::IncorrectChecksum)
+		}
+		Ok("success")
```
ebma, line 213: If we don't care about the content of the Result, let's just go for a `Result<(), Error>` and remove the `"success"`

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+			if let Some(new_release) = runner.try_get_release().await? {
+				let maybe_downloaded_release = runner.downloaded_release();
+				let downloaded_release =
+					maybe_downloaded_release.as_ref().ok_or(Error::NoDownloadedRelease)?;
+				if new_release.checksum != downloaded_release.checksum {
+					log::info!("Found new client release, updating...");
+
+					// Wait for child process to finish completely.
+					// To ensure there can't be two client processes using the same resources (such
+					// as the Bitcoin wallet for vaults).
```
ebma, line 376:
```suggestion
					// as the Stellar wallet for vaults).
```

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+		for client in clients {
+			let mut runner = MockRunner::default();
+			runner.expect_download_path().return_const(mock_path.clone());
+			runner.expect_client_type().return_const(client.clone());
+			let (bin_name, bin_path) = Runner::get_bin_path(
+                &runner,
+                "https://github.com/interlay/interbtc-clients/releases/download/1.17.2/vault-parachain-metadata-kintsugi",
+            )
+                .unwrap();
+			assert_eq!(bin_name, "vault-parachain-metadata-kintsugi");
```
ebma, line 800: Can we change this to something spacewalk-related?

runner/src/runner.rs
```diff
@@ -0,0 +1,1127 @@
...
+			"rpcpassword",
+			"--keyfile",
+			"keyfile.json",
+			"--keyname",
+			"0xa81f76187f1e5d2059f67439c4242a92a5cd66a409579db73f156c6e2aae5102",
+			"--faucet-url",
+			"http://localhost:3033",
+			"--auto-register=KSM=faucet",
+			"--btc-parachain-url",
+			"ws://localhost:9944",
```
ebma, line 881: I assume we actually have to change these arguments because some of them are not supported by our vault client.

ebma (changes requested):

ebma (commented):
Looks good overall, I just left some comments about minor issues.

runner/src/runner.rs
gianfra-t: It does!!

gianfra-t:
Thanks for finding [this](https://github.com/pendulum-chain/spacewalk/pull/400#discussion_r1326310708) @ebma! on that note, let me know if we should open a separate task to build an integration test for the runner.

gianfra-t:
The CI seems to be failing when [connecting to stellar](https://github.com/pendulum-chain/spacewalk/blob/31df1832483199b978ae07f6ed4e018a11ff4dba/clients/stellar-relay-lib/src/connection/services.rs#L21), perhaps we retry later?

ebma:
> On that note, let me know if we should open a separate task to build an integration test for the runner.

How would you suggest we approach this kind of test?

ebma (approved):
The code looks good overall. 

How sure are we that this works as expected though @gianfra-t? You said you tested it locally once and it worked with downloading the binary from a target location and replacing the current vault executable, correct?

gianfra-t:
> The code looks good overall.
> 
> How sure are we that this works as expected though @gianfra-t? You said you tested it locally once and it worked with downloading the binary from a target location and replacing the current vault executable, correct?

Yes. I tested locally with a release of the client and also tested updating it on a local chain. It detected the changes and downloaded/started the new one.

ebma:
I also tested it and can confirm that the runner manages to read the URL stored in the clients-info pallet and downloads the release from there. However, the checksum did not match. 
@gianfra-t During testing, how did you calculate the checksum for the `clientsInfo::setCurrentClientReleases()` extrinsic? Using the 'hash a file' option provided by polkadot.js does not seem to work, or at least the result of that hashing process is not the same as the result of the runner hashing the downloaded binary.

To make it work regardless, I just used the hash that the runner calculates and put that on chain. Afterward, downloading etc. worked. But this is not an ideal solution of course.

gianfra-t:
I actually had the same problem, the polkadot.js hash did not work. I thought it was a problem with it so I used [this tool](https://emn178.github.io/online-tools/sha256_checksum.html). Do you think it is a different hashing algorithm/implementation? 
I it presents a problem I guess I could try a different hashing library on the runner that matches the one created by polkadot.js

ebma:
Yes, let's try to use the same hashing algorithm for convenience. I tried to find what polkadot.js uses and I think it is blake2. The UI calls [this](https://github.com/polkadot-js/apps/blob/2ea78003faed0e63eb9c3dec1694508f45624e9d/packages/react-params/src/Param/Hash256.tsx#L22-L23) function and according to [this](https://github.com/polkadot-js/api/blob/0dcf349831a9b62acc4f00612e0c63b2c8b42de4/packages/types/src/create/registry.spec.ts#L163-L164) test it uses blake2 by default but can be overridden with a custom definition [here](https://github.com/polkadot-js/api/blob/0dcf349831a9b62acc4f00612e0c63b2c8b42de4/packages/api/src/base/Init.ts#L92). I think we don't specify a custom hasher so let's try hashing with blake2 in the vault client and see if the checksum then matches the one generated by polkadot.js upon file upload.

gianfra-t:
I would say It really boils down to our deployment workflow we want.
If we plan to have a release of the client that is automated then I would say stick with this hashing algorithm and do something similar to interlay.

ebma:
Alright. Let's stick to sha256 and not rely on the hash shown in polkadot.js. It's just a UI and we can use other means, like e.g. the tool you linked to. 
So in theory this PR is ready for merge now but I'm hesitant to merge it just yet because the CI is failing.

gianfra-t:
It's the issue related to stellar update, so let's wait for that to be solved and we can merge those changes.

Issue #406, opened 2023-10-04
Add `on_initialize` hook to `reward-distribution` pallet


TorstenStueber:
This ticket is a part of https://github.com/pendulum-chain/tasks/issues/47

This ticket depends on https://github.com/pendulum-chain/spacewalk/issues/405

Add an `on_initialize` hook
  * if `RewardPerBlock` is `None`, return
  * call `ensure_parachain_status_running` of the security pallet and return if error
  * call `parachain_block_expired` of the security pallet providing `RewardsAdaptedAt` and the `DecayInterval` - 1 (because `parachain_block_expired` checks for `>` instead of `>=`). If it is `true`, then multiply `RewardPerBlock` by `DecayRate` and set `RewardsAdaptedAt` to the current active block number
  * for each collateral currency id in the system
     * call `total_stake` of `PooledRewards`, where `pool_id` is that currency id
     * determine the USD value of the total locked collateral of this currency id (using `convert_to` of the `Amount` type)
  * add up the USD value over all currency ids
  * for each collateral currency id in the system   
     * determine the percentage of the USD value of this currency id over the total USD value (as a `Perquintill` using `Perquintill::from_rational`)
     * call `distribute_reward` of `PooledRewards`, where `pool_id` is the currency id, `currency_id` is the native token, `reward` is `RewardPerBlock` multiplied with the percentage

annatekl:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/406?planning-poker) with Zenhub @adelarja @b-yap @ebma @gianfra-t @TorstenStueber

Issue #404, opened 2023-10-04
Add pooled reward pallet


TorstenStueber:
This ticket is a part of https://github.com/pendulum-chain/tasks/issues/47

Add the current reward pallet from Interlay to Spacewalk: https://github.com/interlay/interbtc/tree/1b90616692dcce55749616b4c60bdab228279909/crates/reward

This pallet is different from the old reward pallet in Spacewalk because it supports pools, a feature we need for the rewards system.

We should call this pallet `PooledRewards`.

## Additional tasks
* Identify all places in Spacewalk that reserve or unreserve collateral of vaults of nominators.
   * Every time collateral is reserved, call `deposit_stake` of `PooledRewards`, where `pool_id` is the currency id of the collateral, `stake_id` is the account id of the vault itself or the nominator account, `amount` is the increment of reserved collateral
   * Every time collateral is unreserved, call `withdraw_stake` accordingly

annatekl:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/404?planning-poker) with Zenhub @adelarja @b-yap @ebma @gianfra-t @TorstenStueber

PR #415, opened 2023-10-10
404 add pooled reward pallet

gianfra-t:
Issue #404 and #406, part of this [pendulum task](https://github.com/pendulum-chain/tasks/issues/47).

Adds the `pooled-rewards` pallet, that will allow to distribute the static native currency rewards for the nominators/operators of vaults, as well as the fee distribution in arbitrary currency.

This pallet is taken [from interlay ](https://github.com/interlay/interbtc/tree/1b90616692dcce55749616b4c60bdab228279909/crates/reward).
This will also replace the current `rewards` pallet that is used to distribute fees currently.

At any point the stake of each `vault_id` on the `pooled-rewards` pallet for it's corresponding collateral currency pool is maintained to be equal to that `vault_id`'s  `total_staking` amount from the staking pallet. See for example [here](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/vault-registry/src/pool_staking_manager.rs#L57).

This update is performed whenever the stake for the `vault_id` changes, and is managed by the [`pool_staking_manager` interface](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/vault-registry/src/pool_staking_manager.rs#L4):
- Deposit and Withdraw of collateral by a nominator [here](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/nomination/src/lib.rs#L244) and [here](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/nomination/src/lib.rs#L283)
- When slashing occurs [here](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/vault-registry/src/lib.rs#L1029) 
- When vault operator [withdraws](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/vault-registry/src/lib.rs#L946) or [deposits](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/vault-registry/src/lib.rs#L923)
- [Opt out](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/nomination/src/lib.rs#L325) of nomination.
- Vault Replace [here](https://github.com/pendulum-chain/spacewalk/blob/dd4f8edd119392a9dbebaecc92eda34165d4da43/pallets/vault-registry/src/lib.rs#L1505) 

This PR will also merge changes for #406, where reward percentage per collateral pool is calculated (see [here](https://github.com/pendulum-chain/spacewalk/blob/718345e8761c399b8b58baf84e1ae9e24d3174ad/pallets/reward-distribution/src/lib.rs#L254)), and exposes a [function](https://github.com/pendulum-chain/spacewalk/blob/718345e8761c399b8b58baf84e1ae9e24d3174ad/pallets/reward-distribution/src/lib.rs#L246) used for fee pallet to distribute arbitrary rewards.

### Additions to reward-distribution pallet.
- An `on_initialize` hook, that calculates the block specific reward, and updates it if required.
- A `distrubute_reward` function that calculates the percentages in USD of the total stake, and distributes to each collateral pool a proportional amount. This function will be used both from external pallets (eg Fee) and to distribute the per-block rewards in native token.
- Trait interfaces for this function to be called from other pallets.  
- A new extrinsic to collect the rewards for a given `vault_id`, `account_id` pair.

The [collect_reward](https://github.com/pendulum-chain/spacewalk/blob/000d0c01521e3b4bc62475298fa13f78f4c47d46/pallets/reward-distribution/src/lib.rs#L162) extrinsic replaces the extrinsic previously located in the fee pallet which only allowed to withdraw the reward for the wrapped currency and native( see [here](https://github.com/pendulum-chain/spacewalk/blob/000d0c01521e3b4bc62475298fa13f78f4c47d46/pallets/fee/src/lib.rs#L505))


pallets/reward-distribution/src/lib.rs
```diff
@@ -99,5 +161,130 @@ pub mod pallet {
...
+		//prices are unchanged
+		let bounded_vec = BoundedVec::try_from(percentages_vec)
+			.expect("should not be longer than max currencies");
+		RewardsPercentage::<T>::set(Some(bounded_vec));
+
+		Ok(error_reward_accum)
+	}
+}
+//Distribute Rewards interface
+pub trait DistributeRewardsToPool<Balance, CurrencyId> {
```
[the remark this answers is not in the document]
gianfra-t: Agreed, I will change it to `DistributeRewards`.

pallets/reward-distribution/src/lib.rs
```diff
@@ -99,5 +161,130 @@ pub mod pallet {
...
+			}
+
+			percentages_vec.push((currency_id, percentage))
+		}
+
+		//we store the calculated percentages which are good as long as
+		//prices are unchanged
+		let bounded_vec = BoundedVec::try_from(percentages_vec)
+			.expect("should not be longer than max currencies");
+		RewardsPercentage::<T>::set(Some(bounded_vec));
```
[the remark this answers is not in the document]
ebma: > I left it because I wasn't sure that the price inside one block could change. If it couldn't and the prices change at the beginning only (given the oracles off-chain workers), I thought we could use this percentages and save the first iteration to calculate the amount for the fee distributions.

The price feeding happens as part of a regular extrinsic call. The 'specialty' about this is that this call is done by the off-chain worker but at the end of the day it does not really matter. It's just that every block will have a call to `diaOracleModule::setUpdatedCoinInfos()`, as you can see in eg. [this](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Famplitude-rpc.dwellir.com#/explorer/query/0x90d32c4894e8d47081b947fff8d4dd9b341d6ad5628b50227b28960ef7837e13) block.
But regardless of that, I am not sure how we could use the `RewardsPercentage` in order not to do the first iteration. 

We have to assume that the price of each asset changes every block (since the off-chain worker feeds a new price every block, and it may or may not be the same as in the last block but we can't know).
Thus, even if the values of the `total_stakes` iter did not change between the blocks (which we also don't know), the returned results of any call to `T::OracleApi::currency_to_usd` might change and thus the whole computation has to be re-done.

Or am I missing something? As far as I understand, storing the `RewardsPercentage` on-chain only makes sense if we also add some logic to know if the stakes of the pooled rewards pallet changed or the prices. And this logic would probably cause more overhead than just redoing the calculation each time.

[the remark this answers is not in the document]
gianfra-t: Thanks for the explanation, it is more clear now. Indeed, it does not make sense to hold this vector at all. I will remove it.

pallets/reward-distribution/src/lib.rs
```diff
@@ -99,5 +155,144 @@ pub mod pallet {
...
+
+			if rewards == (BalanceOf::<T>::zero()) {
+				return Err(Error::<T>::NoRewardsForAccount.into())
+			}
+
+			//either transfer from fee account if the reward is
+			//not the native currency, or just mint it if it is
+			let native_currency_id = T::GetNativeCurrencyId::get();
+			if reward_currency_id == native_currency_id {
+				T::Balances::deposit_creating(&caller, rewards);
```
ebma, line 189: There is a potential flaw here when it comes to handling the native currency. The `collect_reward()` function resembles the `Fee::withdraw_rewards()` extrinsic, but it differs in the way that you can specify the `reward_currency_id` that you want to get paid out. 

However, these two extrinsics differ in another way: Here, we mint the rewards for the native currency, and in `Fee::withdraw_rewards()`, they are paid out from the reward pool. 
Now the problem is that, if a user tried to call `Fee::withdraw_rewards()`, the `Fee` pallet would not be able to pay out the whole reward of the user because it just does not have that many native tokens to transfer to them. Because they are not minted and just imaginary yet. 

In this `collect_reward` extrinsic we do mint these native tokens. But we neglect the native tokens that were collected into the fee pool. 

I would suggest that we remove the `Fee::withdraw_rewards()` extrinsic altogether, since the `collect_reward` extrinsic offers the same logic but better. 

Besides that, we should change this to first transfer all the native currency that is available in the fee pool and then pay out the rest with `deposit_creating`. Maybe something like this pseudo_code
```
let available_fees_native = T::Balances::balance_of(&Self::fee_pool_account_id(), native_currency_id);
let amount_fees_native: currency::Amount<T> = Amount::new(rewards, native_currency_id);
amount.transfer(&Self::fee_pool_account_id(), &caller)?;
let remaining_rewards = rewards.checked_sub(available_fees_native)?;
T::Balances::deposit_creating(&caller, remaining_rewards);
```

gianfra-t: Thanks @ebma this is a very good observation. Indeed the extrinsic on the fee pallet I left it only to discuss about it but also had the idea of removing it, but nevertheless I did not think about the difference in where the native tokens are coming from. 
Besides removing this extrinsic and it's corresponding code, I had the idea that it would never be the case that a fee is paid in native currency, since only wrapped currencies are the one issued or redeemed which end up as fees on the fee account. Is this correct?
If that case is correct, then we should always pay the native rewards per block by minting (deposit_creating), and we would not have the problem of dealing with native currency in the fee account.

ebma: Right, good point. The griefing is also the only part that I was worried about. 
You can see [here](https://github.com/pendulum-chain/pendulum/blob/08b2eef8b43d216cc176571d75052dee22866b22/runtime/amplitude/src/lib.rs#L1165) that the currency we define for locking the griefing collateral is native.
However, in case this fee is indeed consumed because the user did not execute the issue request in time (ie. the user was griefing), the griefing fee is [transferred directly to the vault accont ](https://github.com/pendulum-chain/spacewalk/blob/8bb0d639c6f2e23126f72e39f56c6a497cf2e2d4/pallets/issue/src/lib.rs#L570)and not distributed to the fee pool. This behaviour also matches step 5) described in the spec [here](https://spec.interlay.io/spec/issue.html).

Soo... off the top of my hat I also don't see where we would collect fees in native. But we need to think carefully about that of course.

But on the other hand I don't see a problem if we mix the fee + reward handling for the native currency. We can always just try to exhaust the native balance that the fee pallet has of the native currency (if any, maybe there will never be any, we need to check this) and then mint the rest. The receiver of the reward does not need to care if they come from a fee pool or are minted here. The overall reward stays the same anyways.

Or am I missing something?

gianfra-t: No there is not a problem at all mixing, and I have to agree that the good thing is that only requires changes on the new pallet. So I will add logic that to handle cases we are perhaps overlooking.
On another related topic

pallets/reward-distribution/src/lib.rs
```diff
@@ -182,22 +183,44 @@ pub mod pallet {
...
+	) -> DispatchResult {
+		let native_currency_id = T::GetNativeCurrencyId::get();
+		if reward_currency_id == native_currency_id {
+			let available_native_funds = T::Balances::total_balance(&Self::fee_pool_account_id());
+
+			if available_native_funds < reward {
+				//if available native is less than reward we transfer first from fee account,
+				//and mint the rest
+				let remaining =
+					reward.checked_sub(&available_native_funds).ok_or(Error::<T>::Underflow)?;
```
ebma, line 206: We could maybe simplify this with
```
// Check if we can pay the whole reward in native currency from the fee pool
			match reward.checked_sub(&available_native_funds) {
				None => {
					// Pay the whole reward from the fee pool
					let amount: currency::Amount<T> = Amount::new(reward, reward_currency_id);
					return amount.transfer(&Self::fee_pool_account_id(), &beneficiary)
				},
				Some(remaining) => {
					// Use the available funds from the fee pool
					let available_amount: currency::Amount<T> =
						Amount::new(available_native_funds, reward_currency_id);
					available_amount.transfer(&Self::fee_pool_account_id(), &beneficiary)?;
					// Mint the rest
					T::Balances::deposit_creating(&beneficiary, remaining);
				},
			}
			Ok(())
```

pallets/reward-distribution/src/lib.rs
```diff
@@ -182,22 +183,44 @@ pub mod pallet {
...
 			} else {
-				let amount: currency::Amount<T> = Amount::new(rewards, reward_currency_id);
-				amount.transfer(&Self::fee_pool_account_id(), &caller)?;
-				return Ok(())
+				let amount: currency::Amount<T> = Amount::new(reward, reward_currency_id);
+				return amount.transfer(&Self::fee_pool_account_id(), &beneficiary)
 			}
+		} else {
+			//we need no checking of available funds, since fee will ALWAYS have enough collected
+			//fees
```
ebma, line 218:
```suggestion
			//we need no checking of available funds, since the fee pool will ALWAYS have enough collected
			//fees of the wrapped currencies

```

pallets/reward-distribution/src/lib.rs
gianfra-t: Nice!

pallets/vault-registry/src/types.rs
```diff
@@ -346,15 +346,6 @@ impl<T: Config> RichVault<T> {
 		self.decrease_issued(tokens)
 	}
 
-	fn update_stake(&self) -> DispatchResult {
```
TorstenStueber, line 349: Why has this logic been removed?

TorstenStueber (changes requested):

pallets/reward-distribution/src/lib.rs
```diff
@@ -99,5 +156,188 @@ pub mod pallet {
...
+				error_reward_accum =
+					error_reward_accum.checked_add(&reward_for_pool).ok_or(Error::<T>::Overflow)?;
+			}
+		}
+
+		Ok(error_reward_accum)
+	}
+
+	fn withdraw_all_rewards_from_vault(vault_id: DefaultVaultId<T>) -> DispatchResult {
+		let mut all_reward_currencies = ext::staking::get_all_reward_currencies::<T>()?;
```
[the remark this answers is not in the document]
gianfra-t: In the current model, fee rewards are payed in wrapped currencies to every vault, not just the native and wrapped. This is also relevant for the [previous question](https://github.com/pendulum-chain/spacewalk/pull/415#discussion_r1368532569), and is the reason why we could just not use the pallet "as is" without that modification.

pallets/reward-distribution/src/lib.rs
```diff
@@ -99,5 +156,188 @@ pub mod pallet {
...
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	fn transfer_reward(
+		reward_currency_id: CurrencyId<T>,
+		reward: BalanceOf<T>,
+		beneficiary: T::AccountId,
+	) -> DispatchResult {
+		let native_currency_id = T::GetNativeCurrencyId::get();
```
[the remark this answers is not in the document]
gianfra-t: Here the rewards will be minted if the fee account does not have enough to pay. We thought that it may be useful since this will also allow for the possibility of charging fees in native currency (something right now is not done) but this will allow it to do so, albeit with the downside of mixing the two systems.

pallets/pooled-rewards/src/lib.rs
```diff
@@ -0,0 +1,592 @@
...
+
+impl<T, I, Balance> RewardsApi<T::PoolId, T::StakeId, Balance, T::PoolRewardsCurrencyId>
+	for Pallet<T, I>
+where
+	T: Config<I>,
+	I: 'static,
+	Balance: BalanceToFixedPoint<SignedFixedPoint<T, I>> + Saturating + PartialOrd + Copy,
+	<T::SignedFixedPoint as FixedPointNumber>::Inner: TryInto<Balance>,
+{
+	//type PoolRewardsCurrencyId = CurrencyId;
```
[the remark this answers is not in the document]
gianfra-t: Just a matter of preference. I will change it back.

pallets/reward-distribution/src/lib.rs
```diff
@@ -99,5 +156,188 @@ pub mod pallet {
...
+		//update the reward per block if decay interval passed
+		if let Ok(expired) = ext::security::parachain_block_expired::<T>(
+			rewards_adapted_at,
+			T::DecayInterval::get() - T::BlockNumber::one(),
+		) {
+			if expired {
+				let decay_rate = T::DecayRate::get();
+				reward_this_block = (Perquintill::one() - decay_rate).mul_floor(reward_per_block);
+				RewardPerBlock::<T>::set(Some(reward_this_block));
+				RewardsAdaptedAt::<T>::set(Some(height));
```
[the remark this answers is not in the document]
gianfra-t: Yes, I agree.

pallets/reward-distribution/src/lib.rs
```diff
@@ -99,5 +156,188 @@ pub mod pallet {
...
+
+		let mut total_stake_in_usd = BalanceOf::<T>::default();
+		for (currency_id, stake) in total_stakes.clone().into_iter() {
+			let stake_in_usd = T::OracleApi::currency_to_usd(&stake, &currency_id)?;
+			total_stake_in_usd = total_stake_in_usd.checked_add(&stake_in_usd).unwrap();
+		}
+		//distribute the rewards to each collateral pool
+		let mut error_reward_accum = BalanceOf::<T>::zero();
+		for (currency_id, stake) in total_stakes.into_iter() {
+			let stake_in_usd = T::OracleApi::currency_to_usd(&stake, &currency_id)?;
```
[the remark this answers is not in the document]
gianfra-t: I think I understand, we can save into an intermediate vector of `stake_in_usd` to avoid recalculating the price in USD in the first line of the second iteration.

pallets/fee/src/lib.rs
```diff
@@ -220,35 +235,13 @@ pub mod pallet {
...
-		}
-
 		/// Changes the issue fee percentage (only executable by the Root account)
 		///
 		/// # Arguments
 		///
 		/// * `origin` - signing account
 		/// * `fee` - the new fee
-		#[pallet::call_index(1)]
+		#[pallet::call_index(0)]
```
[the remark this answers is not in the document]
gianfra-t: I did not know. Will change it back.

pallets/reward-distribution/src/lib.rs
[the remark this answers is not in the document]
TorstenStueber: Exactly!

pallets/staking/src/lib.rs
```diff
@@ -269,6 +273,11 @@ pub mod pallet {
 	pub type Nonce<T: Config> =
 		StorageMap<_, Blake2_128Concat, DefaultVaultId<T>, T::Index, ValueQuery>;
 
+	// store with all the reward currencies in use
+	#[pallet::storage]
+	pub type RewardCurrencies<T: Config> =
```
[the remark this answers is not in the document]
TorstenStueber: What I mean is: doesn't the `RewardCurrencies` in the pooled-rewards pallet already keep track of this?

pallets/reward-distribution/src/lib.rs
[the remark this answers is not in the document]
TorstenStueber: Yes, as discussed on [Slack](https://satoshipay.slack.com/archives/C01V1F56RMJ/p1698072290950939?thread_ts=1698066483.882979&cid=C01V1F56RMJ).

pallets/reward-distribution/src/lib.rs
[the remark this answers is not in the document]
TorstenStueber: I missed this part in the code. It probably works correctly but this is quite a big (implicit) assumption that we can do this – the assumption being that any native fee not in the fee account still needs to be minted.

This feels brittle to me, at least when this assumption is so implicitly hidden here at this code. It could also lead to a particularly serious attack if there is a bug somewhere else in the code and there are way more fees minted than intended.

I think it is cleaner and safer to keep track of native fees that need to be minted in an extra storage field. Thoughts?

pallets/vault-registry/src/types.rs
TorstenStueber: Without looking at this now in detail, does this mean that we changed the logic of the fee rewards and how they are distributed?

pallets/staking/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: Yes it does, it has it by pool id, but it does. It was the other alternative in fact. The rationale behind this change was to have more control of the currencies that are going to be distributed. For example, in case in the future we wanted to change/remove some of them, and also the fact that this new store does not differentiate between pools.

pallets/vault-registry/src/types.rs
gianfra-t: Yes, the fee and native currencies are now distributed based on the amount of collateral in USD, not by the amount of tokens they have bridged. 
As an aside note, this is more aligned to what [interlay does](https://github.com/interlay/interbtc/blob/90a5c1b707c00dc9035c7ab438a7db3c8e651c19/crates/vault-registry/src/pool_manager.rs#L63) at the moment. With the difference we do not divide by the threshold.

pallets/reward-distribution/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: This check is now done [here](https://github.com/pendulum-chain/spacewalk/pull/415/commits/05b2bed97777144de080afcc1c3e5196feff911b#diff-2f743609293dfd3b50158de75d67dabc3018b7fa2abe78bed8fd45a783aa3975R218)

pallets/vault-registry/src/types.rs
TorstenStueber: This change is fine as discussed yesterday.

pallets/fee/src/mock.rs
```diff
@@ -221,6 +227,39 @@ parameter_types! {
...
+pub struct OracleApiMock {}
+impl oracle::OracleApi<Balance, CurrencyId> for OracleApiMock {
+	fn currency_to_usd(
+		_amount: &Balance,
+		currency_id: &CurrencyId,
+	) -> Result<Balance, DispatchError> {
+		let _native_currency = GetNativeCurrencyId::get();
+		match currency_id {
+			_native_currency => return Ok(100),
+			//_ => unimplemented!("unimplemented mock conversion for currency"),
```
ebma, line 244: Let's uncomment this line?

pallets/issue/src/mock.rs
```diff
@@ -310,6 +305,53 @@ impl fee::Config for Test {
...
+pub struct OracleApiMock {}
+impl oracle::OracleApi<Balance, CurrencyId> for OracleApiMock {
+	fn currency_to_usd(
+		_amount: &Balance,
+		currency_id: &CurrencyId,
+	) -> Result<Balance, DispatchError> {
+		let _native_currency = GetNativeCurrencyId::get();
+		match currency_id {
+			_native_currency => return Ok(100),
+			//_ => unimplemented!("unimplemented mock conversion for currency"),
```
ebma, line 338: Same here. Let's do it in all mocks.

pallets/oracle/src/lib.rs
```diff
@@ -367,3 +367,19 @@ impl<T: Config> Pallet<T> {
...
+}
+
+impl<T: Config> OracleApi<BalanceOf<T>, CurrencyId> for Pallet<T> {
+	fn currency_to_usd(
+		amount: &BalanceOf<T>,
+		currency_id: &CurrencyId,
+	) -> Result<BalanceOf<T>, DispatchError> {
+		Pallet::<T>::currency_to_usd(amount.clone(), currency_id.clone())
+	}
+}
```
ebma, line 385: Should we put this definition more to the top of the file or maybe even an extra file? It's a bit hidden down here

pallets/pooled-rewards/src/mock.rs
```diff
@@ -0,0 +1,130 @@
...
+);
+pub const DEFAULT_WRAPPED_CURRENCY3: CurrencyId = CurrencyId::AlphaNum4(
+	*b"MXN\0",
+	[
+		20, 209, 150, 49, 176, 55, 23, 217, 171, 154, 54, 110, 16, 50, 30, 226, 102, 231, 46, 199,
+		108, 171, 97, 144, 240, 161, 51, 109, 72, 34, 159, 139,
+	],
+);
+
+pub const DEFAULT_WRAPPED_CURRENCY4: CurrencyId = CurrencyId::AlphaNum4(
```
ebma, line 58: Nit: Can we change the names to be `...CURRENCY_{INDEX}`?

ebma: Why doesn't it start at 1 though?

pallets/reward-distribution/src/lib.rs
```diff
@@ -70,7 +112,25 @@ pub mod pallet {
 	}
 
 	#[pallet::error]
-	pub enum Error<T> {}
+	pub enum Error<T> {
+		//Overflow
+		Overflow,
+		//underflow
+		Underflow,
+		//if origin tries to withdraw with 0 rewards
```
ebma, line 120: Let's replace these with proper doc comments (`///`) because these will be included in the generated metadata.

pallets/reward-distribution/src/lib.rs
```diff
@@ -70,7 +112,25 @@ pub mod pallet {
...
+		NoRewardsForAccount,
+		//error if amount to be minted is more than rewarded in total
+		NotEnoughRewardsRegistered,
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
+		fn on_initialize(n: T::BlockNumber) -> Weight {
+			Self::execute_on_init(n);
+			//TODO benchmark this weight properly
```
ebma, line 130: You already added some benchmark. Are you not happy with it or is this a stale comment @gianfra-t?

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +159,212 @@ pub mod pallet {
...
+			}
+		} else {
+			log::warn!("Failed to check if the parachain block expired");
+		}
+		//keep track of total native currencies to be minted
+		NativeLiability::<T>::mutate(|current_liability| match current_liability {
+			Some(current_liability) => *current_liability += reward_this_block,
+			None => *current_liability = Some(reward_this_block),
+		});
+		//TODO how to handle error if on init cannot fail?
```
ebma, line 288: Maybe we should 
```
if let Err(err) = Self::distribute_rewards(...)
```
and try to log the error message? Not much we can do otherwise.

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +159,212 @@ pub mod pallet {
...
+		let _ = Self::distribute_rewards(reward_this_block, T::GetNativeCurrencyId::get());
+	}
+
+	//fetch total stake (all), and calulate total usd stake in percentage across pools
```
ebma, line 292:
```suggestion
	//fetch total stake (of each pool), and calculate total USD stake in percentage across pools
	//distribute the reward accordingly
```

pallets/staking/src/lib.rs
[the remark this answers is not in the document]
TorstenStueber: Okay, I did not look properly and only realize now that here the currencies are not accounted by pool id but in general, so that it is indeed a difference.

So then it makes sense.

pallets/pooled-rewards/src/mock.rs
gianfra-t: I think there is already a DEFAULT_WRAPPED_CURRENCY defined elsewhere (which would be 1), but I will update the indexes to `_x`

pallets/reward-distribution/src/lib.rs
gianfra-t: It is a leftover comment. The only "concern" is that the generated file outputted a [strange](https://github.com/pendulum-chain/spacewalk/pull/415/files#diff-d012ec42d9339a0c96c67bf1afef24aed64425b409839559c22b7cf9545417e5R118)  set of commented lines which I have not seen before. Otherwise it should be okay.

pallets/reward-distribution/src/lib.rs
gianfra-t: With a trace. Yes, I will do that!

pallets/pooled-rewards/src/mock.rs
gianfra-t: I will move them all up to testing constants in currency so they are all in the same place.

ebma (commented):
The latest implementation looks good to me overall. I mainly added nit-picking comments.

It's probably not obvious but doc comments on specific pallet items are super useful because they will be extracted into the metadata and tools like polkadot.js use them to show some extra info when eg calling extrinsics.

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +161,213 @@ pub mod pallet {
...
-
+			RewardsAdaptedAt::<T>::put(ext::security::get_active_block::<T>());
 			Self::deposit_event(Event::<T>::RewardPerBlockAdapted(new_reward_per_block));
 			Ok(())
 		}
+
+		#[pallet::call_index(1)]
+		#[pallet::weight(<T as Config>::WeightInfo::collect_reward())]
+		#[transactional]
+		pub fn collect_reward(
```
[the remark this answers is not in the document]
gianfra-t: Sorry, I wasn't aware really paying much attention to the comments. Will have in mind for any future runtime work.

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +161,213 @@ pub mod pallet {
...
+				return
+			},
+		};
+
+		let mut reward_this_block = reward_per_block;
+
+		//update the reward per block if decay interval passed
+		if let Ok(expired) = ext::security::parachain_block_expired::<T>(
+			rewards_adapted_at,
+			T::DecayInterval::get() - T::BlockNumber::one(),
```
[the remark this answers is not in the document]
gianfra-t: I don't think it can ever be 0 or even one (in terms of making sens), but okay, a saturating sub is safer.

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +161,213 @@ pub mod pallet {
...
+			let stake_in_usd = T::OracleApi::currency_to_usd(&stake, &currency_id)?;
+			stakes_in_usd.push(stake_in_usd);
+			total_stake_in_usd = total_stake_in_usd.checked_add(&stake_in_usd).unwrap();
+		}
+		//distribute the rewards to each collateral pool
+		let mut error_reward_accum = BalanceOf::<T>::zero();
+		let mut stakes_in_usd_iter = stakes_in_usd.into_iter();
+		for (currency_id, _stake) in total_stakes.into_iter() {
+			let stake_in_usd =
+				stakes_in_usd_iter.next().expect("cannot be less than previous iteration");
```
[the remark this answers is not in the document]
gianfra-t: Agreed

pallets/reward-distribution/src/tests.rs
```diff
@@ -33,3 +45,132 @@ fn test_set_rewards_per_block() {
...
+#[test]
+fn distributes_rewards_properly() {
+	run_test(|| {
+		ext::pooled_rewards::get_total_stake_all_pools::<Test>.mock_safe(move || {
+			let initial_stakes = build_total_stakes();
+			MockResult::Return(Ok(initial_stakes))
+		});
+
+		let mut expected_pool_ids = vec![XCM(0), XCM(1), XCM(2), XCM(3)].into_iter();
+		let mut expected_stake_per_pool = vec![46, 13, 34, 4].into_iter();
```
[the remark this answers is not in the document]
gianfra-t: They were externally calculated, and are consistent with he model, departing from the fact that 100 is the reward. I will make the calculation explicit.

pallets/reward-distribution/src/tests.rs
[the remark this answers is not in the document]
gianfra-t: We do have it [this](https://github.com/pendulum-chain/spacewalk/pull/428) branch. But it is true it is not explicitly testing that. So I can add those one in this pallet also.

pallets/reward-distribution/src/tests.rs
```diff
@@ -33,3 +45,132 @@ fn test_set_rewards_per_block() {
...
+			.mock_safe(move |_, _| MockResult::Return(Ok(true)));
+
+		let new_rewards_per_block = 10000;
+		assert_ok!(RewardDistribution::set_reward_per_block(
+			RuntimeOrigin::root(),
+			new_rewards_per_block
+		));
+
+		System::set_block_number(200);
+		//Security::<Test>::set_active_block_number(200);
```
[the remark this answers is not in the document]
gianfra-t: It is confusing especially since the test was made, there has been changes where now the active block is used for `RewardsAdapted`. I will mock `ext::security::get_active_block`. 
Also, I believe set_block_number is also setting the active block? Otherwise this test is not working properly.

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +161,213 @@ pub mod pallet {
...
+			rewards_adapted_at,
+			T::DecayInterval::get() - T::BlockNumber::one(),
+		) {
+			if expired {
+				let decay_rate = T::DecayRate::get();
+				reward_this_block = (Perquintill::one() - decay_rate).mul_floor(reward_per_block);
+				RewardPerBlock::<T>::set(Some(reward_this_block));
+				let active_block = ext::security::get_active_block::<T>();
+				RewardsAdaptedAt::<T>::set(Some(active_block));
+			}
```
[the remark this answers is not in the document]
gianfra-t: The only problem I see with the new code is that we will not differentiate between an Err(_) an Ok(false), unless we remove the log bellow then it won't be a problem.

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +161,213 @@ pub mod pallet {
...
+			None => *current_liability = Some(reward_this_block),
+		});
+
+		if let Err(_) = Self::distribute_rewards(reward_this_block, T::GetNativeCurrencyId::get()) {
+			log::warn!("Rewards distribution failed");
+		}
+	}
+
+	//fetch total stake (all), and calulate total usd stake in percentage across pools
+	//distribute the reward accoridigly
```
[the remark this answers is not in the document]
gianfra-t: I really need a grammar checker for vs code!

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +169,229 @@ pub mod pallet {
...
+
+		let decay_interval_sub = match T::DecayInterval::get().checked_sub(&T::BlockNumber::one()) {
+			Some(value) => value,
+			None => {
+				log::warn!("Decay interval underflow");
+				return
+			},
+		};
+
+		if let Ok(expired) =
```
gianfra-t, line 289: With the usage of `checked_sub` this ends up being even more convoluted. Taking into account [that differentiation](https://github.com/pendulum-chain/spacewalk/pull/415#discussion_r1373245872) which I think we should keep let me know @ebma if you can think of a way to simplify this.

pallets/reward-distribution/src/lib.rs
```diff
@@ -94,10 +169,229 @@ pub mod pallet {
...
+						return Err(Error::<T>::NotEnoughRewardsRegistered.into())
+					}
+
+					// Use the available funds from the fee pool
+					let available_amount: currency::Amount<T> =
+						Amount::new(available_native_funds, reward_currency_id);
+					available_amount.transfer(&Self::fee_pool_account_id(), &beneficiary)?;
+					// Mint the rest
+					T::Balances::deposit_creating(&beneficiary, remaining);
+
```
gianfra-t, line 242: There is a potential bug where the reward is below the existential deposit which I have encountered in testing. 
I say potential given that it depends on how one defines the rewards per block. If they are lower than the existential deposit, a user could attempt to collect native bellow the existential value when it's account does not exists, which will result in burning those rewards essentially.

Since this is a very edge case and the logic is already quite big, do you think we should add a check for this @ebma ?

ebma: Hmm okay, nice catch.

> If they are lower than the existential deposit, a user could attempt to collect native bellow the existential value when it's account does not exists, which will result in burning those rewards essentially.

But I think a user who does not have enough native tokens to fulfill the existential deposit also would not have enough native tokens to issue the transaction in the first place, or would he?

ebma: This would be interesting for non-native tokens as well though. IIRC orml-tokens also defines an existential deposit in its config. What would happen if a nominator tries to withdraw some very small amount that is below the exitential deposit of orml-tokens? Something like 100 units should not work probably.

pallets/reward-distribution/src/lib.rs
ebma: Let's just use a [saturating_sub](https://snowbridge-rust-docs.snowfork.com/sp_runtime/traits/trait.saturating#tymethod.saturating_sub) then as this does not throw an error, thus we do not need to add a complex match expression.

ebma: I think you could just do 
```
if Ok(true) == ext::security::parachain_block_expired::<T>(
			rewards_adapted_at,
			T::DecayInterval::get().saturating_sub(T::BlockNumber::one()),
		) {
			let decay_rate = T::DecayRate::get();
			...
...
```

pallets/reward-distribution/src/lib.rs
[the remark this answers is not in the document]
ebma: Do we need to differentiate between them? I mean.. you have a point actually that this might not be ideal in all cases. But looking at the [parachain_block_expired](https://github.com/pendulum-chain/spacewalk/blob/a43141b920b0292ad5ec90f5c5df5729bc92b5ba/pallets/security/src/lib.rs#L256) function, the only problem we could have is an Overflow if we use really large values so maybe it's okay if we miss this case. What do you think? If you think it's critical, we can also try to keep it such that we keep the log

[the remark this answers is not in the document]
gianfra-t: Let's go with not differentiate them. I hadn't looked at the function but as you say the error is not that interesting.

pallets/reward-distribution/src/lib.rs
gianfra-t: The check is done before the withdraw, [here](https://github.com/pendulum-chain/spacewalk/pull/415/files#diff-2f743609293dfd3b50158de75d67dabc3018b7fa2abe78bed8fd45a783aa3975R202)

ebma: Nice, great stuff 👍

ebma:
For some reason, the [checks](https://github.com/pendulum-chain/spacewalk/pull/415/checks) are disabled for this PR. I don't see an option to turn them back on nor do I see a reason why they don't run. Maybe it's because the PR has conflicts with the base branch that must be resolved? @gianfra-t can you resolve these conflicts so we see if this prevents the CI actions from running? 
Otherwise, please run all the relevant checks contained [here](https://github.com/pendulum-chain/spacewalk/blob/404-add-pooled-reward-pallet/.github/workflows/ci-main.yml) locally so we know if there are any issues.

gianfra-t:
Hey @ebma, does this error:

ebma:
Yes exactly, regenerating the metadata should do the trick 👍

gianfra-t:
I just merged with the latest changes in main, but the tests are passing. As soon as they pass again should we merge this? Or do you want to have another look @TorstenStueber?

pallets/reward-distribution/src/lib.rs
```diff
@@ -193,6 +199,21 @@ pub mod pallet {
...
 			ext::staking::distribute_reward::<T>(&vault_id, reward, reward_currency_id)?;
 
+			// We check if the amount to transfer is greater than of the existential deposit to
+			// avoid potential losses of reward currency, in case currency is not native
+			let minimum_transfer_amount =
+				<T as orml_tokens::Config>::ExistentialDeposits::get(&reward_currency_id);
+			let expected_rewards =
+				ext::staking::compute_reward::<T>(&vault_id, &caller, reward_currency_id)?;
+
+			if expected_rewards == (BalanceOf::<T>::zero()) {
```
ebma, line 209: Nit: We don't need the brackets do we (?)
```suggestion
			if expected_rewards == BalanceOf::<T>::zero() {
```

ebma (approved):
Looks good to me now. Thanks for your great efforts with all of this @gianfra-t! 👍 

Let's still wait for the review of @TorstenStueber before merging though

pallets/reward-distribution/src/lib.rs
gianfra-t: Probably not :)

TorstenStueber (approved):
I looked thoroughly at the code again and everything looks good.

Great job!

ebma:
@gianfra-t feel free to push the squash and merge button then.

PR #418, opened 2023-10-12
411 Allow calling extrinsics of `clients-info crate` from authorized accounts

### LINKED ISSUE #411: Allow calling extrinsics of `clients-info` crate from authorized accounts
We should change the `clients-info` crate to allow calling the `set...ClientRelease` extrinsics from root or an authorized account. This way, we don't have to create an extra referendum each time we want to update our Spacewalk clients but can rely on using a multi-sig account instead. 

### Notes
For implementing the logic for the authorized accounts, you can draw inspiration from the [implementation](https://github.com/pendulum-chain/oracle-pallet/blob/master/pallets/dia-oracle/src/lib.rs) of the DIA oracle pallet.

b-yap:
closes #411 

* Authorize other accounts to perform the extrinsics of `clients-info`
* Introduce the `AuthorizedAccounts` storage
* generate new weights

## How to begin the review:
* pallets/clients-info/src/
   * lib.rs
     * new storage: `AuthorizedAccounts`
     * new extrinsics: `fn authorize_account()` and `fn deauthorize_account()`
     * old extrinsics: `if the origin is not _`root`_, check from the `AuthorizedAccounts`
     * new errors: 
        * `ThisAccountIdIsNotAuthorized` 
        * `UserUnableToDeauthorizeThemself`
        * `UserUnableToDeauthorizeRootAccount`
     * new events: `AccountIdAuthorized(T::AccountId)`, `AccountIdDeauthorized(T::AccountId)`
   * tests.rs: new tests;
     * ` fn authorize_account_should_work()`
     * `fn deauthorize_account_should_work()`
     * `fn test_errors()`
   * benchmarking.rs: added`fn deauthorize_account()` and `fn authorize_account()`


pallets/clients-info/src/lib.rs
```diff
@@ -108,22 +129,73 @@ pub mod pallet {
...
+
+		#[pallet::call_index(3)]
+		#[pallet::weight(<T as Config>::WeightInfo::deauthorize_account())]
+		#[transactional]
+		pub fn deauthorize_account(
+			origin: OriginFor<T>,
+			account_id: T::AccountId,
+		) -> DispatchResult {
+			if let Ok(origin_account_id) = ensure_signed(origin.clone()) {
+				Pallet::<T>::check_non_root_rights(origin.clone())?;
```
ebma, line 160: How about we change the return type of `check_non_root_rights()` to return the `AccountId` each time? Then we can get rid of the `if let` in the previous line.

ebma: Also, with the currnt logic, anyone can call this extrinsic unsigned and we will skip this if statement completely and go to the next if statement where we remove. This is a flaw that we would fix with my suggested change.

pallets/clients-info/src/lib.rs
```diff
@@ -108,22 +129,73 @@ pub mod pallet {
...
+					Error::<T>::UserUnableToDeauthorizeThemself
+				);
+
+				// do not deauthorize the root account
+				let account_id_to_origin =
+					OriginFor::<T>::from(RawOrigin::Signed(account_id.clone()));
+				ensure!(
+					ensure_root(account_id_to_origin).is_err(),
+					Error::<T>::UserUnableToDeauthorizeRootAccount
+				);
```
ebma, line 172: I think we don't really have to handle this case since the implementation of `check_origin_rights` does not even care if `root` is part of the list of the `AuthorizedAccounts`. So, while it does not make sense to add `root` to the list of authorized accounts (since it can call all functions anyways), it is not a problem if the user of this pallet adds it to the authorized accounts.
Or am I seeing this wrong @b-yap?

pallets/clients-info/src/lib.rs
```diff
@@ -72,6 +73,26 @@ pub mod pallet {
...
+		}
+
+		fn check_origin_rights(origin: OriginFor<T>) -> DispatchResult {
+			if ensure_root(origin.clone()).is_err() {
+				let _ = Pallet::<T>::check_non_root_rights(origin)?;
+			};
+
+			Ok(())
+		}
+	}
```
ebma, line 94: Can we move this code below the code block for extrinsics? It does not matter really, but I would expect to first read the 'important' functions ie. extrinsics, before reading some helper functions. Or would you prefer to keep it like this @b-yap?

ebma (changes requested):
I left some comments, tests and benchmarks look good :+1:

pallets/clients-info/src/lib.rs
b-yap: `check_non_root_rights()` already returns the account id.

pallets/clients-info/src/lib.rs
b-yap: case removed

ebma (approved):
LGTM now

pallets/clients-info/src/lib.rs
ebma: > check_non_root_rights() already returns the account id.

True, sorry. I actually was referring to `check_origin_rights()`.

pallets/clients-info/src/lib.rs
```diff
@@ -1,7 +1,6 @@
 //! # ClientsInfo Module
 //! Stores information about clients used for the network.
 
-#![deny(warnings)]
```
ebma, line 4: Don't we want to keep this line though?

b-yap: :open_mouth:

PR #424, opened 2023-10-17
421 fix failing runner test

### LINKED ISSUE #421: Fix failing `runner` test
The [test_runner_download_binary](https://github.com/pendulum-chain/spacewalk/blob/468ec562bcee16257335b0ba488c4024983af1c5/clients/runner/src/runner.rs#L709) test does not work anymore because it relies on a hard-coded checksum that does not match the checksum of the latest compiled `vault-rococo` binary anymore. We should make this test more robust against changes to the uploaded binaries.

I would suggest that to bypass the `IncorrectChecksum` error, we could extract [the check comparing the checksums](https://github.com/pendulum-chain/spacewalk/blob/468ec562bcee16257335b0ba488c4024983af1c5/clients/runner/src/runner.rs#L205-L208) to an extra function `checksums_match(first, second) -> boolean` which we then mock in the test to always return `true`.

b-yap:
closes #421 .

### DIFF
--- a/clients/runner/src/runner.rs
+++ b/clients/runner/src/runner.rs
@@ -147,7 +147,7 @@ impl Runner {
 	) -> Result<DownloadedRelease, Error> {
 		//recheck checksum to see if binary is the one we have
 		if let Some(downloaded_release) = runner.downloaded_release() {
-			if downloaded_release.checksum.eq(&release.checksum) {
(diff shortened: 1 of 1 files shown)


gianfra-t (approved):
Looks okay to me!

Issue #423, opened 2023-10-16
Integration tests for new vault rewards implementation 


gianfra-t:
Following the completion of tasks in [this pendulum issue](https://github.com/pendulum-chain/tasks/issues/47), it would be nice to add some integration tests that ensure that the rewards are being distributed according to the model.

We need to test these main things:
- That the rewards per block are properly distributed based on total collateral staked in usd.
- That the fees are also properly distributed based on collateral staked in usd.
- That the nominators/vault operators get the corresponding value based on their vault's stakes and can properly withdraw.

For this, we need to ensure in the set-up of the tests that:
- There exists multiple vaults, with multiple collateral and wrapped currencies.
- That this vaults are both backed in collateral by their operators and nominators.

Following the setup, we must ensure that this vaults issue some amount of tokens to test the fee distribution, and that some blocks move forward to test the per block distribution. This two tests can be independent.

Finally, we must ensure that the rewards distributed follows the model:
 reward `vault_i` = ( stake_usd(`collateral_i`)/Total_stake_usd ) * `reward`.

Where `stake_usd()` is the total sum in usd of collateral across vaults_id that share that collateral and `Total_stake_usd` is the sum of `stake_usd` across all collateral currencies in the system.

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/423?planning-poker) with Zenhub @b-yap @ebma @gianfra-t @TorstenStueber

PR #428, opened 2023-10-20
Rewards pallet-integration tests

gianfra-t:
Regarding #423.

As discussed in the issue, we need to ensure the proper issuance and distribution of the fee and per block reward tokens.

This is tested as follows:
- In the `vault-registry` pallet, used due to the fact that it already has functions, constants and genesis configs appropriate to deploy vaults, we test here the block reward distribution. This is done in [these](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/vault-registry/src/tests.rs#L1717) `integration_*` test functions. These tests cover the registry of a vault, execution of the [on init](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/vault-registry/src/tests.rs#L1839) function (called directly) and assertion that the vault owners can [collect](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/vault-registry/src/tests.rs#L1842) the rewards.

- In the `issue` pallet, given that it has the proper mocks for issuing and triggering the fee distribution, we test in [these](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/issue/src/tests.rs#L627) `integration` module tests, we ensure that the fees are distributed to the many collateral registered vaults, and that all the vaults in the `pooled-reward` pallet [receive](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/issue/src/tests.rs#L886) these rewards accordingly. (also have in mind the [price](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/issue/src/mock.rs#L324) mock for this example distribution).
On the same module, we also test that [nominated vault](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/issue/src/tests.rs#L935) is taken into account and that the nominator gets the [proper rewards](https://github.com/pendulum-chain/spacewalk/blob/386ba48e1404972fa1a7aa1ef32bee6d027bfdac/pallets/issue/src/tests.rs#L952) and can collect them.


ebma:
@gianfra-t let's make this PR target the `404-add-pooled-reward-pallet` because I assume you based these changes on it. Otherwise it's hard to see what changes are added in this PR

gianfra-t:
Sure!

pallets/issue/src/tests.rs
```diff
@@ -624,3 +624,338 @@ fn test_request_issue_reset_interval_and_succeeds_with_rate_limit() {
...
+			assert_eq!(get_reward_for_vault(&VAULT3, issue_asset2), 304);
+			assert_eq!(get_reward_for_vault(&VAULT4, issue_asset2), 345);
+			assert_eq!(get_reward_for_vault(&VAULT5, issue_asset2), 243);
+
+			//assertions for issue asset3
+			assert_eq!(get_reward_for_vault(&VAULT, issue_asset3), 50);
+			assert_eq!(get_reward_for_vault(&VAULT2, issue_asset3), 101);
+			assert_eq!(get_reward_for_vault(&VAULT3, issue_asset3), 152);
+			assert_eq!(get_reward_for_vault(&VAULT4, issue_asset3), 172);
+			assert_eq!(get_reward_for_vault(&VAULT5, issue_asset3), 122);
```
ebma, line 906: The values for `VAULT3`, `VAULT4` and `VAULT5` don't make sense considering their collaterals defined in the preceding lines, do they? How can `VAULT3` always have more reward than `VAULT2` if it provides less collateral?

ebma (commented):
Even though it's just a draft, I already had a look because I was curious how you implemented the integration test. 

Some remarks on the code style. These are my personal preference and we can also discuss with the other techies what we want to use:
- I would prefer if we put the number suffix with another `_`. So `vault_2` instead of `vault2`. 
- If we have several similar things, then they should be `vault_1` and `vault_2` instead of `vault` and `vault_2`. 
- `expect()` is preferable over `unwrap()` also in tests so that it's easier to see what failed
- I would prefer if we explicitly calculate the expected numbers instead of using magic numbers or explaining them in comments. This way, we can easily change the used variables and don't have to touch the expected results each time.

pallets/issue/src/tests.rs
gianfra-t: That is because since the fee in the model are distributed also given the value of the pool in USD. In the mock for this tests defined [here](https://github.com/pendulum-chain/spacewalk/pull/428/files#diff-d6cdd749af964b12aacc03febe5032c67c92157c7f9864bdaec757186d686165R272), the value of collateral for 3,4 and 5 is higher in USD.

pallets/vault-registry/src/tests.rs
```diff
@@ -1711,3 +1712,146 @@ fn test_offchain_worker_unsigned_transaction_submission() {
...
+		//collect rewards
+		let origin = RuntimeOrigin::signed(id.account_id);
+		assert_ok!(<reward_distribution::Pallet<Test>>::collect_reward(
+			origin.into(),
+			id.clone(),
+			DEFAULT_NATIVE_CURRENCY,
+			None,
+		));
+
+		//expected value  = (1000*5/(5000*5 + 1000*5))*100000 = 16666
```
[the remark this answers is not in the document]
gianfra-t: That is to get explicit the price in USD, ([value in usd is defined here](https://github.com/pendulum-chain/spacewalk/pull/428/files#diff-d6cdd749af964b12aacc03febe5032c67c92157c7f9864bdaec757186d686165R273). It is confusing that the issue is also 5, I will change that so there is no confusion since as you say, it does not matter in this example.

pallets/issue/Cargo.toml
```diff
@@ -35,6 +35,7 @@ stellar-relay = { path = "../stellar-relay", default-features = false }
 vault-registry = { path = "../vault-registry", default-features = false }
 pooled-rewards = { path = "../pooled-rewards", default-features = false }
 reward-distribution = { path = "../reward-distribution", default-features = false }
+nomination = {path = "../nomination", default-features = false}
```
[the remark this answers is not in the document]
gianfra-t: True!

pallets/issue/src/tests.rs
```diff
@@ -624,3 +624,338 @@ fn test_request_issue_reset_interval_and_succeeds_with_rate_limit() {
...
+	}
+	fn setup_execute_with_vault(
+		issue_amount: Balance,
+		issue_fee: Balance,
+		griefing_collateral: Balance,
+		amount_transferred: Balance,
+		vault: DefaultVaultId<Test>,
+	) -> Result<H256, DispatchError> {
+		let vault_clone = vault.clone();
+		let vault_clone_2 = vault.clone();
```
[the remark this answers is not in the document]
gianfra-t: This is something I did not fully understand but was "required" by the compiler. I was going to look into it to improve this since it's not very elegant.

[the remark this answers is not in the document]
gianfra-t: So after exploring a bit more, the first clone is needed since we are going to pass vault to 2 closures, that is expected. The second one inside the closure is because each closure is an FnMut, which means that if it were to take ownership even if it is a clone of vault and is not used any more after that, it would not be available for any further call to the closure. 
Hope it makes sense!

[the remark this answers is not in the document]
ebma: Makes sense, thanks for looking into this :+1:

pallets/issue/src/tests.rs
```diff
@@ -624,3 +624,643 @@ fn test_request_issue_reset_interval_and_succeeds_with_rate_limit() {
...
+			issue_amount_5,
+			issue_fee_5,
+			griefing_collateral_5,
+			amount_transferred_5,
+			VAULT_5.clone(),
+		)
+		.unwrap();
+		assert_ok!(execute_issue(USER, &issue_id_5));
+
+		return (issue_fee_1 + issue_fee_4, issue_fee_2 + issue_fee_5, issue_fee_3)
```
ebma, line 899: This is a little confusing. When looking at the mock.rs file I can see that eg. vaults 1 and 4 use the same wrapped currency and so do vault 2 and 5. But it's not clear here.
Let's either add a comment here or at the top of this function so that a reader knows why these values are summed up and what they represent.

pallets/issue/src/tests.rs
```diff
@@ -624,3 +624,643 @@ fn test_request_issue_reset_interval_and_succeeds_with_rate_limit() {
...
+	}
+
+	#[test]
+	fn integration_multiple_vaults_many_collateral() {
+		run_test(|| {
+			//ARRAGE
```
ebma, line 905:
```suggestion
			//ARRANGE
```

pallets/issue/src/tests.rs
```diff
@@ -624,3 +624,643 @@ fn test_request_issue_reset_interval_and_succeeds_with_rate_limit() {
...
+					total_amount_usd as f64) *
+					fee as f64)
+					.floor() * (collateral_4 as f64 / (collateral_3 + collateral_4) as f64))
+					.floor() as u128
+			};
+
+			let get_expected_value_vault_5 = |fee: Balance| -> Balance {
+				(((vault_5_collateral_usd as f64 / total_amount_usd as f64) * fee as f64).floor())
+					as u128
+			};
```
ebma, line 986: Here it is not clear at first, why vault 1+2, vault 3+4 and vault 5 are grouped together.
Again, looking at the mocks you can see that these vaults share the same collateral currency. But as a reader I wouldn't understand this from just looking at the test. Let's maybe add a comment above the getter functions that explains why some vault collaterals are grouped together.

pallets/vault-registry/src/mock.rs
```diff
@@ -360,6 +369,14 @@ pub const RICH_ID: VaultId<AccountId, CurrencyId> = VaultId {
 		wrapped: DEFAULT_WRAPPED_CURRENCY,
 	},
 };
+pub const ID_COLLATERAL_21: VaultId<AccountId, CurrencyId> = VaultId {
+	account_id: 6,
+	currencies: VaultCurrencyPair { collateral: XCM(1), wrapped: DEFAULT_WRAPPED_CURRENCY },
+};
+pub const ID_COLLATERAL_22: VaultId<AccountId, CurrencyId> = VaultId {
```
ebma, line 376: Let's try to find better names. How about `COLLATERAL_2_VAULT_ID_1` and `COLLATERAL_2_VAULT_ID_2`? Also not perfect but maybe clearer

ebma: We could then also rename `DEFAULT_ID` and `OTHER_ID` to `COLLATERAL_1_VAULT_ID_1` and `COLLATERAL_1_VAULT_ID_2`

pallets/currency/src/testing_constants.rs
```diff
@@ -14,6 +14,22 @@ pub const DEFAULT_WRAPPED_CURRENCY: CurrencyId = CurrencyId::AlphaNum4(
 	],
 );
 
+pub const WRAPPED_CURRENCY2: CurrencyId = CurrencyId::AlphaNum4(
```
ebma, line 17:
```suggestion
pub const WRAPPED_CURRENCY_2: CurrencyId = CurrencyId::AlphaNum4(
```

pallets/currency/src/testing_constants.rs
```diff
@@ -14,6 +14,22 @@ pub const DEFAULT_WRAPPED_CURRENCY: CurrencyId = CurrencyId::AlphaNum4(
...
+	*b"USDT",
+	[
+		20, 209, 150, 49, 176, 55, 23, 217, 171, 154, 54, 110, 16, 50, 30, 226, 102, 231, 46, 199,
+		108, 171, 97, 144, 240, 161, 51, 109, 72, 34, 159, 139,
+	],
+);
+
+pub const WRAPPED_CURRENCY3: CurrencyId = CurrencyId::AlphaNum4(
```
ebma, line 25:
```suggestion
pub const WRAPPED_CURRENCY_3: CurrencyId = CurrencyId::AlphaNum4(
```

ebma (commented):
Thanks for addressing the requested changes @gianfra-t, I really like the tests and the explicit calculations. I added some nitpicky comments again but it looks good to me overall. Great job :+1:

pallets/vault-registry/src/mock.rs
gianfra-t: I agree, looking at it after a few days this is not clear anymore. Is better long but explicit.

ebma (approved):
LGTM. Thanks for addressing all my comments and great job overall @gianfra-t 👍

PR #435, opened 2023-10-25
434 add configuration for automatic changelog generation

### LINKED ISSUE #434: Add configuration for automatic changelog generation
In order to facilitate the creation of new releases, we should use a tool to automatically generate changelogs based on the commit history. 
We should use the [git-chglog](https://github.com/git-chglog/git-chglog) tool because it nicely integrates into GitHub actions, see for example [this](https://github.com/interlay/interbtc-clients/blob/11d6d99c9ce74ab36dfbe7b8efe6f5382f9fe3bc/.github/workflows/cargo-test.yml) workflow.

This is related to #410.

ebma:
This template can be used to create a changelog. Its main purpose is to be used in Spacewalk's release flow to automatically generate a description for the new releases. 

To generate the changelog you need to install git-chglog, see [here](https://github.com/git-chglog/git-chglog?tab=readme-ov-file#installation) and then run `git-chglog -o CHANGELOG.md`.

The CI job will run the generation with the latest tag as the argument so that the changelog is only generated containing the changes for the latest release. For example to generate the release notes for release `v1.0.3`, it will use `git-chglog -o CHANGELOG.md v1.0.3..`.

Closes #434.

### DIFF
--- a/.chglog/config.yml
+++ b/.chglog/config.yml
@@ -5,12 +5,14 @@ info:
   repository_url: https://github.com/pendulum-chain/spacewalk
 options:
   commits:
-    # filters:
-    #   Type: []
   commit_groups:
-    # title_maps: []
   header:
-    pattern: "^(.*)$"
+    pattern: "^(\\w*)[!]?\\:\\s(.*)$"
+    pattern_maps:
+      - Type
+      - Subject
+  merges:
+    pattern: "^.*$"
     pattern_maps:
       - Subject
   notes:
--- a/.chglog/config.yml
+++ b/.chglog/config.yml
@@ -4,13 +4,6 @@ info:
   title: CHANGELOG
   repository_url: https://github.com/pendulum-chain/spacewalk
 options:
-  commits:
-  commit_groups:
-  header:
-    pattern: "^(\\w*)[!]?\\:\\s(.*)$"
-    pattern_maps:
-      - Type
-      - Subject
   merges:
     pattern: "^.*$"
     pattern_maps:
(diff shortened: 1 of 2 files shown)


TorstenStueber:
@ebma The description makes it sound like this will be run by CI. But how does it, it looks to me like you did not add this to any GitHub action.

.chglog/config.yml
```diff
@@ -0,0 +1,20 @@
+style: github
+template: CHANGELOG.tpl.md
+info:
+  title: CHANGELOG
+  repository_url: https://github.com/pendulum-chain/spacewalk
+options:
+  commits:
+  commit_groups:
+  header:
+    pattern: "^(\\w*)[!]?\\:\\s(.*)$"
```
TorstenStueber, line 10: I don't understand why we use this commit header pattern here.

TorstenStueber, line 10: I don't understand why we use this commit header pattern here. (what even is the commit header?)

ebma: AFAIK the commit header is like the first line of your commit which can be 50 characters and the body would be the content you add in new lines.
But we don't need that anyways, so I removed it. Since we squash and merge every feature, it works fine to use the remaining config for merge commits.

ebma:
My description might have been a little confusing but the CI I'm talking about is run on Gitlab pipelines. That's why I didn't add any changes here. This PR is about adding the necessary configuration so that the CI pipeline can run the commands I put into the description to generate meaningful changelogs. 
Changing the Gitlab pipeline is not part of this ticket and something we need to do afterwards.

TorstenStueber (approved):
Great addition, looks good.

ebma:
Merging this although the CI failed because the changes of this PR are unrelated to any relevant code and we already know that [this](https://github.com/pendulum-chain/spacewalk/pull/470) is required to fix the CI.

PR #442, opened 2023-11-01
425 add a new asset category currencyid to the runtime for stablecoins

### LINKED ISSUE #425: Add a new asset category (CurrencyID) to the runtime for stablecoins
## Context
As we are planning a PoC with an EUR stabelcoin issuer to issue the token on Pendulum, we need to add a new category of assets to our runtime.

## Requirement
Define a new asset type (CurrencyID) for assets that are natively issued on Pendulum where the issuer can mint and burn as required.

## Technical details
We should add this new type as `Token(index: u64)` on the `CurrencyId`

ebma:
This PR adds the `Token` to the `CurrencyId` enum used in Spacewalk and also adjusts the other usages accordingly. 

Closes #425.

### DIFF
--- a/clients/runtime/src/types.rs
+++ b/clients/runtime/src/types.rs
@@ -159,6 +159,7 @@ pub mod currency_id {
 						"ZenlinkLPToken({},{},{},{})",
 						token1_id, token1_type, token2_id, token2_type
 					)),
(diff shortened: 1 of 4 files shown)


gianfra-t (approved):
I think this is okay! I checked at the CI failed and it was the test_scheduler_... one, so I re run them. It should pass all tests.

PR #447, opened 2023-11-07
313 - retry integration tests for `tx_bad_seq` errors

### LINKED ISSUE #313: Retry transaction submission in integration tests
Currently, the integration tests might fail if they are executed on two machines roughly at the same time (eg when you execute them on your local machine and in some CI job). The reason is that the hard-coded testing account is used in the integration tests to submit transactions to the Stellar network, which might throw a `tx_bad_seq` error if two submissions happen simultaneously on two jobs. To mitigate this timing issue, we can just make the integration tests retry the submission in case they fail.

b-yap:
closes #313 

Created a loop for `send_payment_to_address` to keep resending the same transaction, should the error be `tx_bad_seq` error. 
This is _only_ for integration tests.

### DIFF
--- a/clients/vault/tests/helper/helper.rs
+++ b/clients/vault/tests/helper/helper.rs
@@ -1,7 +1,7 @@
 use crate::helper::DEFAULT_TESTING_CURRENCY;
 use async_trait::async_trait;
 use frame_support::assert_ok;
-use primitives::{CurrencyId, StellarStroops, H256, stellar::Asset as StellarAsset };
+use primitives::{stellar::Asset as StellarAsset, CurrencyId, StellarStroops, H256};
 use runtime::{
 	integration::{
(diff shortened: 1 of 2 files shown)


ebma (approved):
Nice. I love it, good job @b-yap 👍 

Did you do some testing for this locally and in CI?

b-yap:
@ebma yes I did.

ebma:
Alright. I see that one of the jobs failed with 
```
thread 'tests::stellar_overlay_disconnect_works' panicked at 'Incorrect stellar relay message received', clients/stellar-relay-lib/src/tests/mod.rs:182:9
```
but that's unrelated to the integration tests. Should we create an extra ticket for this?

Since the other actions worked well, I guess we can merge this already.

b-yap:
@ebma   since it  occurred twice  (here and Zoltan's) now then yes i agree about an extra ticket.

PR #453, opened 2023-11-24
452 vault client retries transaction submission for unrecoverable errors

### LINKED ISSUE #452: Vault client retries transaction submission for unrecoverable errors
At the moment, the vault client will permanently retry the submission of a transaction, even if it received some error that makes clear that a resubmission will not succeed anyways.

For example looking at the following logs, when receiving the `InvalidQuorumSetNotEnoughValidators` it's clear that a resubmission of the same extrinsic will result in the same error. Thus, we should not retry because the vault needs to pay the transaction gas fee for each extrinsic call, also the failing ones.

```
Nov 23 18:50:55.065  INFO vault::oracle::agent: get_proof(): Successfully build proof for slot 49136081
Nov 23 18:52:10.528  WARN runtime::retry: Subxt runtime error: Runtime error: Module(ModuleError { pallet: "StellarRelay", error: "InvalidQuorumSetNotEnoughValidators", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [12, 0, 0, 0] } }) - next retry in 0.860 s    
Nov 23 18:52:10.528  WARN runtime::retry: Subxt runtime error: Runtime error: Module(ModuleError { pallet: "StellarRelay", error: "InvalidQuorumSetNotEnoughValidators", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [12, 0, 0, 0] } }) - next retry in 1.240 s    
Nov 23 18:52:10.543  WARN runtime::retry: Subxt runtime error: Runtime error: Module(ModuleError { pallet: "StellarRelay", error: "InvalidQuorumSetNotEnoughValidators", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [12, 0, 0, 0] } }) - next retry in 1.053 s    
Nov 23 18:52:10.776  WARN runtime::retry: Subxt runtime error: Runtime error: Module(ModuleError { pallet: "StellarRelay", error: "InvalidQuorumSetNotEnoughValidators", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [12, 0, 0, 0] } }) - next retry in 1.214 s    
Nov 23 18:52:15.169 ERROR vault::issue: Issue task for slot 49136081 failed with error: "SubxtRuntimeError(Runtime(Module(ModuleError { pallet: \"StellarRelay\", error: \"InvalidQuorumSetNotEnoughValidators\", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [12, 0, 0, 0] } })))"
```

## TODO
We should adjust the [retry logic](https://github.com/pendulum-chain/spacewalk/blob/8bb0d639c6f2e23126f72e39f56c6a497cf2e2d4/clients/runtime/src/rpc.rs#L236) and not retry the submission for unrecoverable errors returned by the Stellar relay pallet.

gianfra-t:
Closes #452 

### Changes
- Uses the string `data` filed on the parsed error to detect recoverable errors (lack of funds, incorrect nonce).
- Otherwise, the service will avoid retry of the same call, since it is impossible to obtain a different result without modifying it (except vault operator's input).
- A new condition is added so that if the error is of type `POOL_UNACTIONABLE` or `POOL_UNKNOWN_VALIDITY` (which are not necessarily invalid transactions) the vault will retry the transaction.
- If the service fails to start it will send a [shutdown signal](https://github.com/pendulum-chain/spacewalk/compare/452-vault-client-retries-transaction-submission-for-unrecoverable-errors?expand=1#diff-0f76594421d4493b3bba1c9ed7864cdab7148f19c5d3a18a9bcc9a174bb85aa8R286) to avoid waiting a dead process.

### DIFF
--- a/clients/runtime/src/error.rs
+++ b/clients/runtime/src/error.rs
@@ -151,9 +151,7 @@ impl Error {
 
 	pub fn is_pool_issue(&self) -> bool {
 		self.map_custom_error(|custom_error| {
-			if custom_error.code() == POOL_UNACTIONABLE ||
-				custom_error.code() == POOL_UNKNOWN_VALIDITY
-			{
+			if custom_error.code() == POOL_UNKNOWN_VALIDITY {
 				Some(())
 			} else {
 				None
@@ -252,4 +250,3 @@ const BASE_ERROR: i32 = 1000;
 const POOL_INVALID_TX: i32 = BASE_ERROR + 10;
 const POOL_UNKNOWN_VALIDITY: i32 = POOL_INVALID_TX + 1;
 const POOL_TOO_LOW_PRIORITY: i32 = POOL_INVALID_TX + 4;
-const POOL_UNACTIONABLE: i32 = POOL_INVALID_TX + 8;
--- a/clients/runtime/src/error.rs
+++ b/clients/runtime/src/error.rs
@@ -46,6 +46,8 @@ pub enum Error {
 	BlockHashNotFound,
 	#[error("Transaction is invalid: {0}")]
 	InvalidTransaction(String),
+	#[error("Transaction is invalid and cannot be recovered: {0}")]
+	InvalidTransactionUnrecoverable(String),
 	#[error("Request has timed out")]
 	Timeout,
 	#[error("Block is not in the relay main chain")]
@@ -121,15 +123,27 @@ impl Error {
 		}
 	}
 
-	pub fn is_invalid_transaction(&self) -> Option<String> {
+	pub fn is_invalid_transaction(&self) -> Option<Error>{
+		// TODO define elsewhere
+		let not_recoverable_errors = ["InvalidQuorumSetNotEnoughValidators"];
+	
 		self.map_custom_error(|custom_error| {
 			if custom_error.code() == POOL_INVALID_TX {
-				Some(custom_error.data().map(ToString::to_string).unwrap_or_default())
+				let data_string = custom_error.data().map(ToString::to_string).unwrap_or_default();
+				
+				for error in not_recoverable_errors {
+					if data_string.contains(error) {
+						return Some(Error::InvalidTransactionUnrecoverable(data_string));
+					}
+				}
+				Some(Error::InvalidTransaction(data_string))
 			} else {
 				None
 			}
 		})
 	}
+
+
 
 	pub fn is_pool_too_low_priority(&self) -> Option<()> {
 		self.map_custom_error(|custom_error| {
--- a/clients/runtime/src/error.rs
+++ b/clients/runtime/src/error.rs
@@ -92,9 +92,14 @@ pub enum Error {
 	TransactionPoolIssue,
 }
 
-pub enum Recoverability{
+pub enum Recoverability {
 	Recoverable(String),
-	Unrecoverable(String)
+	Unrecoverable(String),
+}
+
+enum RecoverableError {
+	InabilityToPayFees,
+	OutdatedTransaction,
 }
 
 impl Error {
@@ -128,21 +133,18 @@ impl Error {
 		}
 	}
 
-	pub fn is_invalid_transaction(&self) -> Option<Recoverability>{
-		// TODO define elsewhere
+	pub fn is_invalid_transaction(&self) -> Option<Recoverability> {
 		let recoverable_errors = ["Inability to pay some fees", "Transaction is outdated"];
-	
+
 		self.map_custom_error(|custom_error| {
 			if custom_error.code() == POOL_INVALID_TX {
 				let data_string = custom_error.data().map(ToString::to_string).unwrap_or_default();
-				
-				for error in recoverable_errors {
-					if data_string.contains(error) {
-						
-						return Some(Recoverability::Recoverable(data_string))
-					}
+
+				if RecoverableError::is_recoverable(&data_string) {
+					return Some(Recoverability::Recoverable(data_string))
 				}
-				return Some(Recoverability::Unrecoverable(data_string));
+
+				return Some(Recoverability::Unrecoverable(data_string))
 			} else {
 				None
 			}
@@ -151,7 +153,9 @@ impl Error {
 
 	pub fn is_pool_issue(&self) -> Option<()> {
 		self.map_custom_error(|custom_error| {
-			if custom_error.code() == POOL_UNACTIONABLE || custom_error.code() == POOL_UNKNOWN_VALIDITY {
+			if custom_error.code() == POOL_UNACTIONABLE ||
+				custom_error.code() == POOL_UNKNOWN_VALIDITY
+			{
 				Some(())
 			} else {
 				None
@@ -210,6 +214,22 @@ impl Error {
 			SECURITY_MODULE,
 			&format!("{:?}", SecurityPalletError::ParachainNotRunning),
 		)
+	}
+}
+
+impl RecoverableError {
+	// The string which is used to match the error type
+	fn as_str(&self) -> &'static str {
+		match self {
+			RecoverableError::InabilityToPayFees => "Inability to pay some fees",
+			RecoverableError::OutdatedTransaction => "Transaction is outdated",
+		}
+	}
+
+	fn is_recoverable(error_message: &str) -> bool {
+		[RecoverableError::InabilityToPayFees, RecoverableError::OutdatedTransaction]
+			.iter()
+			.any(|&error| error_message.contains(error.as_str()))
 	}
 }
 
--- a/clients/runtime/src/error.rs
+++ b/clients/runtime/src/error.rs
@@ -125,18 +125,19 @@ impl Error {
 
 	pub fn is_invalid_transaction(&self) -> Option<Error>{
 		// TODO define elsewhere
-		let not_recoverable_errors = ["InvalidQuorumSetNotEnoughValidators"];
+		let recoverable_errors = [""];
 	
 		self.map_custom_error(|custom_error| {
 			if custom_error.code() == POOL_INVALID_TX {
 				let data_string = custom_error.data().map(ToString::to_string).unwrap_or_default();
 				
-				for error in not_recoverable_errors {
+				for error in recoverable_errors {
 					if data_string.contains(error) {
-						return Some(Error::InvalidTransactionUnrecoverable(data_string));
+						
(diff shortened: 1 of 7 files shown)


clients/runtime/src/error.rs
```diff
@@ -185,6 +215,22 @@ impl Error {
 	}
 }
 
+impl RecoverableError {
+	// The string which is used to match the error type
+	fn as_str(&self) -> &'static str {
+		match self {
```
b-yap, line 221: please add comment to know where these strings came from: 
```rust
   // based on https://github.com/paritytech/polkadot-sdk/blob/b0b4451f31dcd9da04e2952d7c4cdcc219271b88/substrate/primitives/runtime/src/transaction_validity.rs#L100-L118
```

b-yap (changes requested):
is it possible to recreate `POOL_UNACTIONABLE` and/or `POOL_UNKNOWN_VALIDITY` and test it? 

examples: 
[POOL_INVALID_TX](https://github.com/pendulum-chain/spacewalk/blob/82f9e432490b85234f16538b996c589e6af4146e/clients/runtime/src/rpc.rs#L399-L402)
[POOL_TOO_LOW_PRIORITY](https://github.com/pendulum-chain/spacewalk/blob/82f9e432490b85234f16538b996c589e6af4146e/clients/runtime/src/rpc.rs#L431-L434)

clients/runtime/src/error.rs
```diff
@@ -121,10 +133,28 @@ impl Error {
...
+				}
+
+				return Some(Recoverability::Unrecoverable(data_string))
+			} else {
+				None
+			}
+		})
+	}
+
+	pub fn is_pool_issue(&self) -> Option<()> {
```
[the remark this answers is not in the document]
gianfra-t: Sure thing!

gianfra-t:
> is it possible to recreate `POOL_UNACTIONABLE` and/or `POOL_UNKNOWN_VALIDITY` and test it?
> 
> examples: [POOL_INVALID_TX](https://github.com/pendulum-chain/spacewalk/blob/82f9e432490b85234f16538b996c589e6af4146e/clients/runtime/src/rpc.rs#L399-L402) [POOL_TOO_LOW_PRIORITY](https://github.com/pendulum-chain/spacewalk/blob/82f9e432490b85234f16538b996c589e6af4146e/clients/runtime/src/rpc.rs#L431-L434)

I am not sure we can test this since we would need to modify the pool logic. I could not find much information about these errors, but the assumption was that a pool error would not be a problem with the transaction itself.

clients/runtime/src/error.rs
```diff
@@ -121,10 +133,28 @@ impl Error {
...
+		self.map_custom_error(|custom_error| {
+			if custom_error.code() == POOL_UNACTIONABLE ||
+				custom_error.code() == POOL_UNKNOWN_VALIDITY
```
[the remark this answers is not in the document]
gianfra-t: Update: I guess `UnknownTransaction::CannotLookup` could be the only possible recoverable error. `UnknownTransaction::NoUnsignedValidator` is not relevant for us, and `UnknownTransaction::Custom` is only used [once](https://github.com/paritytech/polkadot-sdk/blob/ec3a61ed86305e5b70ce6e466cf127addf560186/bridges/primitives/runtime/src/extensions.rs#L138) and to be honest does not explain anything.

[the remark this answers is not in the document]
ebma: Maybe we should all of them to be non-recoverable. We would have to dig deep to understand the implications of `UnknownTransaction::CannotLookup` so maybe it's best to ignore it.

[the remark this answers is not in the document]
gianfra-t: Yes. I will remove `pool_issue` altogether. If we find this error in the future we can take a look back.

clients/runtime/src/error.rs
gianfra-t: Both error strings are defined [here](https://github.com/pendulum-chain/spacewalk/pull/453/files#diff-6469b7c5aed83b33ec6bbd2a61154190361c333995a98a3c172436491a901b9bR207)

b-yap (changes requested):
Ah, the error still occurred:
```
Dec 05 18:42:57.878  WARN runtime::retry: Subxt runtime error: Runtime error: Module(ModuleError { pallet: "StellarRelay", error: "InvalidQuorumSetNotEnoughValidators", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [12, 0, 0, 0] } }) - next retry in 0.904 s    
Dec 05 18:42:57.879  INFO vault::requests::structs: Successfully executed Redeem request #0x6ec8…9392
Dec 05 18:42:57.879  INFO vault::requests::execution: Successfully executed Redeem request #0x6ec8…9392
Dec 05 18:42:58.198  INFO vault::requests::structs: Successfully executed Redeem request #0xcb6f…4fbd
Dec 05 18:42:58.200  INFO vault::requests::execution: Successfully executed Redeem request #0xcb6f…4fbd
Dec 05 18:43:50.791  WARN runtime::retry: Subxt runtime error: Runtime error: Module(ModuleError { pallet: "StellarRelay", error: "InvalidQuorumSetNotEnoughValidators", description: [], error_data: ModuleErrorData { pallet_index: 68, error: [12, 0, 0, 0] } }) - next retry in 1.691 s 
```

Let's take @ebma 's [assumption](https://github.com/pendulum-chain/spacewalk/issues/452#issuecomment-1825749472) and go ahead flag as "unrecoverable" the module errors.

gianfra-t:
Thanks @b-yap for testing this. I found that there is a [second call](https://github.com/pendulum-chain/spacewalk/pull/453/files#diff-9904d95ed135662bcdf8bdae77e7596c23f821b13632750729b8abeab1685e3aR191) to `notify_retry` inside the vault module. Most likely this is why you still were able to see the vault retrying an unrecoverable error.

As soon as I can test this and replicate that scenario I will update this comment.

b-yap (approved):
let's wait for @ebma

ebma (commented):
It looks good overall. Though one remark that I noticed while testing it:
If the extrinsic to execute the request fails and returns an error, then we are still [retrying](https://github.com/pendulum-chain/spacewalk/blob/0251fcf8eddaec081e972aaf5917c23772590113/clients/vault/src/requests/execution.rs#L154) it up to three times. I think we should not increment the retry_counter [here](https://github.com/pendulum-chain/spacewalk/blob/0251fcf8eddaec081e972aaf5917c23772590113/clients/vault/src/requests/execution.rs#L175) but rather `break` and exit from the loop. Because we know that the request cannot be executed when we retry. 
Or do you remember if there was a good reason to also add the retry here @b-yap? It makes sense in [this](https://github.com/pendulum-chain/spacewalk/blob/0251fcf8eddaec081e972aaf5917c23772590113/clients/vault/src/requests/execution.rs#L177-L183) error block but not in the other one IMO.

clients/vault/src/requests/structs.rs
```diff
@@ -199,11 +199,26 @@ impl Request {
 				)
 			},
 			|result| async {
-				match result {
-					Ok(ok) => Ok(ok),
-					Err(err) if err.is_rpc_disconnect_error() =>
-						Err(runtime::RetryPolicy::Throw(err)),
-					Err(err) => Err(runtime::RetryPolicy::Skip(err)),
+				match result.map_err(Into::<EnrichedError>::into) {
+					Ok(te) => Ok(te),
```
ebma, line 203: What does `te` mean by the way?

gianfra-t: TBH I don't know. It was already [defined](https://github.com/pendulum-chain/spacewalk/pull/453/files#diff-67810aa5914ac045b15035e36b5de374c73cabd4d44ae06f760cfc3dd23c802eL235)  like `te` before.

gianfra-t:
> Or do you remember if there was a good reason to also add the retry here @b-yap? It makes sense in [this](https://github.com/pendulum-chain/spacewalk/blob/0251fcf8eddaec081e972aaf5917c23772590113/clients/vault/src/requests/execution.rs#L177-L183) error block but not in the other one IMO.

But if there is a `Skip` and we keep retrying (inner retry loop) we could eventually hit [here](https://github.com/pendulum-chain/spacewalk/blob/main/clients/runtime/src/retry.rs#L49C5-L49C5) which will also return an error, but not necessarily unrecoverable as we have defined. Perhaps we could break unless it is a `Timeout` variant.

ebma (commented):
I think we can ignore the `Timeout` that is thrown in the backoff policy [here](https://github.com/pendulum-chain/spacewalk/blob/13d25ae642cf43a65ebcb6028f5f0bc5b4e92160/clients/runtime/src/retry.rs#L49C5-L49C5). 
Maybe I'm wrong but the way I see it, then with our current implementation, we will stay [here](https://github.com/pendulum-chain/spacewalk/blob/0251fcf8eddaec081e972aaf5917c23772590113/clients/vault/src/requests/execution.rs#L161) until we either hit an `Unrecoverable` error or the error is recoverable and it was already retried so many times that we encounter the [Timeout](https://github.com/pendulum-chain/spacewalk/blob/13d25ae642cf43a65ebcb6028f5f0bc5b4e92160/clients/runtime/src/retry.rs#L49C5-L49C5) error you mentioned. In both cases we wouldn't want to retry in this outer loop again. Because, if it was `Unrecoverable` we don't want to retry and if we get the `Timeout` then we already retried a lot anyways.

clients/vault/src/requests/structs.rs
gianfra-t: haha yes but in the other [call to notify_retry](https://github.com/pendulum-chain/spacewalk/blob/main/clients/runtime/src/rpc.rs#L235) was like that. Don't worry I will put both back to `ok`.

gianfra-t:
Yes you may be right about the outer retry loop. Either way, if we needed more retries just changing the backoff policy would be enough. Let's wait for @b-yap take on it? Otherwise we just remove the counter there and put a break like you say.

clients/vault/src/requests/structs.rs
ebma: Ah that's what you meant, sorry I didn't see the other one haha.

ebma (approved):
Great changes, good job @gianfra-t 👍 

Tested it locally and looks good to me. Ready for merge once CI passed.

PR #467, opened 2023-12-14
release: Fix bug in test and retry release 1.0.3

ebma:
Fixes some tests. The test to get the current proof was changed to use a random keypair so that we could rule out issues with the client already being connected to the overlay node, thus encountering a timeout in that test.


ebma:
I ran clippy and the tests locally and it passed.

PR #470, opened 2023-12-22
Fix failing test after Stellar testnet reset

### LINKED ISSUE #469: Fix failing test after Stellar testnet reset
The `horizon_get_claimable_balance_success()` test fails because it was referring to a claimable balance created on testnet. We need to fix this.

ebma:
Changes the hardcoded reference to a claimable balance on Stellar mainnet instead of testnet so that this issue doesn't appear for the next reset. I issued this claimable balance entry from one of our accounts so we don't have a risk of someone claiming this entry in the meantime.

Closes #469.


TorstenStueber (approved):
👍

PR #474, opened 2024-01-09
312 use stellar fee statistics to derive transaction fee

### LINKED ISSUE #312: Use Stellar fee statistics to derive transaction fee
## Context 

The vault client currently always uses the same hardcoded fee parameter for creating payment transactions on Stellar. However, at times with heavy load on the Stellar network it might make sense to choose this fee based on the current network statistics. There is an endpoint we can query this fee data from. It's a simple GET request to [this](https://horizon.stellar.org/fee_stats) URL (different for testnet and mainnet of course).

## TODO

Change the vault client so that it uses a fee for Stellar transactions based on the current fee statistics. To do this,
- Poll the fee statistics from https://horizon.stellar.org/fee_stats every 10 minutes
- Use the `p90` field of `fee_charged` as the fee per operation instead of the [DEFAULT_STROOP_FEE_PER_OPERATION](https://github.com/pendulum-chain/spacewalk/blob/a5cc81ececb9a464fc8caa12c191eeaecf453c4b/clients/wallet/src/mock.rs#L25)

b-yap:
closes #312 

1. **Remove** _`DEFAULT_STROOP_FEE_PER_OPERATION`_ constant in production code.
2. **Remove** _`stroop_fee_per_operation`_ parameters on method **`fn send_to_address()`**; since
3. fee will be queried from horizon; the method is called`fn get_fee_stats()`
4. Memoization using the [cached](https://docs.rs/cached/0.47.0/cached/) lib to return the same result until 10 minutes (or 600 seconds) have elapsed.

~~todo: only memoize for `Ok()` results.~~

----  
## How to begin the review:    
1. `clients/wallet/src/types.rs` -> has the `FeeAttribute` enum, where you can pick which one to base the operation on
2. `clients/wallet/src/horizon/responses.rs` -> Defining the response into a struct of `FeeStats` and `FeeDistribution`
3. `clients/wallet/src/horizon/serde.rs` -> Implements deserialization of the fields in `FeeStats` and `FeeDistribution` to u32
5.  `clients/wallet/src/horizon/traits.rs` -> add another method to the trait: `fn get_fee_stats()` and should return the struct `FeeStats`
6. ` clients/wallet/src/horizon/horizon.rs` -> implements `fn get_fee_stats()`
7. `clients/wallet/src/horizon/tests.rs` -> testing the method `fn get_fee_stats()`
8. `clients/wallet/src/stellar_wallet.rs`
    * the method `fn get_fee()` is introduced, with caching: 
       ```
          #[cached(result = true, time = 600)]
      ```
   * the method `fn get_fee()` is called inside `fn send_to_address(...)`, removing the `stroop_fee_per_operation` param
9. ALL calls of `fn send_to_address(...)` will not need to pass a `stroop_fee_per_operation`

### DIFF
--- a/clients/wallet/src/horizon/tests.rs
+++ b/clients/wallet/src/horizon/tests.rs
@@ -213,3 +213,9 @@ async fn fetch_horizon_and_process_new_transactions_success() {
 
 	assert!(!slot_env_map.read().await.is_empty());
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn horizon_get_fee() {
+	let horizon_client = reqwest::Client::new();
+	assert!(horizon_client.get_fee_stats(false).await.is_ok());
+}
--- a/clients/wallet/src/horizon/tests.rs
+++ b/clients/wallet/src/horizon/tests.rs
@@ -218,4 +218,5 @@ async fn fetch_horizon_and_process_new_transactions_success() {
 async fn horizon_get_fee() {
 	let horizon_client = reqwest::Client::new();
 	assert!(horizon_client.get_fee_stats(false).await.is_ok());
+	assert!(horizon_client.get_fee_stats(true).await.is_ok());
 }
--- a/clients/wallet/src/stellar_wallet.rs
+++ b/clients/wallet/src/stellar_wallet.rs
@@ -234,7 +234,7 @@ impl StellarWallet {
 /// This function will be re-executed after the cache expires (according to `time` seconds) OR
 /// when the result is NOT `Ok`.
 #[cached(result = true, time = 600)]
-async fn get_fee(is_public_network: bool, fee_attr: FeeAttribute) -> Result<u32, String> {
+async fn get_fee_stat_for(is_public_network: bool, fee_attr: FeeAttribute) -> Result<u32, String> {
 	let horizon_client = Client::new();
 	let fee_stats = horizon_client
 		.get_fee_stats(is_public_network)
@@ -360,11 +360,12 @@ impl StellarWallet {
 		request_id: [u8; 32],
 		operations: Vec<Operation>,
 	) -> Result<TransactionResponse, Error> {
-		let stroop_fee_per_operation = get_fee(self.is_public_network, FeeAttribute::default())
-			.await
-			.map_err(|e| Error::FailedToGetFee(e))?;
-
 		let _ = self.transaction_submission_lock.lock().await;
+
+		let stroop_fee_per_operation =
+			get_fee_stat_for(self.is_public_network, FeeAttribute::default())
+				.await
+				.map_err(|e| Error::FailedToGetFee(e))?;
 
 		let account = self.client.get_account(self.public_key(), self.is_public_network).await?;
 		let next_sequence_number = account.sequence + 1;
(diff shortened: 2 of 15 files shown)


clients/wallet/src/horizon/tests.rs
```diff
@@ -213,3 +213,9 @@ async fn fetch_horizon_and_process_new_transactions_success() {
 
 	assert!(!slot_env_map.read().await.is_empty());
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn horizon_get_fee() {
+	let horizon_client = reqwest::Client::new();
+	assert!(horizon_client.get_fee_stats(false).await.is_ok());
```
ebma, line 220: Just for the sake of testing both networks, how about we also test `get_fee_stats(true)` in this test?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -229,6 +230,20 @@ impl StellarWallet {
 	}
 }
 
+/// Returns a fee for performing an operation.
+/// This function will be re-executed after the cache expires (according to `time` seconds) OR
+/// when the result is NOT `Ok`.
+#[cached(result = true, time = 600)]
+async fn get_fee(is_public_network: bool, fee_attr: FeeAttribute) -> Result<u32, String> {
```
ebma, line 237: Maybe rename to

```suggestion
async fn get_fee_stat_for(is_public_network: bool, fee_attr: FeeAttribute) -> Result<u32, String> {
```

clients/wallet/src/stellar_wallet.rs
```diff
@@ -339,16 +352,18 @@ impl StellarWallet {
...
 	pub(crate) async fn send_to_address(
 		&mut self,
 		request_id: [u8; 32],
-		stroop_fee_per_operation: u32,
 		operations: Vec<Operation>,
 	) -> Result<TransactionResponse, Error> {
+		let stroop_fee_per_operation = get_fee(self.is_public_network, FeeAttribute::default())
+			.await
+			.map_err(|e| Error::FailedToGetFee(e))?;
+
```
ebma, line 366: I don't think it matters too much but we could move the lock above this function to avoid accidentally querying too often, right?

clients/wallet/src/stellar_wallet.rs
```diff
@@ -718,38 +703,32 @@ mod test {
...
 
 		assert!(response.is_ok());
 
-		let err_insufficient_fee = wallet
+		let tx_failed = wallet
 			.send_payment_to_address(
 				default_destination(),
 				asset.clone(),
-				amount,
+				amount + 100_000_000_000,
```
ebma, line 723: What are we doing here? Transferring a super large amount that we cannot transfer, to force the transaction to fail? If so, maybe worth adding a comment to explain what's happening.

ebma (commented):
I love it, these are some really great changes :tada: Left some minor comments

clients/wallet/src/stellar_wallet.rs
b-yap: moved down, under the lock.

clients/wallet/src/stellar_wallet.rs
b-yap: i added a comment : `// forcefully fail the transaction`

ebma (approved):
LGTM ready to merge once CI passed

PR #475, opened 2024-01-18
Replacing Tokio's `TcpStream` with `async-std` `TcpStream`

b-yap:
Somehow in **_K8s_**, Tokio's [TcpStream](https://github.com/tokio-rs/tokio/blob/1b1b9dc7e388d0619fe7bfe6a7618fff596fdee1/tokio/src/net/tcp/stream.rs#L69-L72) is stuck at "write ready" and is never "read ready". 
Even after connecting using Rust's [Tcpstream](https://doc.rust-lang.org/src/std/net/tcp.rs.html#156-158) and then converting it to [Tokio's](https://github.com/tokio-rs/tokio/blob/1b1b9dc7e388d0619fe7bfe6a7618fff596fdee1/tokio/src/net/tcp/stream.rs#L202-L206), the stream _never_ changes to "read ready".

That's when I decided to use ~Rust's [Tcpstream](https://doc.rust-lang.org/std/net/struct.TcpStream.html) completely.~ [`async-std` library](https://github.com/async-rs/async-std).

In terms of code changes, these are the difference you will see:
| Tokio | Async-Std | ~Rust Std~ | 
| --- | --- | --- |
| Can [split the stream to read/write halves](https://github.com/tokio-rs/tokio/blob/1b1b9dc7e388d0619fe7bfe6a7618fff596fdee1/tokio/src/net/tcp/stream.rs#L1271-L1273), with the write half owned by a struct (`Connector`)| provided an [example](https://book.async.rs/patterns/small-patterns.html?highlight=split#splitting-streams) but `Connector` now owns the `TcpStream`, so I found no need to split | ~Can only [try to clone](https://doc.rust-lang.org/src/std/net/tcp.rs.html#254-256)~ |
| Everything is async | Async | ~Doesn't need async~ |
| Need to add timeout logic | potentially will need it | ~Can [set it in the stream](https://doc.rust-lang.org/src/std/net/tcp.rs.html#298-300)~ |

This also implies removing some tokio features that were added primarily for TCP reasons: `net` and `io-util`.

----  
## How to begin the review:   
1. clients/stellar-relay-lib/src/connection/connector/connector.rs 
     *  As mentioned above, the whole `TcpStream` is owned by the Connector:
        ```rust
             pub struct Connector {
                 ...
                 - write_stream_overlay: OwnedWriteHalf,
                 + tcp_stream: TcpStream,
             }
        ```
     * renamed `fn new(...)` to `fn start(...)` where it also **_starts connecting to Stellar Node_**
2. clients/stellar-relay-lib/src/overlay.rs
     * `StellarOverlayConnection`'s `fn connect(...)` will directly use `Connector`'s `fn start(...) . No more `fn create_stream(...)` calls.
     * renamed `fn disconnect()` to `fn stop()` to make it similar to other structs (`Agent`, `Connector`).
3. clients/stellar-relay-lib/src/connection/helper.rs
     * removal of `fn create_stream(...)`
4. clients/stellar-relay-lib/src/connection/connector/message_reader.rs
     * all `r_stream: &mut tcp::OwnedReadHalf` is replaced with `mut stream: TcpStream`, removing `async` in these methods.
     * in `fn read_message_from_stellar(...)`:
          * instead of "peeking" first, I changed it to actually **"reading"** the 1st 4 bytes, to avoid double operations. On the previous logic, it usually 1. peeks first, 2. and then read. Doing only 1 operation saves time.
          * ~to pass the stream around, it has to be cloned; hence the call to`try_clone()`~ 
          * a reference of the `Connector` is being passed around. This eliminates cloning.
5. clients/vault/src/oracle/testing_utils.rs
    * created a helper function: `fn specific_stellar_relay_config(...)` to specifically connect to certain config. And there are only 3 choices, based on the _**index**_ of its list:
        * testnet have: sdftest1, sdftest2, sdftest3
        * mainnet have: Iowa, Frankfurt, Singapore

### DIFF
--- a/clients/stellar-relay-lib/src/connection/connector/message_reader.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/message_reader.rs
@@ -106,6 +106,24 @@ async fn read_message_from_stellar(
 							  tokio::io::Interest::WRITABLE)
 		).await {
 			Ok(Ok(res)) => {
+				match r_stream.peer_addr() {
+					Ok(peer_addr) => {
+						log::trace!("read_message_from_stellar(): peer_addr: {peer_addr}");
+					}
+					Err(e) => {
+						log::error!("read_message_from_stellar(): no peer addresses returned: {e:?}");
+					}
+				}
+
+				match r_stream.local_addr() {
+					Ok(local_addr) => {
+						log::trace!("read_message_from_stellar(): local_addr: {local_addr}");
+					}
+					Err(e) => {
+						log::error!("read_message_from_stellar(): no local addresses returned: {e:?}");
+					}
+				}
+
 				if res.is_readable() {
 					log::trace!("read_message_from_stellar(): stream is readable");
 				}
@@ -117,6 +135,13 @@ async fn read_message_from_stellar(
 				if res.is_empty() {
 					log::trace!("read_message_from_stellar(): stream is empty");
 				}
+
+				if res.is_read_closed() {
+					log::error!("read_message_from_stellar(): stream's read half is closed");
+					drop(r_stream);
+					return Err(Error::Disconnected);
+				}
+
 			}
 			Ok(Err(e)) => {
 				log::error!("read_message_from_stellar(): Error occurred during checking ready status: {e:?}");
@@ -125,26 +150,6 @@ async fn read_message_from_stellar(
 				log::error!("read_message_from_stellar(): timeout elapsed for checking ready status");
 				return Err(Error::Timeout)
 			}
-		}
-		let read_ready =
-			r_stream.ready(tokio::io::Interest::READABLE).await
-			.map_err(|e| {
-				log::error!("read_message_from_stellar(): failed to check if stream is ready: {e:?}");
-				Error::ReadFailed(e.to_string())
-			})?;
-
-		if read_ready.is_empty() {
-			log::trace!("read_message_from_stellar(): reading returned empty.");
-			continue
-		}
-
-		if read_ready.is_read_closed() {
-			log::trace!("read_message_from_stellar(): read stream is closed");
-			return Err(Error::Disconnected);
-		}
-
-		if read_ready.is_readable() {
-			log::trace!("read_message_from_stellar(): stream is read ready");
 		}
 
 		match timeout(Duration::from_secs(timeout_in_secs), r_stream.peek(&mut buff_for_peeking))
--- a/clients/stellar-relay-lib/src/connection/connector/message_reader.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/message_reader.rs
@@ -2,9 +2,11 @@ use crate::connection::{xdr_converter::get_xdr_message_length, Connector, Error,
 use std::{
 	io::Read,
 	net::{Shutdown, TcpStream},
+	sync::{Arc, Mutex},
 };
 use substrate_stellar_sdk::{types::StellarMessage, XdrCodec};
 
+use std::{thread, time::Duration};
 use tokio::sync::{mpsc, mpsc::error::TryRecvError};
 
 /// Polls for messages coming from the Stellar Node and communicates it back to the user
@@ -20,7 +22,7 @@ pub(crate) async fn poll_messages_from_stellar(
 	mut send_to_node_receiver: mpsc::Receiver<StellarMessage>,
 ) {
 	log::info!("poll_messages_from_stellar(): started.");
-
+	// clone the stream to perform a read operation on the next function calls
 	loop {
 		if send_to_user_sender.is_closed() {
 			log::info!("poll_messages_from_stellar(): closing receiver during disconnection");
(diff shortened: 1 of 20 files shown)


TorstenStueber:
@b-yap as this could be a bug in Tokios TCP implementation, would it make sense to report a bug ticket to Tokio?

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -35,37 +31,56 @@ pub(crate) async fn poll_messages_from_stellar(
...
+				Err(e) => {
+					log::error!("poll_messages_from_stellar(): Error occurred during cloning tcp stream: {e:?}");
+					break
+				},
+				Ok(stream_clone) => stream_clone,
+			};
+
 		// check for messages from Stellar Node.
-		match read_message_from_stellar(&mut read_stream_overlay, connector.timeout_in_secs).await {
+		let xdr = match read_message_from_stellar(stream_clone) {
```
gianfra-t, line 55: Would it make sense here to use `tokio::task::spawn_blocking` or `std::thread::spawn` ? I imagine this reading operation could take a while if the network has delays, or the stellar node is slow, etc. and it may block the thread. With the `async` version of the `TcpStream` this wasn't a concern.

For message passing `mpsc::unbounded_channel` apparently would [work](https://stackoverflow.com/questions/61168507/wrapping-blocking-mpsc-in-async-rust-tokio).

Or perhaps this is overkill for this application and we are fine with blocking?

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -74,36 +89,25 @@ pub(crate) async fn poll_messages_from_stellar(
...
-) -> Result<Xdr, Error> {
+fn read_message_from_stellar(mut stream: TcpStream) -> Result<Xdr, Error> {
 	// holds the number of bytes that were missing from the previous stellar message.
 	let mut lack_bytes_from_prev = 0;
 	let mut readbuf: Vec<u8> = vec![];
 
-	let mut buff_for_peeking = vec![0; 4];
+	let mut buff_for_reading = vec![0; 4];
 	loop {
+		let stream_clone = stream.try_clone()?;
```
gianfra-t, line 99: I don't know how much of a cost is to clone the stream every time in the reading loop, would an `Arc<Mutex<TcpConecction>` be useful to avoid cloning? 
Then something like `tcp_stream.lock().unwrap();` would give you the stream.

Also could be needed for the other comment about `spawn_blocking` (in case that solution was needed).

gianfra-t (commented):
Very nice you were able to solve this issue @b-yap !! Super strange why the stream gets stuck and is not able to read only in Kubernetes.

I just left some comments that in general are about using the non `async` version of the `TcpStream` and if they may block the thread in a meaningful way, when reading particularly.

bogdanS98 (commented):
Everything looks fine except I'm not really sure if synchronous execution is going to achieve the same performance as async. Is there any particular reason why `std::net::TcpStream` was used instead of `async_std::net::TcpStream`?

b-yap:
@bogdanS98 @gianfra-t 
After Gian's comment about using `spawn_blocking(...)`, I am also finding `std` as NOT the best solution.

If we're moving to `async_std`, it makes sense to remove everything related to tokio, yes? 
I just checked [`async_std`'s repo](https://github.com/async-rs/async-std) and the last update was about 9 months ago?
I'm running through their issues and found:
* https://github.com/async-rs/async-std/issues/1063

clients/vault/src/oracle/agent.rs
```diff
@@ -215,11 +221,15 @@ mod tests {
...
+		// let it run for a few seconds, making sure that the other tests have successfully shutdown
+		// their connection to Stellar Node
+		sleep(Duration::from_secs(2)).await;
+
 		let shutdown_sender = ShutdownSender::new();
 
 		// We use a random secret key to avoid conflicts with other tests.
 		let agent = start_oracle_agent(
-			get_test_stellar_relay_config(true),
+			specific_stellar_relay_config(true, 0),
```
gianfra-t, line 232: Just curious, why do we want to get always the first node choice here? and then on the other test always the [second](https://github.com/pendulum-chain/spacewalk/pull/475/files#diff-732d3f237cd7cd9cf03fc64a9ae387ce0e80ad0928e995a3b9e61df4a4c24429R266).

gianfra-t (approved):
All looks good to me @b-yap! Amazing that you found out how to solve this weird problem.

clients/vault/src/oracle/agent.rs
b-yap: ah, sometimes an error `already connected peer` happens. I tried connecting to different nodes just to see if it works :crossed_fingers:. The next best thing is to actually connect with different stellar accounts. Currently we only have 1 test account to play with.

ebma (approved):
Looks good to me too, great job 🎉 I think these changes even simplify the connectivity logic we had because we don't have to pass the two halfs of the stream around, which is even better.

Did you already test with Zoltan if this implementation works in Kubernetes @b-yap?

b-yap:
Couldn't have done it without you guys, @bogdanS98 @gianfra-t .

@ebma  yes I did

PR #481, opened 2024-01-24
Replace tokio's TcpStream with async-std

b-yap:
[As suggested by](https://github.com/pendulum-chain/spacewalk/pull/475#pullrequestreview-1831996669) @bogdanS98, replace _`tokio's`_ `TcpStream` with **_`async-std`_**'s.

This is to continue from https://github.com/pendulum-chain/spacewalk/pull/475


clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -93,18 +79,8 @@ async fn read_message_from_stellar(stream_clone: Arc<Mutex<TcpStream>>) -> Resul
...
-		};
-
-		match result {
-			Ok(size) if size == 0 => {
-				// No data available to read
-				tokio::task::yield_now().await;
-				continue
-			},
+		match connector.tcp_stream.read(&mut buff_for_reading).await {
+			Ok(size) if size == 0 => continue,
```
gianfra-t, line 83: We ended up not needing `yield_now` here? that is great news! That must mean that `async_std` is indeed compatible with Tokio.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -141,17 +116,14 @@ async fn read_message_from_stellar(stream_clone: Arc<Mutex<TcpStream>>) -> Resul
 				// The next few bytes was read. Add it to the readbuf.
 				lack_bytes_from_prev -= size;
 				readbuf.append(&mut buff_for_reading);
+				// make sure to cleanup the buffer
+				buff_for_reading = vec![0; 4];
```
gianfra-t, line 120: Why did we not needed before to clean the buffer at this stage? I imagined that if we don' clean it, on the next read it would be overwritten. Or is it because it may read less than 4 and therefore have partially old data on it?
Anyway it seems a good addition!

gianfra-t (approved):
All looks good 🎉.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
b-yap: This is actually a surprise I found: 
Some of the integ tests failed because a proof could not be formed. I logged [`fn listen()` of overlay](https://github.com/pendulum-chain/spacewalk/blob/cce0b3d25bf9e4b8add7b53fe38f1cdad1f74791/clients/stellar-relay-lib/src/overlay.rs#L75) and it was always empty, even after sending a message to Stellar Node that would send back a legit Stellar Message.

It seems like the bytes concatenated was wrong; I had to flush the first few bytes before reading a next batch.

The first 4 bytes is to determine the length (how many bytes) of the next Stellar Message .

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
b-yap: yes!!! it actually was because of https://github.com/pendulum-chain/spacewalk/pull/481#discussion_r1467226284.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
gianfra-t: Makes sense!

PR #483, opened 2024-01-29
431 Add environment variable options to Spacewalk

### LINKED ISSUE #431: Add environment variable options to Spacewalk
Can we add the parameters that spacewalk uses as environment variables as well? So the way it would work that spacewalk would take the environment variables instead of the command line parameters if there is nothing defined in the command line.

b-yap:
closes #431 

adding the ff. arguments an ENV option:
1. KEYRING
2. KEYFILE
3. KEYNAME
4. SPACEWALK_PARACHAIN_URL
5. STELLAR_VAULT_SECRET_KEY_FILEPATH
6. STELLAR_OVERLAY_CONFIG_FILEPATH
7. AUTO_REGISTER

note for **_AUTO_REGISTER_**:  since the value must be a tuple, make sure to **_enclose it with double quotes_**. 

:no_good: :x: `export AUTO_REGISTER=0,GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN:USDC,1000000` :x: :no_good:

:white_check_mark: :ok_hand: **`export AUTO_REGISTER="0,GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC:USDC,1000000"`** :ok_hand: :white_check_mark:

For arguments that accept _`option`_,  make sure to **_enclose it with double quotes_** with the _ACTUAL_ value.
:no_good: :x: `export MAX_NOTIFS_PER_SUBSCRIPTION="1"` :x: :no_good:
:no_good: :x: `export MAX_NOTIFS_PER_SUBSCRIPTION="Some(1)"` :x: :no_good:

:white_check_mark: :ok_hand: **`export MAX_NOTIFS_PER_SUBSCRIPTION=1`** :ok_hand: :white_check_mark:

### DIFF
--- a/clients/runtime/src/cli.rs
+++ b/clients/runtime/src/cli.rs
@@ -13,17 +13,17 @@ use crate::{
 #[derive(Parser, Debug, Clone)]
 pub struct ProviderUserOpts {
 	/// Keyring to use, mutually exclusive with keyfile.
-	#[clap(long, required_unless_present = "keyfile", parse(try_from_str = parse_account_keyring))]
+	#[clap(long, env = "KEYRING", required_unless_present = "keyfile", parse(try_from_str = parse_account_keyring))]
 	pub keyring: Option<AccountKeyring>,
 
 	/// Path to the json file containing key pairs in a map.
 	/// Valid content of this file is e.g.
 	/// `{ "MyUser1": "<Polkadot Account Mnemonic>", "MyUser2": "<Polkadot Account Mnemonic>" }`.
-	#[clap(long, conflicts_with = "keyring", requires = "keyname")]
+	#[clap(long, env = "KEYFILE", conflicts_with = "keyring", requires = "keyname")]
 	pub keyfile: Option<String>,
 
 	/// The name of the account from the keyfile to use.
-	#[clap(long, conflicts_with = "keyring", requires = "keyfile")]
+	#[clap(long, env = "KEYNAME", conflicts_with = "keyring", requires = "keyfile")]
 	pub keyname: Option<String>,
 }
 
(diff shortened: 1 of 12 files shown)


TorstenStueber (approved):
Looks good.

ebma (commented):
Looks good, but is there a reason not to make all configuration parameters available as environment variables? We don't really use the ones that you didn't add but they could be useful for other users.

b-yap:
@ebma I added env option for ALL args.

clients/runtime/src/cli.rs
```diff
@@ -93,15 +93,15 @@ pub struct ConnectionOpts {
 	pub spacewalk_parachain_url: String,
 
 	/// Timeout in milliseconds to wait for connection to spacewalk-parachain.
-	#[clap(long, parse(try_from_str = parse_duration_ms), default_value = "60000")]
+	#[clap(long, env = "SPACEWALK_PARACHAIN_CONN_TIMEOUT_MS", parse(try_from_str = parse_duration_ms), default_value = "60000")]
```
ebma, line 96:
```suggestion
	#[clap(long, env = "SPACEWALK_PARACHAIN_CONNECTION_TIMEOUT_MS", parse(try_from_str = parse_duration_ms), default_value = "60000")]
```

clients/runtime/src/cli.rs
```diff
@@ -93,15 +93,15 @@ pub struct ConnectionOpts {
...
 	pub spacewalk_parachain_connection_timeout_ms: Duration,
 
 	/// Maximum number of concurrent requests
-	#[clap(long)]
+	#[clap(long, env = "MAX_CONCURRENT_REQUESTS")]
 	pub max_concurrent_requests: Option<usize>,
 
 	/// Maximum notification capacity for each subscription
-	#[clap(long)]
+	#[clap(long, env = "S")]
```
ebma:
```suggestion
	#[clap(long, env = "MAX_NOTFIS_PER_SUBSCRIPTION")]
```

ebma (commented):
Since I guess there is no way to spit out all available names of environment variables from the vault binary, I think it's best to give each env variable the very same name as the config item. Then it's at least somewhat obvious which names are used/available.

clients/runtime/src/cli.rs
b-yap: I use `--help` to show all args.

ebma (approved):

Issue #277, opened 2023-02-27
Rename type `Ledger` to `Slot`


b-yap:
As [mentioned](https://github.com/pendulum-chain/spacewalk/pull/276#discussion_r1118595357) by @ebma :
> While 'Ledger' is used to describe a block in the Stellar world, the term 'slot' is rather used in the SCP consensus paper etc. But at the end of the day it boils down to the same thing when it comes to the number we store for a ledger and a slot. For one transaction x the slot and ledger will be equal to each other.

> Thus, I would prefer if we find a better way for describing this. If I remember correctly we already have a type for Slot, maybe we could just re-use that one.

b-yap:
@ebma  
While I'm all about Slot == Ledger, there's a difference of datatypes.

[Slot is defined as u64](https://github.com/pendulum-chain/spacewalk/blob/6c272183f3a9a766c90adc665a0f2e0929903ccf/clients/vault/src/oracle/types.rs#L10): 
```rust
pub type Slot = Uint64;
```
and [Ledger is u32](https://github.com/pendulum-chain/spacewalk/blob/6c272183f3a9a766c90adc665a0f2e0929903ccf/clients/wallet/src/lib.rs#L12):
```rust
pub type Ledger = u32;
``` 

Shall we address where to handle the conversion from Ledger to Slot? And also, where to define the `Slot` type to make it easily accessible?

Can we actually cleanup the following:
* [`let slot = transaction.ledger as Slot;`](https://github.com/pendulum-chain/spacewalk/blob/dev/milestone-3/clients/vault/src/execution.rs#L274) 
* [`let slot_index: u32 = slot.try_into().unwrap();` in impl.rs](https://github.com/pendulum-chain/spacewalk/blob/6c272183f3a9a766c90adc665a0f2e0929903ccf/clients/vault/src/oracle/storage/impls.rs#L419)
* [`let slot = response.ledger as u64;`](https://github.com/pendulum-chain/spacewalk/blob/6c272183f3a9a766c90adc665a0f2e0929903ccf/clients/vault/tests/vault_integration_tests.rs#L110)
* [`let slot_index: u32 = slot.try_into().unwrap();` in proof_builder.rs](https://github.com/pendulum-chain/spacewalk/blob/dev/milestone-3/clients/vault/src/oracle/collector/proof_builder.rs#L238)

---

> **_'Ledger' is used to describe a block in the Stellar world,_**  

[...]

PR #486, opened 2024-02-05
417 Fix 4FC-05 commented out code

### LINKED ISSUE #417: 4FC-05 Commented out code
We need to clean up some commented-out code and TODO comments that were introduced by #385.

b-yap:
closes #417  and https://github.com/pendulum-chain/spacewalk/issues/277

Please review https://github.com/pendulum-chain/substrate-stellar-sdk/pull/25 first.

Aside from the [following todos](https://github.com/pendulum-chain/spacewalk/issues/417#issuecomment-1931416492) that will not be included in this PR, 2 big things need more info:

* The benchmark changes can be reviewed easily with this commit: https://github.com/pendulum-chain/spacewalk/commit/efc306c8de377468c6f0e633b44f30f96d4a55c6 .

----
For https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/wallet/src/resubmissions.rs#L250-L251 
* Firstly get the iterator:
   ```rust
   let mut iter = match self.get_all_transactions_iter().await {
			Ok(iter) => iter,
			Err(e) => {
				tracing::warn!("is_transaction_already_submitted(): failed to get iterator: {e:?}");
				return false
			},
		};
   ```
   of datatype _`TransactionsResponseIter`_  containing a list of transactions (field `records`). 

* The `records` is sorted in DESCENDING order, which means the `sequence number` of the transaction will be _smaller_ the further we iterate. The loop goes through the `records` in this order:
   * _top_ of the list
   * _middle_ of the list
   * _end_ of the list

*  Compare the target transaction to the transaction of the list in this order:
    ```rust
     fn _check_transaction_match(
	  tx: &Transaction,
	  tx_resp: &TransactionResponse,
	  public_key: &PublicKey,
     ) -> Result<bool, Option<SequenceNumber>> {
	  // Make sure that we are the sender and not the receiver because otherwise an
	  // attacker could send a transaction to us with the target memo and we'd wrongly
	  // assume that we already submitted this transaction.
	  if !is_source_account_match(public_key, &tx_resp) {
		return Err(None)
	  }

	  let Ok(source_account_sequence)  = tx_resp.source_account_sequence() else {
		tracing::warn!("_check_transaction_match(): cannot extract sequence number of transaction response: 
      {tx_resp:?}");
		return Err(None)
	  };

	  // check if the sequence number is the same as this response
	  if tx.seq_num == source_account_sequence {
		// Check if the transaction contains the memo we want to send
		return Ok(is_memo_match(tx, &tx_resp))
	  }

	  Err(Some(source_account_sequence))
    }
    ```
    a description can be found on top of this function.
* If the transaction at the middle of the list is not a match, use its sequence number to determine where a match is possible. Remove half of that list to narrow down the search.
   ```rust
   match _check_transaction_match(tx, &response, public_key) {
		Ok(res) => return Some(res),
		Err(Some(source_account_sequence)) => {
			// if the sequence number is GREATER than this response,
			// then a match must be in the first half of the list.
			if tx_sequence_num > source_account_sequence {
				iter.remove_last_half_records();
			}
			// a match must be in the last half of the list.
			else {
				iter.remove_first_half_records();
			}
		},
		_ => {},
	}
   ```
* If the transaction at the end of the list is not a match, use its sequence number to determine if a match can be found in the current `records` list, or on the `next page`.
  ```rust
	match _check_transaction_match(tx, &response, public_key) {
		Ok(res) => return Some(res),
		Err(Some(source_account_sequence)) => {
			// if the sequence number is LESSER than this response,
			// then a match is possible on the NEXT page
			if tx_sequence_num < source_account_sequence {
				if let None = iter.jump_to_next_page().await {
					// there's no pages left, meaning there's no other transactions to compare
					return Some(false)
				}
			}
		},
		_ => {},  
  ```


clients/runtime/Cargo.toml
```diff
@@ -14,9 +14,9 @@ testing-utils = [
     "tempdir",
     "rand",
     "testchain",
-    "testchain-runtime",
+    "testchain-runtime/testing-utils",
```
b-yap, line 17: addresses https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/testchain/runtime/src/lib.rs#L451-L460.

this way, testing ONE intergration test case will work.

clients/runtime/src/rpc.rs
```diff
@@ -1436,17 +1436,24 @@ impl ReplacePallet for SpacewalkParachain {
 
 #[async_trait]
 pub trait StellarRelayPallet {
-	async fn is_public_network(&self) -> Result<bool, Error>;
+	async fn is_public_network(&self) -> bool;
```
b-yap, line 1439: addresses https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/vault/src/system.rs#L826

By default we're assuming `public_network` = true anyway:
https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/vault/src/system.rs#L713-L714 So might as well have this return a bool.

This function will be called multiple times in `system.rs`.
Updating the faucet itself will need to check the stellar network.

clients/stellar-relay-lib/src/connection/xdr_converter.rs
```diff
@@ -1,10 +1,8 @@
-#![allow(dead_code)] //todo: remove after being tested and implemented
-
 use crate::sdk::types::{
 	AuthenticatedMessage, AuthenticatedMessageV0, HmacSha256Mac, MessageType, StellarMessage,
 };
 use std::fmt::Debug;
-use substrate_stellar_sdk::XdrCodec;
+use substrate_stellar_sdk::{parse_stellar_type, StellarSdkError, XdrCodec};
```
b-yap, line 5: `parse_stellar_type` was moved to address
https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/stellar-relay-lib/src/connection/xdr_converter.rs#L40-L41.
See https://github.com/pendulum-chain/substrate-stellar-sdk/pull/25

clients/stellar-relay-lib/src/connection/xdr_converter.rs
```diff
@@ -37,36 +44,6 @@ pub(crate) fn from_authenticated_message(message: &AuthenticatedMessage) -> Resu
 	message_to_bytes(message)
 }
 
-// todo: move to substrate-stellar-sdk
```
b-yap, line 40: see https://github.com/pendulum-chain/substrate-stellar-sdk/pull/25

clients/stellar-relay-lib/src/connection/xdr_converter.rs
```diff
@@ -135,21 +112,19 @@ pub fn log_decode_error<T: Debug>(source: &str, error: T) -> Error {
 	Error::DecodeError(source.to_string())
 }
 
-// extra function.
-fn is_xdr_complete_message(data: &[u8], message_len: usize) -> bool {
```
b-yap, line 139: only used in test cases, thus is moved under `mod test { .. }`.

clients/vault/src/oracle/agent.rs
```diff
@@ -210,7 +210,7 @@ impl OracleAgent {
 #[cfg(test)]
 mod tests {
 	use crate::oracle::{
-		get_random_secret_key, get_test_secret_key, specific_stellar_relay_config,
+		get_mainnet_secret_key, get_random_secret_key, specific_stellar_relay_config,
```
b-yap, line 213: addresses https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/vault/tests/helper/mod.rs#L33.
We have 2 testnet secret keys. To cleanly switch between testing at mainnet or testnet, separeted what was once just `fn get_test_secret_key()` to:
* `get_mainnet_secret_key()`
* `get_testnet_secret_key(...)`

clients/vault/src/oracle/collector/collector.rs
```diff
@@ -374,47 +371,44 @@ mod test {
...
+		let txset = txsets_map.get(&txset_slot).expect("should return a tx set");
+		collector.txset_map.write().insert(env_slot, txset.clone());
 
 		assert!(collector.envelopes_map.read().contains(&env_slot));
-		//assert!(collector.txset_map.read().contains(&env_slot));
-		// assert!(collector.slot_watchlist.read().contains_key(&env_slot));
+		assert!(collector.txset_map.read().contains(&env_slot));
 
 		collector.remove_data(&env_slot);
+		collector.remove_data(&txset_slot);
```
b-yap, line 387: I did not generate new sample files for the `env_slot`, so it makes sense to call `remove_data()` for the `txset_slot`, and check if it was successful.

clients/vault/src/oracle/storage/impls.rs
```diff
@@ -227,37 +227,36 @@ mod test {
 			assert_eq!(&file_name, &expected_name);
 		}
 
-		// todo: enable after generating a new sample of txsetmap
```
b-yap, line 230: new sample files have been generated. These tests can be re-enabled.

clients/vault/src/oracle/testing_utils.rs
```diff
@@ -43,9 +43,16 @@ fn stellar_relay_config_abs_path(
...
 
-pub fn get_test_secret_key(is_mainnet: bool) -> String {
-	let file_name = if is_mainnet { "mainnet" } else { "testnet" };
-	let path = format!("./resources/secretkey/stellar_secretkey_{file_name}");
+pub fn get_mainnet_secret_key() -> String {
+	let path = "./resources/secretkey/stellar_secretkey_mainnet";
+	std::fs::read_to_string(path).expect("should return a string")
+}
+
+/// return a secret key for testnet with 2 choices; one has xlm, the other does not
```
b-yap, line 51: again, this is to address 
https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/vault/tests/helper/mod.rs#L33

clients/vault/src/system.rs
```diff
@@ -431,11 +431,18 @@ impl VaultService {
 			oracle_agent,
 			self.config.payment_margin_minutes,
 		);
+
+		let shutdown_clone = self.shutdown.clone();
 		service::spawn_cancelable(self.shutdown.subscribe(), async move {
-			// TODO: kill task on shutdown signal to prevent double payment
```
b-yap, line 435: I followed how [Interlay]([Interlay](https://github.com/pendulum-chain/spacewalk/pull/486/commits/80fba1217cb81e368ee6815d3f75516d87c214e1)) did it.

clients/vault/src/system.rs
```diff
@@ -813,9 +812,7 @@ impl VaultService {
 	}
 
 	async fn register_public_key_if_not_present(&mut self) -> Result<(), Error> {
-		if let Some(_faucet_url) = &self.config.faucet_url {
-			// TODO fund account with faucet
-		}
+		let _ = self.try_fund_from_faucet().await;
```
b-yap, line 815: here it tries to fund the account with faucet.

clients/vault/src/system.rs
```diff
@@ -832,6 +829,37 @@ impl VaultService {
...
+	/// Only works when the stellar network is testnet
+	async fn try_fund_from_faucet(&self) -> bool {
+		let Some(faucet_url) = &self.config.faucet_url else {
+			return false
+		};
+
+		let is_public_network = self.spacewalk_parachain.is_public_network().await;
+
+		// fund the account if on stellar TESTNET
+		if !is_public_network {
```
b-yap, line 841: the deployed faucet can only be used for stellar TESTNET.

clients/vault/src/system.rs
```diff
@@ -860,11 +888,7 @@ impl VaultService {
 				)
 				.await
 				.map_err(|e| Error::RuntimeError(e))
-		} else if let Some(_faucet_url) = &self.config.faucet_url {
-			tracing::info!("[{}] Automatically registering...", vault_id.pretty_print());
-			// TODO
-			// faucet::fund_and_register(&self.spacewalk_parachain, faucet_url, &vault_id)
-			// 	.await?;
+		} else if self.try_fund_from_faucet().await {
```
b-yap, line 891: here we want to know if it was successful or not; if it was, then all good. If not, an error is returned.

clients/wallet/src/horizon/horizon.rs
```diff
@@ -128,11 +128,11 @@ impl HorizonClient for reqwest::Client {
 	) -> Result<TransactionResponse, Error> {
 		let seq_no = transaction_envelope.sequence_number();
 
-		tracing::debug!("submitting transaction with seq no: {seq_no:?}: {transaction_envelope:?}");
-
 		let transaction_xdr = transaction_envelope.to_base64_xdr();
 		let transaction_xdr = std::str::from_utf8(&transaction_xdr).map_err(Error::Utf8Error)?;
 
+		tracing::debug!("submitting transaction with seq no: {seq_no:?}: {transaction_xdr:?}");
```
b-yap, line 134: it is best to log the xdr format of the envelope; easier to decode using stellar laboratory

clients/wallet/src/horizon/mod.rs
```diff
@@ -8,6 +8,3 @@ mod tests;
 
 pub use horizon::*;
 pub use traits::HorizonClient;
-
-// todo: change to Slot
-pub type Ledger = u32;
```
b-yap, line 13: `Slot` is available and can be used: 
   https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/wallet/src/types.rs#L6

so there's really no need for `type Ledger`.

clients/wallet/src/horizon/responses.rs
```diff
@@ -471,29 +473,54 @@ impl<C: HorizonClient> TransactionsResponseIter<C> {
...
-				self.records = response.records();
-
+				let _ = self.jump_to_next_page().await?;
 				self.get_top_record()
 			},
 		}
 	}
+
+	/// returns the TransactionResponse in the middle of the list
+	pub fn middle(&mut self) -> Option<TransactionResponse> {
```
b-yap, line 496: this is important for https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/wallet/src/resubmissions.rs#L250

clients/wallet/src/horizon/tests.rs
```diff
@@ -175,17 +175,22 @@ async fn fetch_transactions_iter_success() {
...
 	for _ in 0..txs_iter.records.len() {
 		assert!(txs_iter.next().await.is_some());
 	}
 
-	// the list should be empty, as the last record was returned.
+	// the list should be empty, as the last record of this page was returned.
 	assert_eq!(txs_iter.records.len(), 0);
 
-	// todo: when this account's # of transactions is more than 200, add a test case for it.
+	// if the next page
```
b-yap, line 185: 200 is the maximum # of transactions per page. 
The account has more that 200 transactions. This tests the `fn next()` actually returns the first transaction of the next page.

pallets/oracle/src/tests.rs
```diff
@@ -204,13 +204,6 @@ fn test_is_invalidated() {
 
 		assert_ok!(Oracle::feed_values(3, vec![(key.clone(), rate)]));
 		mine_block();
-
-		// max delay is 60 minutes, 60+ passed
-		// assert!(Oracle::is_outdated(&key, now + 3601));//TODO
```
b-yap, line 209: this method does not exist anymore. We can just remove them.

testchain/runtime/src/lib.rs
```diff
@@ -448,16 +448,13 @@ impl NativeCurrencyKey for SpacewalkNativeCurrencyKey {
...
-		// todo: \"Unable to get required collateral for amount\":
-		// todo: Module(ModuleError { index: 19, error: [0, 0, 0, 0], message: None })", data: None
-		// } cfg_if::cfg_if! {
-		// 	if #[cfg(not(feature = "testing-utils"))] {
-		// 		if token_symbol == 0 {
-		// 			return Some((b"Kusama".to_vec(), b"KSM".to_vec()))
-		// 		}
-		// 	}
-		// }
+		cfg_if::cfg_if! {
```
b-yap, line 451: re-enabled, as the only thing required in the vault's integration test, was to specify the `testing-utils` feature in the cargo.toml of the _client/runtime_

testchain/node/src/service.rs
```diff
@@ -104,13 +104,6 @@ pub fn new_partial(
 
 	let select_chain = sc_consensus::LongestChain::new(backend.clone());
 
-	// TODO check if this is needed
```
b-yap, line 107: https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/runtime/src/rpc.rs#L132-L134 The vault calls `manual_seal()` when doing unit/integration testing, so this can be removed.

pallets/replace/src/benchmarking.rs
```diff
@@ -151,8 +156,15 @@ benchmarks! {
 		VaultRegistry::<T>::issue_tokens(&vault_id, &amount).unwrap();
 		VaultRegistry::<T>::try_increase_to_be_replaced_tokens(&vault_id, &amount).unwrap();
 
-		// TODO: check that an amount was actually withdrawn
+		let vault : DefaultVault::<T> = VaultRegistry::<T>::get_vault_from_id(&vault_id).expect("should return a vault");
```
b-yap, line 159: as stated in t[his comment](https://github.com/pendulum-chain/spacewalk/issues/417#issuecomment-1943764696): 
added a `verify` to check that the `to_be_replaced_tokens` was updated.

testchain/node/src/cli.rs
```diff
@@ -10,7 +10,7 @@ pub struct Cli {
 }
 
 #[derive(Debug, clap::Subcommand)]
-#[allow(clippy::large_enum_variant)] //todo: fix large size difference between variants
```
b-yap, line 13: i'll assume that we're going to just allow this.

clients/runtime/Cargo.toml
ebma: What do you mean with 'testing ONE integration test case will work'? Do these 'testing-utils' import solve the problem you linked to?

clients/vault/src/oracle/testing_utils.rs
ebma: I think for these functions, we also need two different secret keys for the mainnet tests. I think we need one for the 'source' account and another one for the 'destination' at least in the integration tests.

clients/wallet/src/resubmissions.rs
```diff
@@ -245,48 +322,74 @@ impl StellarWallet {
...
-					}
+			// if the sequence number is greater than this response,
+			// no other transaction will ever match with it.
+			if tx_sequence_num > source_account_sequence {
+				break
+			}
+
+			// if the sequence number is greater than the last response of the iterator,
+			// then the transaction is possibly in THIS page
+			if tx_sequence_num >= source_account_sequence.saturating_sub(default_page_size) {
```
ebma, line 370: This cannot be reached with the previous if statement. I think we should switch the order of the two `if` statements.

clients/wallet/src/resubmissions.rs
```diff
@@ -193,6 +193,83 @@ impl StellarWallet {
...
+///  * if the sequence number of the transaction is > than what we're looking for, then update the
+///    iterator by removing the first half of the list;
+///  * else remove the last half of the list
+///
+/// # Arguments
+///
+/// * `iter` - the iterator to iterate over a list of `TransactionResponse`
+/// * `tx` - the transaction we want to confirm if it's already submitted
+/// * `public_key` - the public key of the wallet
+fn _is_transaction_already_submitted(
```
ebma, line 231: Some checks in this function are duplicated with `is_transaction_already_submitted&#40;&#41;`. Wouldn't it be enough to have these checks in one of the two functions?

ebma (commented):
Great changes @b-yap, thanks for taking the time to clean this all up 👍 
I left just a few minor comments. As far as I understand you improved that resubmission check with some kind of variation of a binary search which should speed it up indeed.

clients/vault/src/system.rs
b-yap: here: https://github.com/interlay/interbtc-clients/commit/622f36b5efd538c2011ecfebce2425de6be17388

clients/runtime/Cargo.toml
b-yap: when we wanted to test only 1 case: 
```
cargo test test_redeem_succeeds_on_mainnet -p vault --features standalone-metadata,integration
```

It goes inside the 
```rust
if #[cfg(not(feature = "testing-utils"))] {
```
block of code.

For 
```
cargo test --all --all-features
```
it  doesn't. 

My theory is: when running a _singular test case_ in the **`vault`**, it doesn't make use of the `runtime`'s _`dev-dependencies`_. But the `runtime` clearly specifies to use the _`testing-utils`_ feature of **`testchain-runtime`**.
https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/clients/runtime/Cargo.toml#L71


The `vault`'s `dev-dependencies` calls for the _`testing-utils`_ feature of the `runtime`. But as you can see in `runtime`, its _`testing-utils`_ feature is using the defaults of `testchain-runtime` and `oracle`.
That's why it got confusing why the `cfg-if` didn't work in `testchain-runtime`.

clients/wallet/src/resubmissions.rs
b-yap: i updated the code so it  does not "assume" the last element; I MADE it check the last element.

clients/wallet/src/resubmissions.rs
b-yap: got it in 1 function: ` fn _check_transaction_match(...)` though the return datatype is quite confusing, so I added comments on the function; 
including the ff functions: 
* `fn check_last_transaction_match(...)`
* `check_middle_transaction_match(...)`

It is is some kind of binary search; but the checking of the last element plays a role: whether the transaction can be found on the next page or not.

clients/vault/tests/helper/mod.rs
```diff
@@ -29,9 +31,8 @@ pub type StellarPublicKey = [u8; 32];
 
 lazy_static! {
 	pub static ref CFG: StellarOverlayConfig = random_stellar_relay_config(false);
-	pub static ref SECRET_KEY: String = get_test_secret_key(false);
-	// TODO clean this up by extending the `get_test_secret_key()` function
-	pub static ref DESTINATION_SECRET_KEY: String = "SDNQJEIRSA6YF5JNS6LQLCBF2XVWZ2NJV3YLC322RGIBJIJRIRGWKLEF".to_string();
+	pub static ref SECRET_KEY: String = get_testnet_secret_key(true);
```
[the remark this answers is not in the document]
b-yap: i updated the function from 2 to only 1:
```rust
pub fn get_secret_key(with_currency: bool, is_mainnet: bool) -> String {
	let suffix = if with_currency { "_with_currency" } else { "" };
	let directory = if is_mainnet { "mainnet" } else { "testnet" };

	let path = format!("./resources/secretkey/{directory}/stellar_secretkey_{directory}{suffix}");
	std::fs::read_to_string(path).expect("should return a string")
}

```

ebma (approved):
The return types of the helper functions are not ideal but alternatively we'd have to create enum types for the return values which is overkill just for these helpers so I'm fine if we keep it like this. The doc comments you put are great and should be good enough for this. 

Can we close [this](https://github.com/pendulum-chain/spacewalk/issues/277) ticket once this PR is merged? I saw you changed some things about the `Slot` after all.

ebma:
The ubuntu CI passed so we can already merge this one though, can't we @b-yap?

Issue #153, opened 2022-12-20
Add `clippy --fix` to pre-commit script


ebma:
It might be a good idea to add a `clippy --fix` command to the existing pre-commit script. It can automatically apply some best practices to the code, which the CI is likely to complain about anyways.

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/153?planning-poker) with Zenhub @b-yap @ebma @TorstenStueber

prayagd:
@ebma Do we need this for the public launch ?

ebma:
No, it's just for our own developer experience. It's not necessary for the product.

TorstenStueber:
@ebma can we move this to the Ready column?

TorstenStueber:
@ebma can we move this to the Ready column?

PR #488, opened 2024-02-09
Adding cargo fix to pre-commit hook

gianfra-t:
Regarding #153 

I will leave this as a draft pull request since I am not totally convinced about the results.

### What is added
We add `cargo clippy --fix` command to run over the whole project after every commit. 

Since there may be files that were modified but not staged, we will not add them to the commit, although they will be fixed by `clippy`.

### What could improve
Ideally, we would just run `fix` only on the modified files like we do with `rustfmt` but this is not possible with `clippy`.

Another possibility that was tested was to identify the package to which the modified file belongs and run `clippy fix` on the package level. Although this works, there is a problem to identify which flags each crate requires. Passing `--all-features` to every package fails on some of them.

### Alternatively
Since it takes quite some time for the compilation to take place, it perhaps is better to perform `clippy fix` on a `pre-push` hook.

### DIFF
--- a/.maintain/add-precommit-hook.sh
+++ b/.maintain/add-precommit-hook.sh
@@ -29,6 +29,36 @@ if [ ${#rust_files[@]} -ne 0 ]; then
 else
     echo "No changes, formatting skipped"
 fi
+wait
+
+echo "Running cargo clippy --fix on project"
+before_clippy=$(git diff --name-only)
+$(command -v cargo) clippy --fix --allow-dirty --allow-staged --all-features
+after_clippy=$(git diff --name-only)
+wait
+
+# Files modified by clippy
+readarray -t before_files <<< "$before_clippy"
+readarray -t after_files <<< "$after_clippy"
+clippy_modified_files=()
+for i in "${after_files[@]}"; do
+    skip=
+    for j in "${before_files[@]}"; do
+        [[ $i == $j ]] && { skip=1; break; }
+    done
+    [[ -n $skip ]] || clippy_modified_files+=("$i")
+done
+
+# Add only the files modified by clippy
+if [ ${#clippy_modified_files[@]} -ne 0 ]; then
+    git add ${clippy_modified_files[@]}
+    echo "Clippy modifications added for: ${clippy_modified_files[@]}"
+else
+    echo "No modifications by clippy"
+fi
+
+echo "Pre-commit hook execution completed."
+
 EOF
 
 chmod +x .git/hooks/pre-commit


ebma:
Thanks for the detailed description @gianfra-t. Although it's just a draft it might have been best to either add all 'devs' as reviewers or ping them in the comments so that this PR doesn't get missed. 

About the changes itself, you mention that you are not convinced about the results. What would you consider a dealbreaker here? I agree it's not ideal that apparently it's not possible to run clippy only on the modified files but I think it's fine if it's also run on files that were not modified. Eventually all Rust files in the project should adhere to the same formatting. And your suggestion to add this to a 'pre-push' hook sounds reasonable.

gianfra-t:
> What would you consider a dealbreaker here?

At least in my computer it just takes a LOT of time for each commit, and I think it may be a bit annoying. To be honest, I would not really activate this hook as is. But I guess it is a matter of personal preference! Let me know if you prefer to try this as is, or attempt the pre-push alternative.

TorstenStueber:
I also think that pre-push is the better solution.

.maintain/add-precommit-hook.sh
```diff
@@ -29,6 +29,36 @@ if [ ${#rust_files[@]} -ne 0 ]; then
...
+
+echo "Running cargo clippy --fix on project"
+before_clippy=$(git diff --name-only)
+$(command -v cargo) clippy --fix --allow-dirty --allow-staged --all-features
+after_clippy=$(git diff --name-only)
+wait
+
+# Files modified by clippy
+readarray -t before_files <<< "$before_clippy"
+readarray -t after_files <<< "$after_clippy"
```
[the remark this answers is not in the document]
gianfra-t: Oh, I will try a fix and test it on my laptop also! Thanks for pointing out.

TorstenStueber (commented):
Can you please also update the [README file](https://github.com/pendulum-chain/spacewalk/blob/325079f0c2a5560b7caf30ec872c1052f9c5b4e9/README.md?plain=1#L105)

I assume that `.maintain/add-prepush-hook.sh` needs to be run manually once? Can you please also add this to the README?

README.md
```diff
@@ -95,16 +95,27 @@ cargo run --bin spacewalk-standalone --release -- --dev
...
+To ease the maintenance of proper code formatting and good practices, please consider installing the pre-commit hook and/or the pre-push hook contained in this
 repository.
-This hook runs `rustfmt` on all the changed files in a commit, to ensure that changed files are properly formatted
+
+The pre-commit hook runs `rustfmt` on all the changed files in a commit, to ensure that changed files are properly formatted
 before committing.
 You can install the hook by running the following commands.
 
 ```
 .maintain/add-pre-commit.sh
```
ebma, line 106: This was already wrong before but let's fix it. 
```suggestion
.maintain/add-precommit-hook.sh
```

README.md
```diff
@@ -95,16 +95,27 @@ cargo run --bin spacewalk-standalone --release -- --dev
...
 ```
 
+The pre-push hool runs clippy checks that are also performed in the CI of the repository. These checks need to be successful so the push actually happens.
+Otherwise, pelase run the corresponding clippy fix command or manually fix the issue.
+
+To install the hook, run:
+```
+.maintain/add-prepush-hook.sh
```
ebma, line 114: I tested this locally and when I wanted to push, I got the message
```
hint: The '.git/hooks/pre-push' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

so I made it executable with `chmod u+x .git/hooks/pre-push`. Maybe we should mention this here as well?

.maintain/add-prepush-hook.sh
```diff
@@ -0,0 +1,22 @@
...
+# write a whole script to pre-push hook
+# NOTE: it will overwrite pre-push file!
+cat > .git/hooks/pre-push <<'EOF'
+#!/bin/bash
+
+echo "Running clippy checks"
+
+set +e
+
+cargo clippy --all-features -- -W clippy::all -A clippy::style -A clippy::forget_copy -A clippy::forget_ref
```
ebma, line 12: Let's add both checks we have in the ci workflow, ie. [this](https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/.github/workflows/ci-main.yml#L84) and [this](https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/.github/workflows/ci-main.yml#L91) one.

README.md
ebma: > Did you have to do the same with the pre-commit one right? 

I don't think so, was also confused that I had to do it this time. But the script in `.maintain` itself was working fine, I just had to make the generated script executable. Maybe it's different between macOS and linux? Not sure where you tested it.

ebma (approved):
Looking great, thanks 👍

bogdanS98 (approved):

README.md
gianfra-t: Apparently it was needed in macOS but not on linux!

PR #490, opened 2024-02-13
[WIP] change price conversion function in oracle pallet to work for assets with different decimals

ebma:
The new conversion function is in place but I still need to clean things up.


pallets/oracle/src/lib.rs
```diff
@@ -310,28 +322,91 @@ impl<T: Config> Pallet<T> {
...
 
-		// The code below performs `amount/rate`, plus necessary type conversions
-		Ok(T::UnsignedFixedPoint::checked_from_integer(amount)
-			.ok_or(Error::<T>::TryIntoIntError)?
-			.checked_div(&rate)
-			.ok_or(ArithmeticError::Underflow)?
-			.truncate_to_inner()
-			.ok_or(Error::<T>::TryIntoIntError)?
-			.unique_saturated_into())
+		let from_amount = T::UnsignedFixedPoint::checked_from_integer(from_amount)
```
TorstenStueber, line 368: Could we run into overflows here?

This will multiply `from_amount` with 10^18. Depending on the amount of decimals of the currency (I think it should not be more than 18 decimals), we could have therefore already 36 decimals at this point for the `from_amount`. Let's say, we want to safely support amounts of at least 1 million (we should of course support more), then this would be the number 10 ** 42.

This will then be bigger than the maximum value of `u128`, which is about 10 ** 39.

I think we should get the same numerical accuracy here without the danger of overflows by just using `from_inner` on this line and `into_inner` on line 385 and 404.

primitives/src/lib.rs
```diff
@@ -498,26 +498,74 @@ pub enum CurrencyId {
...
+pub type DefaultDecimalsLookup = PendulumDecimalsLookup;
+
+pub struct PendulumDecimalsLookup;
+impl DecimalsLookup for PendulumDecimalsLookup {
+	type CurrencyId = CurrencyId;
+
+	fn decimals(currency_id: CurrencyId) -> u32 {
+		(match currency_id {
 			CurrencyId::Stellar(asset) => asset.decimals(),
+			CurrencyId::XCM(index) => match index {
```
TorstenStueber, line 522: I did not check whether this is correct.

pallets/oracle/src/lib.rs
gianfra-t: This is a very good point. I would have tried to avoid using directly `into_inner` as an intermediate representation because it may be confusing to have everything divided by `10**18`, but it is the only way to achieve representing as large values as `Balance` can. 

I wonder, is it any benefit to use `UnsignedFixedPoint` here instead of using the inner (u128) directly? I understand this may be in part because the prices are given by the oracle as `UnsignedFixedPoint` already, but perhaps an alternative is to convert the prices to `Balance` type first.

TorstenStueber: The benefit of `UnsignedFixedPoint` is that it uses a safe multiplication function because it first converts all numbers to some 256 bit number, then multiplies them, then divides by 10 ** 18, then truncates to 128 bit. This way no overflow for the intermediate value will occur.

We would probably want to use that when multiplying with the price as that one also uses 18 decimals (we could of course leave it at 12 decimals as reported by DIA if we don't need to widen it to `UnsignedFixedPoint`, but it is probably still safer this way).

gianfra-t: Okay, I didn't know it converted the intermediate number during multiplication to a larger space. Then it makes sense!

TorstenStueber: If I am not mistaken: [`checked_mul`](https://github.com/paritytech/substrate/blob/98f2e3451c9143278ec53c6718940aeabcd3b68a/primitives/arithmetic/src/fixed_point.rs#L855-L868) calls [`multiply_by_rational_with_rounding`](https://github.com/paritytech/substrate/blob/98f2e3451c9143278ec53c6718940aeabcd3b68a/primitives/arithmetic/src/helpers_128bit.rs#L187-L215), which calls [`product_of` of `Double128`](https://github.com/paritytech/substrate/blob/98f2e3451c9143278ec53c6718940aeabcd3b68a/primitives/arithmetic/src/helpers_128bit.rs#L125-L147), which returns a 256 bit value.

primitives/src/lib.rs
```diff
@@ -498,26 +498,74 @@ pub enum CurrencyId {
...
+				_ => 12,
+			},
+			// We assume that all other assets have 12 decimals
+			CurrencyId::Native | CurrencyId::ZenlinkLPToken(_, _, _, _) | CurrencyId::Token(_) =>
+				12,
+		}) as u32
 	}
+}
+
+impl CurrencyId {
```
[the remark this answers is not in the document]
ebma: Which `pub fn one()` are you referring to? As far as I can tell, [this](https://github.com/pendulum-chain/spacewalk/blob/2c9fe37320dfe707007a364a30792113993a9662/primitives/src/lib.rs#L481) is the only one left and it should be fine to keep it.

pallets/oracle/src/lib.rs
ebma: Thanks for pointing it out @TorstenStueber. You are right that this change makes us support handling amounts up to 10**18 times larger. I [created a new test case](https://github.com/pendulum-chain/spacewalk/pull/490/commits/e427fbd43f55766dfecd95fbcbb44225b161a2bc) to confirm that the conversion works in the expected range of amounts.

primitives/src/lib.rs
ebma: I adopted the decimals from the transactions we created for the asset metadata [here](https://github.com/pendulum-chain/transaction-builder/blob/9c8be6de9b9e0b92ad6150e8b8ee63095894800a/transactions/amplitude_transfer21.js#L74) and [here](https://github.com/pendulum-chain/transaction-builder/blob/9c8be6de9b9e0b92ad6150e8b8ee63095894800a/transactions/pendulum_transfer20.js#L17).

primitives/src/lib.rs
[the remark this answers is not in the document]
TorstenStueber: Strange, I think I confused this.

pallets/vault-registry/src/tests.rs
```diff
@@ -1094,9 +1096,18 @@ fn get_required_collateral_for_wrapped_with_threshold_succeeds() {
...
+			// wrapped may be rounded. So we give 10^(wrapped_decimals - collateral_decimals) as
+			// tolerance.
+			let tolerance = 10u128.pow(
+				<Test as oracle::Config>::DecimalsLookup::decimals(DEFAULT_WRAPPED_CURRENCY)
+					.saturating_sub(<Test as oracle::Config>::DecimalsLookup::decimals(
+						DEFAULT_COLLATERAL_CURRENCY,
+					)),
+			);
+			assert!(max_wrapped_for_min_collateral.amount() + tolerance >= wrapped_amount);
+			assert!(max_wrapped_for_below_min_collateral.amount() < wrapped_amount);
```
gianfra-t, line 1110: Why don't we need the tolerance in the second assertion?

gianfra-t (commented):
This all looks great @ebma. I like the use of the `DecimalsLookup` type, I hope we can keep up with updating it properly.
Just left some comments that are more like questions so I understand fully.

pallets/issue/src/tests.rs
```diff
@@ -447,14 +447,24 @@ fn test_set_issue_period_only_root() {
...
+		let issue_asset = VAULT.wrapped_currency();
+		// First, convert the volume limit to the issue asset
+		let volume_limit_denoted_in_wrapped_asset =
+			Oracle::convert(&Amount::new(volume_limit, volume_currency), issue_asset)
+				.expect("Price conversion should work");
+
+		// We set the issue amount as the volume limit + 100 (we have to use at least 100 because
+		// the issue_amount might get rounded down during the price conversion between assets with
+		// decimals differing by 2)
+		let issue_amount = volume_limit_denoted_in_wrapped_asset.amount() + 100;
```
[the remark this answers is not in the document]
ebma: I think so. If the wrapped vs collateral assets that we use in the test cases differ by more than 2 decimals we'd probably encounter issues again. I guess the perfect solution would be something similar to how I calculate the tolerance [here](https://github.com/pendulum-chain/spacewalk/blob/e427fbd43f55766dfecd95fbcbb44225b161a2bc/pallets/vault-registry/src/tests.rs#L1103-L1108). Doing something similar everywhere would reduce readability but increase the robustness in case we change the decimals in the future. Since this is quite unlikely to happen, I'm a little hesitant but what do you think @gianfra-t?

pallets/oracle/src/lib.rs
```diff
@@ -57,10 +59,16 @@ pub mod types;
...
 pub mod oracle_api;
+
 pub use crate::oracle_api::*;
+
 #[cfg(feature = "testing-utils")]
 pub mod oracle_mock;
 
+// We assume this value to be the decimals of our base currency (USD). It doesn't matter too much as
+// long as it's consistent.
+const USD_DECIMALS: u32 = 12;
```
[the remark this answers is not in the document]
ebma: It's not even linked to how the prices are stored in the DIA oracle. This constant will just determine what we consider 1 unit of USD when calling either the `currency_to_usd()` or `usd_to_currency` function. Since going forward, these functions are only used by the exposed RPC API (because the `convert()` function itself doesn't call them anymore), the `USD_DECIMALS` are only relevant to interpret the result of the RPC call. We can set it to whatever we want without impacting any functionality. 

For example, assuming we keep the `USD_DECIMALS` as `12`, here we would interpret that 1 XLM is worth $ `0.000,000,116,964` (this was taken from the current setup where we have this 6 decimal shifting problem solved by the [other](https://github.com/pendulum-chain/pendulum/pull/413) ticket).
image

pallets/redeem/src/tests.rs
```diff
@@ -1426,9 +1442,9 @@ fn test_execute_redeem_fails_when_exceeds_rate_limit() {
 		);
 
 		let redeemer = USER;
-		// Requesting a second redeem request with just 1 unit of collateral should fail because we
+		// Requesting a second redeem request with a small unit of collateral should fail because we
 		// reached the volume limit already
-		let amount = 1;
+		let amount = 1_000;
```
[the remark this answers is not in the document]
ebma: Yes 👍

pallets/vault-registry/src/tests.rs
ebma: Good question. From empiric testing I can tell you that it is not problematic and the bound holds even without the tolerance. But it would make sense to add the tolerance both ways.

pallets/issue/src/tests.rs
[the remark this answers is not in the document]
gianfra-t: Alright ! I don't think it is necessary either to add the tolerance, just wanted to confirm my understanding. As long as we are aware it is fine.

pallets/oracle/src/lib.rs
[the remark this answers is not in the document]
gianfra-t: Understood. Thanks for the explanation. So although it is only relevant for these two RPC functions it would be nice to set it up so that such that the output here reflects the true usd value of the token. In this case when the other fix comes online, `12` decimals will not be enough for this I imagine right? Since the price is divided by 12 yet here we consider one unit of USD = 1, and not 1/10**12.

I tried to follow the calculation but got a little lost, so not sure if this is correct.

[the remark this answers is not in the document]
ebma: I'll try to explain again how this relates to the change in https://github.com/pendulum-chain/pendulum/pull/413.
By changing the FixedPoint creation to `from_rational()` instead of `from_inner()` in https://github.com/pendulum-chain/pendulum/pull/413, the exchange rates will be represented how they should be internally. So let's say, our DIA pallet has an on-chain price of `123,000,000,000` ($0.123) stored for XLM. Before the change (with `from_inner()`) the fixed point number would be `0.000000123`. With the change to `from_rational()` it will be `0.123`. 

Now, let's look at a call of `currencyToUsd()` converting from XLM to USD, assuming that we are using the 'proper' fixed point number of `0.123`. If we set `USD_DECIMALS` to `12`, the result of `currency_to_usd()` will be `123,000,000,000`. If we set `USD_DECIMALS` to `18`, the result will be `123,000,000,000,000,000`. If we set `USD_DECIMALS` to `9`, the result will be `123,000,000`. What I'm trying to illustrate is that the actual value of this parameter does not matter for the calculation but it matters for how we interpret the result.

bogdanS98 (approved):
Great changes 👍🏼 I already got the answers to my questions from the threads created by @gianfra-t

pallets/vault-registry/src/tests.rs
ebma: I added that tolerance also to the other check now

TorstenStueber (approved):
Looks all good to me now. Thanks!

PR #495, opened 2024-03-01
Improve robustness of Spacewalk integration tests

### LINKED ISSUE #494: Improve robustness of Spacewalk vault integration tests
The integration tests using Stellar mainnet are failing too often, probably when the network is very busy. We need to improve this because our CI needs to pass reliably. 

## TODO
- Check what's causing the mainnet integration tests to fail and try to prevent it. The logic of the tests overall should be fine as it's not always the same test failing.
- To do this, run `cargo test --package vault --features standalone-metadata --features integration-test --test vault_integration_tests ""`

ebma:
Closes #494.

The multiple mainnet integration issues (and its fixes) are as follows:
* cannot determine which `DecimalsLookUp` to use. 
   * fix: make sure for the feature `integration-test`, the clients/vault uses the `standalone-metadata` feature:
    https://github.com/pendulum-chain/spacewalk/blob/ea913fab5ae2a47368d48206fba5e7897e3c118d/clients/vault/src/lib.rs#L52-L53
    * fix: make sure the clients/runtime has access to the [mainnet-runtime](https://github.com/pendulum-chain/spacewalk/tree/main/testchain/runtime/mainnet)
* failed with
   ```
      HorizonResponseError(reqwest::Error { kind: Decode, source: Error("invalid type: string \"ySHX => SHX/XLM DAILY\", expected a borrowed string", line: 9484, column: 44) })
   ```
    * fix: Some memo texts have non-literal characters, messing up deserialization.  A solution is to ~[use `Cow<str>` instead](https://users.rust-lang.org/t/solved-serde-deserialize-str-containig-special-chars/27218/2)~ `String`.
* `'could not find event: Issue::ExecuteIssue' ` potentially because all transactions have been traversed (reached EOF) before the  _memo to check_ hashmap was filled. A.K.A. probably race condition.
  * fix: DO NOT ITERATE OVER the list of transactions UNLESS the hashmap has values inside.
     * introducing `IsEmptyExt` since the parameters being passed are using generic datatypes.
       ```rust
            pub trait IsEmptyExt {
	        fn is_empty(&self) -> bool;
            }
       ```
       https://github.com/pendulum-chain/spacewalk/blob/ea913fab5ae2a47368d48206fba5e7897e3c118d/clients/wallet/src/horizon/horizon.rs#L255-L257

### DIFF
--- a/clients/vault/src/oracle/agent.rs
+++ b/clients/vault/src/oracle/agent.rs
@@ -218,7 +218,7 @@ mod tests {
 	async fn test_get_proof_for_current_slot() {
 		// let it run for a few seconds, making sure that the other tests have successfully shutdown
 		// their connection to Stellar Node
-		sleep(Duration::from_secs(5)).await;
+		sleep(Duration::from_secs(2)).await;
 
 		let shutdown_sender = ShutdownSender::new();
 
@@ -236,6 +236,7 @@ mod tests {
 			sleep(Duration::from_secs(1)).await;
 			latest_slot = agent.last_slot_index().await;
 		}
+		latest_slot +=1;
 		// let's wait for envelopes and txset to be available for creating a proof
 		sleep(Duration::from_secs(5)).await;
 
--- a/clients/vault/src/oracle/agent.rs
+++ b/clients/vault/src/oracle/agent.rs
@@ -236,7 +236,7 @@ mod tests {
 			sleep(Duration::from_secs(1)).await;
 			latest_slot = agent.last_slot_index().await;
 		}
-		latest_slot +=1;
+		latest_slot += 1;
 		// let's wait for envelopes and txset to be available for creating a proof
 		sleep(Duration::from_secs(5)).await;
 
--- a/clients/vault/src/oracle/agent.rs
+++ b/clients/vault/src/oracle/agent.rs
@@ -218,7 +218,7 @@ mod tests {
 	async fn test_get_proof_for_current_slot() {
 		// let it run for a few seconds, making sure that the other tests have successfully shutdown
 		// their connection to Stellar Node
-		sleep(Duration::from_secs(20)).await;
+		sleep(Duration::from_secs(5)).await;
 
 		let shutdown_sender = ShutdownSender::new();
 
@@ -236,9 +236,8 @@ mod tests {
 			sleep(Duration::from_secs(1)).await;
 			latest_slot = agent.last_slot_index().await;
 		}
-		// use 1 slot ahead, to ensure enough messages can be collected
-		// and to avoid "missed" messages.
-		latest_slot += 1;
+		// let's wait for envelopes and txset to be available for creating a proof
+		sleep(Duration::from_secs(5)).await;
 
 		let proof_result = agent.get_proof(latest_slot).await;
 		assert!(proof_result.is_ok(), "Failed to get proof for slot: {}", latest_slot);
--- a/clients/vault/tests/vault_integration_tests.rs
+++ b/clients/vault/tests/vault_integration_tests.rs
@@ -603,7 +603,7 @@ async fn test_issue_cancel_succeeds() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-#[ntest::timeout(900_000)] // timeout at 15 minutes
+#[ntest::timeout(1_200_000)] // timeout at 20 minutes
 #[serial]
 async fn test_issue_execution_succeeds_from_archive() {
 	let is_public_network = true;
--- a/clients/vault/tests/vault_integration_tests.rs
+++ b/clients/vault/tests/vault_integration_tests.rs
@@ -603,6 +603,7 @@ async fn test_issue_cancel_succeeds() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+#[ntest::timeout(900_000)] // timeout at 15 minutes
 #[serial]
 async fn test_issue_execution_succeeds_from_archive() {
 	let is_public_network = true;
--- a/clients/vault/tests/vault_integration_tests.rs
+++ b/clients/vault/tests/vault_integration_tests.rs
@@ -793,7 +793,7 @@ async fn test_issue_overpayment_succeeds() {
 
 #[tokio::test(flavor = "multi_thread")]
 #[serial]
-async fn clear() {
+async fn test_automatic_issue_execution_succeeds_on_mainnet() {
 	let is_public_network = true;
 	test_automatic_issue_execution_succeeds_on_network(is_public_network).await;
 }
--- a/clients/vault/tests/vault_integration_tests.rs
+++ b/clients/vault/tests/vault_integration_tests.rs
@@ -853,7 +853,7 @@ async fn test_automatic_issue_execution_succeeds_on_network(is_public_network: b
 				.await
 				.expect("should return a result");
 
-				tracing::warn!("Sent payment successfully: {:?}", result);
+				tracing::info!("Sent payment successfully: {:?}", result);
 
 				// Sleep 5 seconds to give other thread some time to receive the RequestIssue event
 				// and add it to the set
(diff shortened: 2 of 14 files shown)


b-yap:
one test is failing with:
```
thread 'test_redeem_succeeds_on_mainnet' panicked at 'should return ok: HorizonSubmissionError { title: "Transaction Failed", status: 400, reason: "tx_failed", result_code_op: ["op_underfunded"], envelope_xdr: Some("AAAAAgAAAACne8I5bB4KhDOubC5DCq7O/p4yvnXbjNkOE8fT3v0MHwAAAGUC7t8CAAABLwAAAAAAAAABAAAAHDNCMWdEd2U1YTZUS2dFcHVmdmJZcmRCcVZWNHYAAAABAAAAAQAAAACne8I5bB4KhDOubC5DCq7O/p4yvnXbjNkOE8fT3v0MHwAAAAEAAAAArwuPZBi/hC2gPFK8R7thoU5Xsbh+k5n1ls8DrCkdhJMAAAABVVNEQwAAAAA7mRE4Dv6Yi6CokA6xz+RPNm99vpRr7QdyQPf2JN8VxQAAAAAAAYagAAAAAAAAAAHe/QwfAAAAQElB9VXUO8NynqXVQVrdXuAECle2Iebo48ciDX+d5T0sMq3kz7YVDQqcw3uvPftSmCtRXLDoP2sAleJBIcNopQg=") }', clients/vault/tests/helper/helper.rs:196:6
```

clients/vault/src/oracle/agent.rs
```diff
@@ -236,9 +236,8 @@ mod tests {
 			sleep(Duration::from_secs(1)).await;
 			latest_slot = agent.last_slot_index().await;
 		}
-		// use 1 slot ahead, to ensure enough messages can be collected
-		// and to avoid "missed" messages.
-		latest_slot += 1;
+		// let's wait for envelopes and txset to be available for creating a proof
+		sleep(Duration::from_secs(5)).await;
```
ebma, line 240: Let's hope that this finally fixes it.

clients/vault/tests/vault_integration_tests.rs
```diff
@@ -792,7 +793,7 @@ async fn test_issue_overpayment_succeeds() {
 
 #[tokio::test(flavor = "multi_thread")]
 #[serial]
-async fn test_automatic_issue_execution_succeeds_on_mainnet() {
+async fn clear() {
```
ebma, line 796: I doubt that this was intended?

clients/wallet/src/horizon/horizon.rs
```diff
@@ -259,11 +263,16 @@ impl<C: HorizonClient + Clone> HorizonFetcher<C> {
...
 
-		let mut txs_iter = self.fetch_transactions_iter(last_cursor).await?;
-
 		let (issue_map, memos_to_issue_ids) =
 			future::join(issue_map.read(), memos_to_issue_ids.read()).await;
 
+		if issue_map.is_empty() || memos_to_issue_ids.is_empty() {
+			tracing::info!("fetch_horizon_and_process_new_transactions(): nothing to traverse");
+			return Ok(last_cursor)
+		}
```
ebma, line 272: Nice catch :+1:

clients/wallet/src/horizon/responses.rs
```diff
@@ -319,10 +319,10 @@ pub struct HorizonBalance {
 	#[serde(deserialize_with = "de_string_to_f64")]
 	pub balance: f64,
 	#[serde(default)]
-	#[serde(deserialize_with = "de_string_to_optional_bytes")]
+	#[serde(deserialize_with = "de_str_to_optional_bytes")]
 	pub asset_code: Option<Vec<u8>>,
 	#[serde(default)]
-	#[serde(deserialize_with = "de_string_to_optional_bytes")]
+	#[serde(deserialize_with = "de_str_to_optional_bytes")]
```
ebma, line 325: So we don't want to use `String` here? Or do we? Because I saw you added a new serde function but it's not used here.

clients/vault/tests/vault_integration_tests.rs
b-yap: :scream:

clients/wallet/src/horizon/responses.rs
b-yap: Maybe not. 

The issue was on deserializing special characters, which the `memo` contains now and then:
```
\"ySHX => SHX/XLM DAILY\"
```
see my commit description: https://github.com/pendulum-chain/spacewalk/pull/495/commits/e6aaa8393ac56c024394eef1b88129b4d6e39c50

ebma: Sorry, I was a little confused because just looking at the changed lines it was not clear to me how this applies to the memo as mentioned in the PR description. But now I realized that these changed lines here are not about the memo, but other fields like e.g. [this](https://github.com/pendulum-chain/spacewalk/blob/3c68672cc6e99ddf693604a6649e33c8cf8c71ca/clients/wallet/src/horizon/responses.rs#L198) one are now using a `String` instead of the `str` which was misleading in the function name.

Makes sense then 👍  but why do we need to use `str` for asset code and issuer? Or is it just because we know that there cannot be any weird characters?

b-yap: I did not want to change how the other fields are deserialized. 
Maybe it's even best that we prevent such special characters in these fields.

b-yap: see https://github.com/pendulum-chain/spacewalk/pull/495/commits/c9fe6a4a5129776f258300b32ab57f3b53424e3e

ebma: Yes, makes sense 👍

ebma:
Now the CI job failed with 
```
thread 'test_redeem_succeeds_on_mainnet' panicked at 'should return ok: HorizonSubmissionError { title: "Transaction Failed", status: 400, reason: "tx_insufficient_fee", result_code_op: [], 
```

which is weird because recently already changed the default fee percentile to be `95` instead of `90`. We could set it to `100` but then we potentially pay a very large fee :(

b-yap:
@ebma  
I think it might do us good if we have more than 2 secret keys for testing; Like 5 keys.

ebma:
It might help yes. But does it make sense? Shouldn't in a perfect world our tests be able to just work with one set of accounts instead of us having to rely on switching them randomly?

b-yap:
you are right. :/

clients/vault/src/oracle/agent.rs
gianfra-t: This sleep is ultimately for `poll_messages_from_stellar` [process](https://github.com/pendulum-chain/spacewalk/blob/main/clients/stellar-relay-lib/src/connection/connector/message_reader.rs#L13C21-L13C47) to have time for writing enough messages into `ScpMessageCollector`?

b-yap: Yes, to [build the proof](https://github.com/pendulum-chain/spacewalk/blob/main/clients/vault/src/oracle/agent.rs#L158-L185).

clients/wallet/src/stellar_wallet.rs
```diff
@@ -99,6 +99,14 @@ impl StellarWallet {
...
 
 		let cache = WalletStateStorage::new(cache_path, &pub_key, is_public_network);
+		// using a builder to decrease idle connections
+		// https://users.rust-lang.org/t/reqwest-http-client-fails-when-too-much-concurrency/55644/2
+		let client = reqwest::Client::builder()
+			// default is 90 seconds.
+			.pool_idle_timeout(Some(Duration::from_secs(60)))
+			// default is usize max.
+			.pool_max_idle_per_host(usize::MAX / 2)
+			.build()?;
```
ebma, line 109: Did you notice a meaningful difference with this change?

ebma (commented):
It looks good to me. Do you think you found the problem with macOS or was it more or less a coincidence that it passed this time?

clients/wallet/src/stellar_wallet.rs
b-yap: To be completely honest, I am unsure. 

However we encountered idle connections before and even talked about using `ClientBuilder`. I guess now's the right time.

I set the pool to half of default because [zero](https://users.rust-lang.org/t/reqwest-http-client-fails-when-too-much-concurrency/55644/5) is too huge a leap.

b-yap:
@ebma For macos it's definitely a coincidence. 
I'll be keeping my eye on the mainnet tests.

ebma:
I know I already brought this up a very long time ago, but how about we just remove the macOS job from the workflow completely? I don't see much of a benefit here, besides our workflows taking twice the time and being more likely to fail in some sense. Of course, removing something just because it doesn't work is not the ideal solution but generally, even assuming that the macOS job worked reliably all the time, I'd argue that vault operators would deploy these vaults in a Linux environment and not on a macOS machine. 

WDYT @b-yap? Do you want to keep it or should we remove it?

b-yap:
@ebma 
I'd vote on removing it.

ebma:
Let's remove it and merge this without waiting for the checks

b-yap (approved):

PR #500, opened 2024-03-27
499 make vault retry stellar transaction submission for tx insufficient fee error

### LINKED ISSUE #499: Make vault retry Stellar transaction submission for `tx_insufficient_fee` error
When the vault client submits a Stellar payment transaction to fulfill a redeem request, it might encounter a `tx_insufficient_fee` error because the fee prices surged from one ledger to another. This means the transaction is rejected and the vault needs to create a new payment with a higher fee. This retry is currently not done and the submission is only retried when the vault performs its periodic restart ie. every 3 hours.

## TODO

We need to change the logic so that the payment is retried with a new transaction and a higher fee when the `tx_insufficient_fee` response is returned by Stellar horizon.

ebma:
This PR adds handling for 'tx_insufficient_fee' errors. 

We create a new transaction with a fee that is 10 times higher than the previous one. This 10x factor is adopted from the requirement of FeeBump transactions, see [here](https://developers.stellar.org/docs/learn/encyclopedia/fees-surge-pricing-fee-strategies#fee-bumps-on-past-transactions). We **cannot** use fee bump transactions however, as we did not add proper handling for them in the stellar-relay pallet [here](https://github.com/pendulum-chain/spacewalk/blob/bda384bcc6db3da88198c714851a261dcdea1063/pallets/currency/src/lib.rs#L135). Using fee bump transactions here is not necessary though as they were introduced with a different purpose in mind. For more information see [this](https://stellar.org/blog/developers/how-we-simplified-fees-on-the-stellar-network) blog post. 

The increased fee will not go beyond the maximum of 1 XLM. This amount was chosen arbitrarily and we could think about making this configurable so that each vault operator can choose this value but it should be fine not to make this configurable for the time being. 

Closes #499.

### DIFF
--- a/clients/wallet/src/horizon/horizon.rs
+++ b/clients/wallet/src/horizon/horizon.rs
@@ -267,7 +267,7 @@ impl<C: HorizonClient + Clone> HorizonFetcher<C> {
 			future::join(issue_map.read(), memos_to_issue_ids.read()).await;
 
 		if issue_map.is_empty() || memos_to_issue_ids.is_empty() {
-			tracing::info!("fetch_horizon_and_process_new_transactions(): nothing to traverse");
+			tracing::debug!("fetch_horizon_and_process_new_transactions(): nothing to traverse");
 			return Ok(last_cursor)
 		}
 		let mut txs_iter = self.fetch_transactions_iter(last_cursor).await?;
--- a/clients/wallet/src/horizon/responses.rs
+++ b/clients/wallet/src/horizon/responses.rs
@@ -181,8 +181,8 @@ pub struct TransactionResponse {
 	pub fee_account: Vec<u8>,
 	#[serde(deserialize_with = "de_str_to_u64")]
 	pub fee_charged: u64,
-	#[serde(deserialize_with = "de_str_to_bytes")]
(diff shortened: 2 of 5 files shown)


b-yap (approved):
looks good.

TorstenStueber:
Great PR, this will be a big improvement for the UX. 

In theory the logic could be caught in a cycle where the fee is already at the 1 XLM maximum and will be resubmitted repeatedly. Not sure whether such a scenario will ever happen. But anyway: would that be a problem for the vault?

ebma:
> In theory the logic could be caught in a cycle where the fee is already at the 1 XLM maximum and will be resubmitted repeatedly. Not sure whether such a scenario will ever happen. But anyway: would that be a problem for the vault?

Yes, that's true but it should not be a problem. As you said, the vault will continue resubmitting the transaction with a fee of 1 XLM but assuming that the load on the Stellar network fluctuates, the vault will manage to submit that transaction eventually. If the traffic on the Stellar network ever gets so high that _every_ transaction needs to pay at least 1 XLM to go through, we would definitely need to revisit this change again though.

PR #502, opened 2024-03-29
501 fix occassional horizonresponse decodeerror

### LINKED ISSUE #501: Fix occasional 'HorizonResponse: DecodeError'
Sometimes when the vault client is trying to submit a Stellar transaction as part of redeem requests, the horizon responds with a lot of timeout errors. It seems like at some point, this timeout error changes and the horizon server instead returns a different response that the vault client fails to decode. The vault client then stops trying to execute the redeem request until the next periodic restart.

Example 1: Logs of a Spacewalk client
```
Mar 27 10:37:41.221 ERROR transfer_stellar_asset{request_type=Redeem request_id=0xb93fbb834856873b5afae7841249305eb58557bc10d3f7b2cf2f941591502548}: wallet::horizon::responses: Response returned an error: HorizonSubmissionError { title: "Timeout", status: 504, reason: "Your request timed out before completing.  Please try your request again. If you are submitting a transaction make sure you are sending exactly the same transaction (with the same sequence number).", result_code_op: [], envelope_xdr: None }
Mar 27 10:37:41.221  WARN transfer_stellar_asset{request_type=Redeem request_id=0xb93fbb834856873b5afae7841249305eb58557bc10d3f7b2cf2f941591502548}: wallet::horizon::horizon: submitting transaction to https://horizon.stellar.org with seq no: Some(197273654701064434) failed with HorizonSubmissionError { title: "Timeout", status: 504, reason: "Your request timed out before completing.  Please try your request again. If you are submitting a transaction make sure you are sending exactly the same transaction (with the same sequence number).", result_code_op: [],envelope_xdr: None }
Mar 27 10:38:32.111 ERROR transfer_stellar_asset{request_type=Redeem request_id=0xb93fbb834856873b5afae7841249305eb58557bc10d3f7b2cf2f941591502548}: wallet::horizon::responses: Response returned an error: HorizonSubmissionError { title: "Timeout", status: 504, reason: "Your request timed out before completing.  Please try your request again. If you are submitting a transaction make sure you are sending exactly the same transaction (with the same sequence number).", result_code_op: [], envelope_xdr: None }
Mar 27 10:38:32.111  WARN transfer_stellar_asset{request_type=Redeem request_id=0xb93fbb834856873b5afae7841249305eb58557bc10d3f7b2cf2f941591502548}: wallet::horizon::horizon: submitting transaction to https://horizon.stellar.org with seq no: Some(197273654701064434) failed with HorizonSubmissionError { title: "Timeout", status: 504, reason: "Your request timed out before completing.  Please try your request again. If you are submitting a transaction make sure you are sending exactly the same transaction (with the same sequence number).", result_code_op: [],envelope_xdr: None }
Mar 27 10:39:32.149 ERROR vault::redeem: Failed to process Redeem request #0xb93f…2548 due to error: StellarWalletError: Error fetching horizon data: error decoding response body: expected value at line 1 column 1
```

Example 2: Logs of [this](https://github.com/pendulum-chain/spacewalk/actions/runs/8465174747/job/23191282363?pr=500#step:9:3787) failing CI job:
```
thread 'test_redeem_succeeds_on_mainnet' panicked at 'should return ok: HorizonResponseError(reqwest::Error { kind: Decode, source: Error("expected value", line: 1, column: 1) })', clients/vault/tests/helper/helper.rs:196:6
```

## TODO

Make sure that the vault client does not stop retrying the submission of the Stellar transaction for redeem requests.

b-yap:
closes #501 

The error found was _403_, and as explained by @ebma, it should be recoverable.
Not only that, I've updated the `HorizonResponseError` structure, flexible enough to accept either a `ReqwestError` or an error in string format.

The string format is important to find the correct error.


clients/wallet/src/horizon/responses.rs
```diff
@@ -46,14 +46,28 @@ macro_rules! debug_str_or_vec_u8 {
 pub(crate) async fn interpret_response<T: DeserializeOwned>(
 	response: reqwest::Response,
 ) -> Result<T, Error> {
-	if response.status().is_success() {
-		return response.json::<T>().await.map_err(Error::HorizonResponseError)
+	let status = response.status();
+	tracing::info!("interpret_response(): status: {status:?}");
```
ebma, line 50:
```suggestion
	tracing::debug!("interpret_response(): status: {status:?}");
```

clients/wallet/src/error.rs
```diff
@@ -8,8 +8,12 @@ use thiserror::Error;
 pub enum Error {
 	#[error("Invalid secret key")]
 	InvalidSecretKey,
-	#[error("Error fetching horizon data: {0}")]
-	HorizonResponseError(#[from] ReqwestError),
+	#[error("Error fetching horizon data: reqwest: {reqwest:?}, other: {other:?}")]
+	HorizonResponseError {
+		reqwest: Option<ReqwestError>,
```
ebma, line 13: Maybe we should rename `reqwest` to `error` because just `reqwest` somehow suggests to me that it's the client or a response and not an error. 

```suggestion
		error: Option<ReqwestError>,
```

b-yap:
@ebma by adding this error as "recoverable", it will resend the same transaction again. 
see https://github.com/pendulum-chain/spacewalk/blob/501-fix-occassional-horizonresponse-decodeerror/clients/wallet/src/horizon/horizon.rs#L167.

clients/wallet/src/error.rs
b-yap: renamed to `error`

ebma:
Yes, the logic makes sense to me. I was asking if you tested it because it would actually be interesting to see how this behaves in practice. Because if this error is indeed somehow related to rate limiting, our logic here is not enough to make the transaction eventually go through. Since even if we retry with the same transaction over and over again, the rate-limiting would continue to block it. _If_ this is the case, which I don't know, then we should add some kind of delay here and wait for the rate-limiting window to expire.

b-yap:
Yesterday it was difficult with the accounts not existing.

ebma:
Ahhh true, we already have that delay in place 👍 That's great. 

I found out that the Horizon's [apparently](https://developers.stellar.org/network/horizon/structure/rate-limiting) have IP-based rate limiting of 3600 requests per hour ie. one request per second on average.

ebma (approved):
Let's give these changes a try in production. In any case they won't do any harm and at least improve the decoding of some horizon responses, even if we are not 100% sure whether submissions that returned a 403 error will eventually get accepted or not.

ebma:
Only one integration test failed so we can consider it a shaky test. I checked cargo clippy and rustfmt locally and there are no complaints. Merging.

PR #504, opened 2024-04-01
471 clean up stellar secret keys

### LINKED ISSUE #471: Clean up Stellar secret keys
We currently have some secrets for Stellar accounts lingering around in this repository. These are required for the successful execution of many integration tests. However it's not ideal to have these secrets available in this public repository and we need to clean this up. 

## TODO
- Create a config that holds the required Stellar account secrets required by all tests. We need at least 4 different account secrets for both public and test network in total because we need to have a source account and a destination account on each network. These are used when testing transfers. 
- Pass the necessary account secrets as environment variables instead. It's very easy to add a configuration that is populated with env variables using `clap`, see [here](https://github.com/interlay/interbtc-clients/blob/93a3330a11f267054aebeca689e224afae147a93/bitcoin/src/cli.rs#L23). 
- Replace all occurrences of hard-coded Stellar secret keys with keys passed to the config.
- Change the settings for the CI environment variables of this repository so that the keys are available for the tests. They also need to be provided in the Gitlab CI that is run for new releases, so reach out to @zoveress when working on this.

As part of this change, we should create new Stellar accounts and move the respective funds to those accounts. This way, we are sure that our testing accounts were never exposed publicly. This will also require changes to the script for topping up the testnet account [here](https://github.com/pendulum-chain/stellar-testnet-restoration).

gianfra-t:
Closes #471

Makes use of environmental variables for the stellar secret keys instead of reading them from the `resources` files, when running tests. We don't use `clap` for passing the keys given that is not straightforward to use with `cargo test`.

We maintain the usage of the same cli commands where we specify the `stellar_vault_secret_key_filepath`, but use the environment keys for testing.

The use of `DEFAULT_MAINNET_DEST_SECRET_KEY` and `DEFAULT_TESTNET_DEST_SECRET_KEY` is now replaced by destination keys read from environment. In order to keep the use of only 4 consistent source and destination keys, the keys previously supplied from by these constants are changed.

NOTE: the tests should be failing until we add the github secrets corresponding to the stellar keys.


clients/vault/src/oracle/testing_utils.rs
```diff
@@ -43,12 +44,11 @@ fn stellar_relay_config_abs_path(
 		.expect("should be able to extract config")
 }
 
-pub fn get_secret_key(with_currency: bool, is_mainnet: bool) -> String {
-	let suffix = if with_currency { "_with_currency" } else { "" };
-	let directory = if is_mainnet { "mainnet" } else { "testnet" };
-
-	let path = format!("./resources/secretkey/{directory}/stellar_secretkey_{directory}{suffix}");
-	std::fs::read_to_string(path).expect("should return a string")
+pub fn get_secret_key_from_env(with_currency: bool, is_mainnet: bool) -> String {
```
ebma, line 47: Probably worth adding a comment here about what the `with_currency` flag actually does.

ebma: Maybe the whole function name is a little unfortunate though and it's least confusing to split this into two functions:
- `get_secret_key_from_env_with_currency(is_mainnet: bool)` and 
- `get_secret_key_from_env(is_mainnet: bool)`

clients/vault/src/replace.rs
```diff
@@ -9,7 +9,8 @@ use runtime::{
 	VaultRegistryPallet,
 };
 use service::{spawn_cancelable, Error as ServiceError};
-use wallet::StellarWallet;
+#[allow(unused_imports)]
+use wallet::{StellarWallet, keys::get_source_secret_key_from_env};
```
ebma, line 13: Nit: If this import is only used for tests, we can also move it to the `tests` module.

clients/vault/tests/helper/helper.rs
```diff
@@ -18,12 +18,15 @@ use stellar_relay_lib::sdk::{PublicKey, SecretKey};
...
 pub fn default_destination(is_public_network: bool) -> SecretKey {
 	if is_public_network {
-		SecretKey::from_encoding(crate::helper::DEFAULT_MAINNET_DEST_SECRET_KEY)
+		SecretKey::from_encoding(get_dest_secret_key_from_env(true))
 			.expect("Should work")
 	} else {
-		SecretKey::from_encoding(crate::helper::DEFAULT_TESTNET_DEST_SECRET_KEY)
+		SecretKey::from_encoding(get_dest_secret_key_from_env(false))
 			.expect("Should work")
 	}
```
ebma, line 31: Let's simplify to
```
SecretKey::from_encoding(get_dest_secret_key_from_env(is_public_network)).expect("Should work")
```

clients/wallet/src/horizon/tests.rs
```diff
@@ -89,7 +91,7 @@ async fn build_simple_transaction(
 async fn horizon_submit_transaction_success() {
 	let horizon_client = reqwest::Client::new();
 
-	let source = secret_key_from_encoding(SECRET);
+	let source = secret_key_from_encoding(&get_source_secret_key_from_env(false));
```
ebma, line 94: While not necessary, it might be a good idea to define a new variable`let is_public_network = false` at the beginning of each test function and then replacing it also in all the function calls that currently directly use a `false` boolean. This way, we make it more explicit and make it more robust in case we decide to change the used network in the future.

clients/wallet/src/keys.rs
```diff
@@ -0,0 +1,36 @@
+#[allow(unused_imports)]
+use dotenv::dotenv;
+use std::env;
+// Gets env variables with precedence from the system environment and then from the .env file.
+// If one variable is not on the system environment, and the .env is defined, then all variables 
+// will be overridden by those on the .env file.
```
ebma, line 6:
```suggestion
use std::env;

/// Gets env variables with precedence from the system environment and then from the .env file.
/// If one variable is not on the system environment, and the .env is defined, then all variables 
/// will be overridden by those on the .env file.
```

ebma: > If one variable is not on the system environment, and the .env is defined, then all variables will be overridden by those on the .env file.

I'm not sure I understand how this works. Is it because the call to `dotenv::from_filename()` will overwrite all env variables and this is only called once the first value is not found from the system environment?

ebma (commented):
Looks good to me overall 👍  just left minor comments. 

Do you have permissions to add secrets to this repository? We still need to migrate the accounts that are removed as part of this PR. I propose we create two new accounts per Stellar network (ie. 4 in total, not re-using secret keys for both networks) and then migrate the assets and trustlines that are currently held in the accounts here to the new accounts. Do you think you'd be able to do this @gianfra-t? I can also give you some guidance.

gianfra-t:
@ebma Thanks for the comments. After I change the code for the suggestions I think I can "migrate" the accounts. Just to verify, all 4 accounts need to have the same trustliness and the only difference is that `source` accounts will have the asset in question right? Also, all 4 need some amount of lumen.

clients/vault/src/oracle/testing_utils.rs
gianfra-t: Yes. Makes more sense and is a bit more readable.

gianfra-t: In fact, what if we directly use `get_source_secret_key_from_env` and `get_dest_secret_key_from_env`? Source being = `with_currency`. Would it be too confusing?

clients/vault/tests/helper/helper.rs
gianfra-t: Good catch!

clients/wallet/src/keys.rs
gianfra-t: That was the idea. But I have modified so the .env takes precedence, and only if one variable is not declared there but has been passed to the environment it can also be used.

clients/wallet/src/keys.rs
```diff
@@ -5,16 +5,8 @@ use std::env;
 // If one variable is not on the system environment, and the .env is defined, then all variables 
 // will be overridden by those on the .env file.
```
ebma, line 6: Let's rewrite now that the logic changed. Maybe something like

> Sources all environment variables contained in a local .env file and returns the given variable for `key` if available.

clients/wallet/src/keys.rs
```diff
@@ -5,16 +5,8 @@ use std::env;
...
-        Ok(value) => Some(value),
-        Err(_) => {
-            if dotenv::from_filename("../vault/resources/secretkey/.env").is_ok() {
-                env::var(key).ok()
-            } else {
-                None
-            }
-        }
-    }
+    dotenv::from_filename("../vault/resources/secretkey/.env").ok();
```
ebma, line 8: Should we make this path configurable? I think we at least shouldn't put any nesting here because if you are a vault operator running the binary, you probably don't want to create this folder structure next to your binary. I guess it's best if we change the path to something simple like `"./.env"`

clients/vault/src/oracle/testing_utils.rs
ebma: Sure, we can also do that 👍 @b-yap do you have an opinion here?

clients/wallet/src/keys.rs
gianfra-t: The thing is that as far as I understand the keys fetched from the environment are only for tests right? Operators will supply it via flag and there is no issue them not having the keys for testing. Since we already had that folder structure for testing it won't be a problem for us, what do you think? Or is it that operators may also perform test in some scenarios?

clients/wallet/src/lib.rs
```diff
@@ -12,6 +12,7 @@ pub mod operations;
 mod stellar_wallet;
 mod task;
 pub mod types;
+pub mod keys;
```
ebma, line 15: Let's hide this behind a feature flag then. This should make it obvious that these keys are only used for testing purposes. 
```suggestion
#[cfg(feature = "testing-utils")]
pub mod keys;
```

clients/wallet/src/keys.rs
ebma: Oh you are totally right, I forgot about this. 

Let's still put the .env file in a higher folder level then. We can put it into the `/clients/.env` or `clients/wallet/.env`.

gianfra-t: Agreed, under client will be easier to handle.

clients/stellar-relay-lib/Cargo.toml
```diff
@@ -28,6 +28,8 @@ serde = { version = "1.0.152", features = ["alloc"] }
 serde_json = "1.0.93"
 serde_with = "2.2.0"
 
+wallet = { path = "../wallet" }
```
b-yap, line 31: If I were to choose between duplicate codes vs adding another lib dependency, I would prefer duplicate code.

However it looks like we're just using it in the examples and test cases. We can transfer this line under `dev-dependencies`

clients/wallet/src/keys.rs
```diff
@@ -0,0 +1,30 @@
...
+		true => get_env_variables("DEST_SECRET_MAINNET").expect("Failed to read secret key from environment"),
+		false => get_env_variables("DEST_SECRET_TESTNET").expect("Failed to read secret key from environment"),
+	};
+
+	maybe_secret
+}
+
+pub fn get_source_secret_key_from_env(is_mainnet: bool)-> String{
+	let maybe_secret = match is_mainnet {
+		true => get_env_variables("SOURCE_SECRET_MAINNET").expect("Failed to read secret key from environment"),
```
b-yap, line 25: I know it's going to be long, but is it more readable to name the env variable
`SOURCE_`**_`STELLAR`_**`_SECRET_MAINNET`? just so we know that this secret key is for stellar.

@ebma @gianfra-t ?

clients/vault/src/oracle/testing_utils.rs
b-yap: no strong opinions.  It's understandable. Let's go with 
`get_source_secret_key_from_env` `get_dest_secret_key_from_env`. 👍

clients/wallet/src/horizon/tests.rs
```diff
@@ -7,8 +7,11 @@ use crate::{
 		traits::HorizonClient,
 	},
 	mock::secret_key_from_encoding,
+	keys::get_source_secret_key_from_env,
 };
+#[allow(unused_imports)]
```
b-yap, line 12: should we just remove  `use mockall::predicate::*;`?

clients/wallet/src/horizon/tests.rs
```diff
@@ -170,7 +173,8 @@ async fn horizon_get_transaction_success() {
 #[tokio::test(flavor = "multi_thread")]
 async fn fetch_transactions_iter_success() {
 	let horizon_client = reqwest::Client::new();
-	let secret = secret_key_from_encoding(SECRET);
+	let is_public_network = false;
+	let secret = secret_key_from_encoding(&get_source_secret_key_from_env(is_public_network));
 	let fetcher = HorizonFetcher::new(horizon_client, secret.get_public().clone(), false);
```
b-yap, line 178: since we're introducing the variable `is_public_network`, we can replace `false` with that variable.

In Stellar, _public_ network is the **MAINNET**.

b-yap (changes requested):
After we identify the accounts, we will need to add a step (setting up the environment variable) in the github actions, yes?

clients/wallet/src/keys.rs
ebma: Sure, makes sense 👍 Let's call them `SOURCE_STELLAR_SECRET_MAINNET`, `SOURCE_STELLAR_SECRET_TESTNET`, `DEST_STELLAR_SECRET_MAINNET`, and `DEST_STELLAR_SECRET_TESTNET`

ebma: @gianfra-t I now added the secrets using the naming in my previous comment as repository secrets to this Github repository. 

I did some googling and it seems like to use them we need to echo them into a local `.env` file, see [here](https://stackoverflow.com/questions/61117865/how-to-set-environment-variable-in-node-js-process-when-deploying-with-github-ac). So we'd need to edit the workflow file of the Github CI action and add something like
```
- name: create env file
      run: |
        touch ./clients/.env
        echo SOURCE_STELLAR_SECRET_MAINNET=${{ secrets.SOURCE_STELLAR_SECRET_MAINNET }} >> .env
        echo ...
```

gianfra-t: That would be to create the .env, but I think they can be passed to the environment of the job as:
```
steps:
  - name: Hello world action
    env: 
      super_secret: ${{ secrets.SuperSecret }}
 ```
 At least this is what I understood from [here](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow).

clients/wallet/src/horizon/tests.rs
gianfra-t: Sure, I replaced in a few other places I see the boolean alone.

clients/stellar-relay-lib/Cargo.toml
gianfra-t: Moved!

clients/wallet/src/keys.rs
ebma: If that works, even better 👍 I was not sure if cargo would pick them up or if the scope is somehow limited if we define them like this.

gianfra-t: It seems to work! Tests are failing now because of a `SelfPaymentError` which means that probably we need to shuffle `source` <-> `destination` somewhere, but the environment variables are read.

gianfra-t: Oh okay, thanks! indeed that was hard to guess.

ebma (approved):
I replaced some other keys so that the 4 accounts supplied via env variables are really the only important ones. 

The tests, rustfmt and clippy checks passed on my machine so I think we are good to merge it right away (without waiting for the CI because it might fail due to the `sysinfo` crate issue that is only fixed in [this](https://github.com/pendulum-chain/spacewalk/pull/505) PR. )

b-yap (approved):

PR #505, opened 2024-04-02
503 make spacewalk benchmarks run successfully for pendulum runtimes

### LINKED ISSUE #503: Make Spacewalk benchmarks run successfully for Pendulum runtimes
### Context

Currently there are a few minor issues that prevent Spacewalk pallets benchmarks from being executed successfully on Pendulum, Amplitude and Foucoco runtimes:

- `WeightInfo` trait is not publicly exported for `stellar-relay` pallet.
- Existential deposit of native currency on all Pendulum runtimes is greater than the amount used for deposit in most Spacewalk pallets, therefore benchmarks are failing with `ExistentialDeposit` error [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/2dbcd1de4db5c05c8596fab447323b4b524673e2/tokens/src/lib.rs#L1107-L1110).
- `vault_registry` pallet uses [test wrapped currencies](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/currency/src/testing_constants.rs) in benchmarks and this results in failure when they are executed on any Pendulum runtime, because pairs specified in the genesis configs are [different](https://github.com/pendulum-chain/pendulum/blob/main/node/src/constants/mod.rs).

### Requirement

Provide fixes for all of the above and re-run all benchmarks on Spacewalk.

bogdanS98:
Closes #503.

Fixes included in this PR:

- `WeightInfo` trait is publicly exported for `stellar-relay` pallet so that it can be implemented in Pendulum runtimes. (see [commit](https://github.com/pendulum-chain/spacewalk/pull/505/commits/187cd28ede8d2bed6629ad58076d1dc7ff6da191))
- Created getter for wrapped currencies as a pallet config parameter instead of using [test wrapped currencies](https://github.com/pendulum-chain/spacewalk/blob/main/pallets/currency/src/testing_constants.rs). This caused benchmarks to fail when ran on any Pendulum runtime due to a mismatch between wrapped currencies definitions. (see [commit](https://github.com/pendulum-chain/spacewalk/pull/505/commits/eb445261371ef405278eb8d94898f35439d811c3))
- Amounts passed to `mint_collateral` function in `redeem` benchmarks were increased to match ED definition of Pendulum runtimes' native currencies. This issue was causing failures on deposit [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/2dbcd1de4db5c05c8596fab447323b4b524673e2/tokens/src/lib.rs#L1107-L1110) when running benchmarks. (see [commit](https://github.com/pendulum-chain/spacewalk/pull/505/commits/eb445261371ef405278eb8d94898f35439d811c3))

Also benchmarks were re-ran after adding the fixes mentioned above. (see [commit](https://github.com/pendulum-chain/spacewalk/pull/505/commits/45a67182ce83e3a9859a563695047712b29b70d3))


bogdanS98:
> Nice, great changes 🙏
> 
> I just thought that we might want to remove [this](https://github.com/pendulum-chain/spacewalk/blob/eeb5b28ebb95cd1aaf2f926ceb2006336b99b07c/pallets/currency/src/testing_constants.rs#L40-L42) function to avoid any confusion. Otherwise we have some very similar looking functions and we might forget about it. It's probably best if we use the new `get_wrapped_currency_id<T>()` function everywhere. To make it available also in the other tests and not just runtime benchmarks, we could change the macro condition to:
> 
> ```
> #[cfg(any(feature = "runtime-benchmarks", feature = "testing-constants))]
> ```
> 
> That's just one idea, maybe you have a better one @bogdanS98.

I agree that we should remove [this](https://github.com/pendulum-chain/spacewalk/blob/eeb5b28ebb95cd1aaf2f926ceb2006336b99b07c/pallets/currency/src/testing_constants.rs#L40-L42) function since it's confusing and not really necessary. I think using `get_wrapped_currency_id<T>()` in tests is a good idea, but we can also directly use the constants in `testing_constants`. WDYT @ebma?

ebma:
I think it's fine to replace it with the constant. Depends on how often we use it but should be fine.

ebma (commented):
Nice 👍 code looks good, now we just have to figure out the CI problem

ebma:
@bogdanS98 We still have this 'no space left on device' error but for some reason it only happens on this branch :( 
any ideas?

bogdanS98:
> @bogdanS98 We still have this 'no space left on device' error but for some reason it only happens on this branch :( any ideas?

Looking into this, I think I'm going to use the same storage cleanup step as in [this](https://github.com/pendulum-chain/pendulum/blob/cd2a70799f0ffd3d516f91a310e3a435081d2203/.github/workflows/test-code.yml#L16) workflow.

ebma:
Hmm yes, let's give that a shot 🤞

gianfra-t (approved):
The code looks good 👍! 

To get the new spacewalk benchmarks, in the pendulum repo it was enough to modify the `rev` to point to the latest commit of this branch, added `type GetWrappedCurrencyId = GetWrappedCurrencyId;` to the runtimes and compile ?

I was trying to do the same and get an error, but I am probably missing some spacewalk dependency to compile so I wanted to confirm if I understand the process.

bogdanS98:
> The code looks good 👍!
> 
> To get the new spacewalk benchmarks, in the pendulum repo it was enough to modify the `rev` to point to the latest commit of this branch, added `type GetWrappedCurrencyId = GetWrappedCurrencyId;` to the runtimes and compile ?
> 
> I was trying to do the same and get an error, but I am probably missing some spacewalk dependency to compile so I wanted to confirm if I understand the process.

Yes, that's all you need. What error do you get?

gianfra-t:
Particularly I get the error 
```
 get_relay_chain_currency_id as get_collateral_currency_id, get_wrapped_currency_id, *,
  |                                                                ^^^^^^^^^^^^^^^^^^^^^^^
  |                                                                |
  |                                                                no `get_wrapped_currency_id` in `getters`
  |                                                                help: a similar name exists in the module: `get_native_currency_id`
  ```
  
  Which probably points to an incorrect spacewalk dependency replacement. I just replaced `3c03e759a6f600ea72d636d4f9cc4b05d633201c` with `8699f80e50aeca63485a25119e06cb6123b05d01`, maybe I am missing a spacewalk import with a different revision?

ebma (approved):

PR #513, opened 2024-04-19
506 switch to stable rust toolchain channel

### LINKED ISSUE #506: Switch to 'stable' rust toolchain channel
At the moment, we are using the 'nightly' channel of the Rust toolchain. Going forward, we want to use the 'stable' channel because some tools like the SRTool don't work on 'nightly' anymore, at least not on the latest versions, see [this](https://github.com/paritytech/srtool/issues/62) issue. 
Switching the toolchain was not possible before as we encountered compiler errors. It might be that the recent update of the polkadot dependencies to v0.9.42 resolves those problems.

## TODO
Change the `rust-toolchain.toml` to use a 'stable' channel. Make sure that the vault, testchain and all tests still compile and work as expected.

b-yap:
closes #506 

* `mocktopus` is a lib for rust _nightly_ alone. 
Best to [constraint its import in `test` only](https://stackoverflow.com/questions/78338966/mocktopus-alternative-for-mocking-functions#comment138109993_78338966), until we plan to fully refactor the code and remove it entirely.
   * added a  script ~`build-check`~ `cmd-all` to build (with stable) the project without the `currency`'s `testing-utils`.
* ~:boom: the tests in pallet `redeem` is mocking functions from pallet `currency`. I had to put a wrapper to be able to mock it properly.~ 
   * ~:skull: This is an ugly approach, but it works well.~ 
----  
## How to begin the review:   

1. scripts/~build-check~ cmd-all:
   *  builds all features of all packages except the `testing-utils` of `currency` pallet
1. rust-toolchain.toml
   * instead of nightly, use a not-the-latest stable version: 1.77.0  
3. .github/workflows/ci-dev.yml and .github/workflows/ci-main.yml
   * need to install the nightly version, and I chose ~`2023-12-29`~ _`2024-02-09`_
        *  :exclamation: the _`2023-12-29`_ [had issues](https://github.com/pendulum-chain/spacewalk/actions/runs/8797793379/job/24143453056#step:11:1093) :exclamation: :
           * `error[E0658]: use of unstable library feature 'stdsimd'`
        * see https://github.com/tkaitchuck/aHash/issues/195 
           * suggests to use an updated version, hence _`2024-02-09`_
   * as stated above, `nightly` is required for  testing.
   * `fmt` only works for `nightly`
   *  for building on stable version, use the `cmd-all` script instead
4. Cargo.lock:
   * had to updated based on the issue stated on the previous point
5. all README.md s
   * all cargo tests need to override rust version with `+nightly` or a nightly version of your choice
6. clients/README.md
   * forgot to mention from https://github.com/pendulum-chain/spacewalk/pull/483 to setup environment variables as another option  
7. clients/runner/src/runner.rs
    * cleanup **_`not used`_** warnings. (_need your help on this @gianfra-t_)
 5. clients/stellar-relay-lib/Cargo.toml:
    * to fix sole running of tests in `stellar-relay-lib`,
    * need the _`testing-utils`_ feature of `wallet` to access the functions for retrieving the secret keys 
6.  clients/wallet/src/lib.rs:
    * need access to the keys when running tests only for `wallet`  ( _`cargo test`_)
8. clients/wallet/src/keys.rs
    * when running the example in `stellar-relay-lib`, it failed with `file not found`. 
9. ~pallets/currency/Cargo.toml:~
    * ~`mocktopus` only in `cfg test`.~
10.  ~pallets/currency/src/amount.rs:~
     * ~allow `mocktopus` only in `cfg test`~
11. ~:boom: pallets/redeem/src/ext.rs :boom::~
    * ~_Again_ this is **_quite ugly_** but _`mockable`_ feature of `currency`'s _`Amount`_ is inaccessible.~
    * ~I wrapped it in a trait :small_red_triangle_down: **`AmountExt`** :small_red_triangle_down: where we can mock the ff methods of _`Amount`_:~
      * ~`fn lock_on()`~
      * ~`fn burn_from()`~
      * ~`fn unlock_on()`~
      * ~`fn transfer()`~
     * ~implement the trait ^ by calling the actual methods of _`Amount`_~ 
12.  ~:boom: pallets/redeem/src/lib.rs :boom: :~
     * ~replace the call of `Amount` methods, with methods of   trait **`AmountExt`**:~
        * ~`fn _lock_on()`~
        * ~`fn _burn_from()`~
        * ~`fn _unlock_on()`~
       * ~`fn _transfer()`~
13. ~:boom:  pallets/redeem/src/tests.rs :boom: :~
    * ~replace the call of `Amount` methods, with methods of   trait **`AmountExt`**~
14. all pallets/*/rpc/Cargo.toml:
    * `jsonrpsee` will need the `client` feature 
14. files not specifically mentioned, are to resolve fmt and clippy issues.


.github/workflows/ci-main.yml
```diff
@@ -53,37 +76,44 @@ jobs:
...
+      - name: Clippy -- Tests and Examples
+        if: matrix.rust == 'nightly'
         uses: actions-rs/cargo@v1
         with:
+          toolchain: nightly-2024-02-09
           command: clippy
-          # We are a bit more forgiving when it comes to the code in tests and only check for correctness
-          args: --all-features --all-targets -- -A clippy::all -W clippy::correctness -A clippy::forget_copy -A clippy::forget_ref
\ No newline at end of file
+          args: --all-features --tests --benches --examples -- -A clippy::all -W clippy::correctness -A forgetting_copy_types -A forgetting_references
```
b-yap, line 119: we cannot use `--all-targets` here because the `test` must use the nightly version, while the libs and bins must use the stable version

clients/runner/src/runner.rs
```diff
@@ -515,10 +513,6 @@ impl RunnerExt for Runner {
 		&self.opts.download_path
 	}
 
-	fn parachain_url(&self) -> String {
```
b-yap, line 518: This was found to be unused. I also checked, and found it unused anywhere in the code. Please confirm @gianfra-t

clients/runner/src/runner.rs
```diff
@@ -568,10 +562,6 @@ impl RunnerExt for Runner {
 		Runner::try_load_downloaded_binary(self, release)
 	}
 
-	async fn get_request_bytes_wrapper(&self, url: String) -> Result<Bytes, Error> {
```
b-yap, line 571: same case with https://github.com/pendulum-chain/spacewalk/pull/513/files#r1577590058

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -80,7 +80,7 @@ async fn read_message_from_stellar(connector: &mut Connector) -> Result<Xdr, Err
 		// 1. the length of the next stellar message
 		// 2. the remaining bytes of the previous stellar message
 		match connector.tcp_stream.read(&mut buff_for_reading).await {
-			Ok(size) if size == 0 => continue,
+			Ok(0) => continue,
```
b-yap, line 83: suggestion from version **_`nightly-2024-02-09`_**

clients/wallet/src/horizon/horizon.rs
```diff
@@ -35,8 +35,7 @@ const BASE_BACKOFF_DELAY_IN_SECS: u64 = 10;
 pub fn horizon_url(is_public_network: bool, is_need_fallback: bool) -> &'static str {
 	if is_public_network {
 		if is_need_fallback {
-			let other_urls =
-				vec!["https://horizon.stellarx.com", "https://horizon.stellar.lobstr.co"];
+			let other_urls = ["https://horizon.stellarx.com", "https://horizon.stellar.lobstr.co"];
```
b-yap, line 38: suggestion from version `nightly-2024-02-09` to use the array directly

clients/wallet/src/keys.rs
```diff
@@ -7,7 +7,11 @@ use std::env;
 // previously passed to the environment.
 // A variable not .env but passed to the environment will not be overridden.
 fn get_env_variables(key: &str) -> Option<String> {
-	dotenv::from_filename("../.env").ok();
+	if let None = dotenv::from_filename("../.env").ok() {
```
b-yap, line 10: had a problem with running the example in `stellar-relay-lib`.
I moved the env file to the root directory; that's why I added a condition to try looking at the current directory.

clients/wallet/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub use task::*;
 mod cache;
 pub mod error;
 mod horizon;
-#[cfg(feature = "testing-utils")]
+#[cfg(any(test, feature = "testing-utils"))]
```
b-yap, line 11: keys are also used in the unit test of this lib, so I added `test`.

pallets/currency/src/amount.rs
```diff
@@ -1,5 +1,10 @@
 // We allow these lints because they are only caused by the 'mockable' attribute.
-#![allow(clippy::forget_non_drop, clippy::swap_ptr_to_ref, clippy::forget_ref, clippy::forget_copy)]
+#![allow(
+	clippy::forget_non_drop,
+	clippy::swap_ptr_to_ref,
+	forgetting_references,
```
b-yap, line 5: * `clippy::forget_ref`is updated to `forgetting_references`
* `clippy::forget_copy` is updated to `forgetting_copy_types` 
in the version `nightly-2024-02-09`

pallets/currency/src/mock.rs
```diff
@@ -104,11 +102,14 @@ impl pallet_balances::Config for Test {
 parameter_types! {
 	pub const GetNativeCurrencyId: CurrencyId = DEFAULT_NATIVE_CURRENCY;
 	pub const GetRelayChainCurrencyId: CurrencyId = DEFAULT_COLLATERAL_CURRENCY;
-	#[cfg(feature = "runtime-benchmarks")]
```
b-yap, line 107: this causes error to building/testing other features. 
for example: 
```
cargo +nightly test -p vault --features integration-test`
``` 
actually fails with a duplicate  lang item issue.

`cfg` macros should be done _outside_ of `parameter_types!` macros.

pallets/oracle/src/oracle_mock.rs
```diff
@@ -15,7 +15,7 @@ pub struct DataKey {
 
 impl PartialOrd<Self> for DataKey {
 	fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-		self.blockchain.partial_cmp(&other.blockchain)
+		Some(self.cmp(other))
```
b-yap, line 18: this is a suggestion in version `nightly-2024-02-09`

TorstenStueber:
Hey @b-yap, looks like a lot of work went into this. Thanks!

I see that a couple of changes are also pure formatting changes. How is that? Did the formatter settings change or did we not properly use a formatter before?

b-yap:
@TorstenStueber I did not touch the fmt toml file. I think these are all new preferences from the new version? 
Like adding `;` of a return statement, or removing `{` `}` for single statements.

Because I ran `cargo +nightly-2024-02-09 fmt --all` [after the CI failed](https://github.com/pendulum-chain/spacewalk/actions/runs/8803926726/job/24163183669#step:13:1).

TorstenStueber:
Hmm, is this expected that the default formatting changes between different versions of rust (and therefore rust fmt)? Or did we not properly format before the change?

b-yap:
@TorstenStueber I confirmed it by running 
`cargo +nightly-2023-04-15 fmt --all` on a clean `main` branch, and there were no changes. We did it properly.

Screenshot 2024-04-24 at 11 25 08 PM

^ In this screenshot, the differences occurred after `cargo +nightly-2024-02-09 fmt --all` ; the formatting changes are coming from the new version.

TorstenStueber:
Interesting and unexpected. Looking again, there seem to be only a few changes that are only due to formatting. One of them is [this](https://github.com/pendulum-chain/spacewalk/pull/513/files#diff-945cf6fcb57d25950f6c53bcd090426caa4c1b08b3a32d1da6670febd283e5abR297) and it is clear that this it enacts the [`trailing_semicolon = false`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#trailing_semicolon) rule. However, `trailing_semicolon` was already defined to be `false` before your change ([see on `main`](https://github.com/pendulum-chain/spacewalk/blob/b1dc81f9de7bca3408b490587280a10732f7b9dd/rustfmt.toml#L22)). So it looks like the formatter did not properly format all files before your change. Maybe a bug in that version of rust. But all good now!

clients/runner/src/runner.rs
gianfra-t: We remove some logic from the original runner code, so it makes sense that some methods are not used. I think it is fine!

clients/runner/src/runner.rs
gianfra-t: Also I think it is fine to remove.

clients/runner/src/runner.rs
```diff
@@ -183,6 +183,7 @@ impl Runner {
 			// The set permissions are: -rwx------
 			.mode(0o700)
 			.create(true)
+			.truncate(true)
```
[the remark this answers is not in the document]
gianfra-t: I think this is a good suggestion. I just checked at the docs for this option and it will essentially ensure that the opened file is clean, which is what we want since we will save the new binary in there.

In any case at the end of the operation the checksums will be compared, just in case.

README.md
```diff
@@ -78,7 +78,7 @@ The handling of redeem events is implemented in `clients/vault/src/redeem.rs`.
 ## Run all tests
 
 ```
-cargo test --lib --features standalone-metadata -- --nocapture
+cargo +nightly --lib --features standalone-metadata -- --nocapture
```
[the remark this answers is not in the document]
b-yap: Yes you are right! 🤦🏽‍♀️ My bad.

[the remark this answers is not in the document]
b-yap: https://github.com/pendulum-chain/spacewalk/pull/513/commits/536798f7df58ba5868d27f47abd1ead2bee8a0f8

pallets/redeem/src/lib.rs
```diff
@@ -526,12 +528,12 @@ mod self_redeem {
 	) -> DispatchResult {
 		// burn the tokens that the vault no longer is backing
 		consumed_issued_tokens
-			.lock_on(account_id)
+			._lock_on(account_id)
```
ebma, line 531: Can we find a way not to have to call functions with this underscore? I understand you created an extra trait to make this mockable. Can we move that trait next to the actual definition and unify It so that we don't need the underscore? 
Otherwise, this creates some confusion and it's not easy to see without context why this is necessary.

pallets/vault-registry/README.md
```diff
@@ -8,7 +8,7 @@ Run `cargo build` from the root folder of this directory.
 
 ## Testing
 
-Run `cargo test` from the root folder of this directory.
+Run `cargo +nightly` from the root folder of this directory.
```
ebma, line 11:
```suggestion
Run `cargo +nightly-2024-02-09 test` from the root folder of this directory.
```

ebma (commented):
Nice changes 👍 

- In all the READMEs we now mention `+nightly-<some_version>`. This is okay and if we want to change it in the future, we can just do a find-and-replace search to change it without much effort. However, I'm still curious if there is a way for us to change it to `+nightly` but have cargo run it with a specific version. Can we define both stable and nightly in the rust-toolchain.toml or similar so that both versions are fixed?
- We also need to make sure that the changes are applied to our other compilation pipeline in the other infrastructure.

gianfra-t:
It looks very good 👍! 
I am surprised why the `mockable` macro works different between versions and now it doesn't allow access to the trait anymore. In any case I think the wrapper trait workaround is not that "ugly" and solves the problem.

The only thing remaining to me is to make it work with the `--features runtime-benchmarks` flag.

b-yap:
@ebma 

> In all the READMEs we now mention +nightly-<some_version>. 

 - Testing using the latest version of nightly is fine; as long as it's installed. The danger was testing using an older version. I changed to `+nightly` in the READMEs.
 
 >  Can we define both stable and nightly in the rust-toolchain.toml or similar so that both versions are fixed?

 - I checked the [profile](https://rust-lang.github.io/rustup/concepts/profiles.html) and there's nothing there.
    It's [not even possible per package](https://users.rust-lang.org/t/configuring-the-toolchain-for-a-specific-package-in-a-workspace/93208); I doubt it can differentiate between `build` and `test`.
 - I tried [cargo aliases](https://doc.rust-lang.org/nightly/cargo/reference/config.html#alias) but the override is not a "subcommand", so:
    ```
         error: no such command: `+nightly`

	   Cargo does not handle `+toolchain` directives.
	   Did you mean to invoke `cargo` through `rustup` instead?
    ```

clients/runner/Cargo.toml
```diff
@@ -23,7 +23,7 @@ signal-hook = "0.3.14"
 signal-hook-tokio = { version = "0.3.1", features = ["futures-v0_3"] }
 futures = "0.3.21"
 backoff = { version = "0.3.0", features = ["tokio"] }
-subxt = { version = "0.29.0", default_features = false,  features = ["jsonrpsee-ws"] }
```
b-yap, line 26: `default_features` is deprecated.

pallets/redeem/src/lib.rs
b-yap: I agree; in a few weeks I will have forgotten this.

I have reverted the code to allow `mocktopus` in pallet _`currency`_'s feature: _`testing-utils`_. 
That means **`cargo b --all-features`** will not work.

I created a script `build-check` to build all known packages for stable rust version. 
`testing-utils` will only compile on nightly version; or when running `cargo test`.

.github/workflows/ci-main.yml
```diff
@@ -36,10 +40,29 @@ jobs:
...
         # Call `rustup show` as a hack so that the toolchain defined in rust-toolchain.toml is installed
         run: rustup show
 
+      - name: Remove rust-toolchain.toml for nightly
+        if: matrix.rust == 'nightly'
+        # To make sure that the nightly version will be used all throughout
+        run: |
+          rm /home/runner/work/spacewalk/spacewalk/rust-toolchain.toml
+
+      - name: Setup nightly Rust toolchain
```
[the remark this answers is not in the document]
b-yap: dev.yml and main.yml should look the same, now.

clients/README.md
```diff
@@ -77,22 +95,24 @@ You can also find an example for setting these variables in the `.env.example` f
 
 ### Running the tests
 
+**Note** Tests should run using Rust **_`nightly-2023-12-29`_** version. Make sure to install and add the target `wasm32-unknown-unknown`.
```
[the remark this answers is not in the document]
b-yap: changed to "nightly";

.github/workflows/ci-dev.yml
```diff
@@ -75,38 +95,39 @@ jobs:
...
           args: --all --all-features
 
       - name: Rustfmt
+        if: matrix.rust == 'nightly'
         uses: actions-rs/cargo@v1
         with:
+          toolchain: nightly-2024-02-09
           command: fmt
-          args: --all -- --check
+          args: --all -- -check
```
ebma, line 127:
```suggestion
          args: --all -- --check
```

pallets/redeem/src/tests.rs
```diff
@@ -358,6 +358,7 @@ fn test_request_redeem_succeeds_with_self_redeem() {
 fn test_liquidation_redeem_succeeds() {
 	run_test(|| {
 		let total_amount = 10 * 100_000_000;
+		println!("the expected redeemer balance: {total_amount}");
```
ebma, line 361: Nit: remove

scripts/build-check
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+echo "Checks the build of all packages"
+
+cargo b -p spacewalk-primitives --all-features
+
+echo "Check all pallets"
+cargo b -p clients-info --all-features
+echo "excluding the feature testing-utils of pallet currency; that is for nightly build."
+cargo b -p currency --features std,testing-constants,try-runtime,runtime-benchmarks
```
ebma, line 10: So the `currency` crate is the only reason we need to do this, correct?

b-yap: yes

pallets/issue/rpc/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 
 [dependencies]
 codec = { package = "parity-scale-codec", version = "3.0.0" }
-jsonrpsee = { version = "0.16.0", features = ["server", "macros"] }
+jsonrpsee = { version = "0.16.0", features = ["client", "server", "macros"] }
```
b-yap, line 9: need the feature `client`, else:
```
error[E0433]: failed to resolve: could not find `client` in `core`
  --> pallets/.../rpc/src/lib.rs:16:1
   |
16 | #[rpc(client, server)]
   | ^^^^^^^^^^^^^^^^^^^^^^ could not find `client` in `core`
   |
note: found an item that was configured out
  --> /.../.cargo/registry/src/index.crates.io-6f17d22bba15001f/jsonrpsee-core-0.16.3/src/lib.rs:55:10
   |
55 |     pub mod client;
   |             ^^^^^^
   = note: the item is gated behind the `client` feature
   = note: this error originates in the attribute macro `rpc` (in Nightly builds, run with -Z macro-backtrace for more info)
```

pallets/fee/Cargo.toml
```diff
@@ -75,6 +75,7 @@ runtime-benchmarks = [
     "frame-benchmarking",
     "frame-support/runtime-benchmarks",
     "frame-system/runtime-benchmarks",
+    "currency/runtime-benchmarks",
```
b-yap, line 78: need the `"currency/runtime-benchmarks"` to be able to run tests with the feature `runtime-benchmarks`.

ebma (commented):
Looks good to me overall 👍 

There is a CI warning raised by clippy that we need to fix. Besides that, I would very much appreciate it if we could add a new section to the README that explains how/why we do this distinction between Rust stable vs nightly, how and when the `cmd-all` script needs to be used (ie. which cargo commands do still work and which don't and need to be called from that instead), also mentioning that for some IDEs built-in sync feature may break because the `rust-toolchain.toml` defines a stable toolchain while some features require nightly. 

I tested it and I can still run my custom commands from within RustRover by manually choosing the 'nightly' channel in the 'Run Configurations' dialog, so this is not a dealbreaker.

scripts/cmd-all
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+set -e
+
+##### This script executes the command to ALL packages of this project.
```
b-yap, line 4: @ebma I added a comment here, inside the script.

README.md
```diff
@@ -74,11 +74,24 @@ specified in the redeem event.
 The handling of redeem events is implemented in `clients/vault/src/redeem.rs`.
 
 # Build and Run
+There was a [plan to move out from Rust `nightly` version and move to stable](https://github.com/pendulum-chain/spacewalk/issues/506). The [default toolchain](./rust-toolchain.toml) is set to stable.  
+However the [pallet `currency`](./pallets/currency)'s
+feature _`testing-utils`_ is using [`mocktopus`](https://docs.rs/mocktopus/latest/mocktopus/#) — a nightly only lib — and will potentially break your IDEs:
```
b-yap, line 79: @ebma explained the error that people will see, when relying on IDEs.

README.md
```diff
@@ -93,6 +106,12 @@ cargo run --bin spacewalk-standalone --release -- --dev
 ./target/release/node-template --dev
 ```
 
+## [`cmd-all` script](./scripts/cmd-all)
```
b-yap, line 109: @ebma  added description of the script `cmd-all`.

ebma: Maybe we should also mention our specific use-case for this, ie. that it's needed to circumvent the issue with the currency crate, etc. Otherwise it's not really clear why we have the need for such a script at all.

b-yap: I updated the [description](https://github.com/pendulum-chain/spacewalk/pull/513/commits/38547a661cc1f3f5a9157dc5a85848f53ff31b37)

b-yap:
@ebma @gianfra-t  had to rebase.

README.md
ebma: Nice, I'm happy now 👍

gianfra-t (approved):
Thanks @b-yap looks really good!

ebma:
@zoveress as part of this PR, the CI workflows we do for Spacewalk changed. Do you think you are able to adjust the Gitlab testing pipeline based on the contents of [this](https://github.com/pendulum-chain/spacewalk/blob/1c3ea1c163e721267e45a8f8baae3b4964df6ff3/.github/workflows/ci-main.yml) file? Otherwise please reach out and we are happy to help. The current testing pipeline will not work anymore on Github because we switched to a different Rust toolchain and also had to run the CI a bit differently.

PR #517, opened 2024-04-29
Allow Tokio Console support

### LINKED ISSUE #512: Investigate occasional high CPU usage of vault clients
It seems like sometimes, the vault clients' CPU usage increases up to 100% without any apparent reason. This leads to the vault client becoming unresponsive and not handling bridge requests anymore.

What we know:
- This issue happens across **all** Amplitude _and_ Pendulum vaults at around the same time.
- In the monitoring, we can see that the CPU usage increases from 50% to 100% within 45 minutes.
- Memory and disk usage are not increasing during this time window.
- After the periodic restart, the CPU usage goes back to ~50%
- There are no messages appearing in the logs (at least on 'INFO' log level).
- The Spacewalk-related metrics are unavailable for this timeframe which indicates that the whole client becomes unresponsive

image

image

b-yap:
This does not completely close #512, but it helps with tracking.

But if we can see which threads have been running the longest; which ones are idle for a long time, it could help determine where/when a vault gets stuck.

The [`tokio-console`](https://docs.rs/tokio-console/latest/tokio_console/) is described as:
> a debugging and profiling tool for asynchronous Rust applications, which collects and displays in-depth diagnostic data on the asynchronous tasks, resources, and operations in an application. 

^ From using tokio-console,  a zombie task was found.
When the vault restarts, it shuts down all running tasks except for 1:
https://github.com/pendulum-chain/spacewalk/blob/e7a672ba1e21c98a70df30a6ee458317951dd597/clients/wallet/src/resubmissions.rs#L40-L47
<br>
Aside from this, I found out that the stream also gets stuck:
https://github.com/pendulum-chain/spacewalk/blob/e7a672ba1e21c98a70df30a6ee458317951dd597/clients/stellar-relay-lib/src/connection/connector/message_reader.rs#L82
Requires a timeout to trigger reconnection. 

----  
## How to begin the review:   
I have added comments to help with the review.
1. all affected toml files:
    * _`tracing`_ feature of `tokio` is necessary for `tokio-console`
    *  `console-subscriber` dependency is necessary for `tokio-console`
2. clients/wallet/src/stellar_wallet.rs:
   * I needed a channel to stop the task, the reason for a new field of `StellarWallet`:
      ```rust
            /// a sender to 'stop' a scheduled resubmission task
	    pub(crate) resubmission_end_signal: Option<mpsc::Sender<()>>,
      ```
3. clients/wallet/src/resubmissions.rs:
    * Added a `fn stop_periodic_resubmission_of_transactions(&mut self)`
    * Update inside `fn start_periodic_resubmission_of_transactions_from_cache(...)`:
       * create a new channel and store the sender in `StellarWallet`'s _`resubmission_end_signal`_ field.
       * inside the spawned task, break the loop once the receiver receives a shutdown signal
         ```rust
              loop {
			// a shutdown message was sent. Stop the loop.
			if let Some(_) = receiver.recv().await {
				break;
			}
	     ...		
         ```
4. clients/vault/src/system.rs:
    * calls the `fn stop_periodic_resubmission_of_transactions(...)` from the `wallet` field of `VaultService`:
       ```rust
            /// shuts down the resubmission task running in the background
	    async fn shutdown_wallet(&self) {
		 ...
		 let mut wallet = self.stellar_wallet.write().await;
		 wallet.stop_periodic_resubmission_of_transactions().await;
		...
	   }
       ```
   * calls the `fn shutdown_wallet(...)` when the service stops.
      ```rust
          	async fn start(&mut self) -> Result<(), ServiceError<Error>> {
		let result = self.run_service().await;

		self.shutdown_wallet().await;
		...
       ```
5. clients/stellar-relay-lib/src/connection/connector/message_reader.rs:
    * introduce `READ_TIMEOUT_IN_SECS`; how long to wait on reading from the stream.
    * add a timeout to reading the stream
      ```rust 
           timeout(
			 Duration::from_secs(READ_TIMEOUT_IN_SECS),
			 connector.tcp_stream.read(&mut buff_for_reading),
            ).await
      ```
6. clients/README.md:
    * documentation on how to use `tokio-console`

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,0 @@
-[build]
-rustflags = ["--cfg", "tokio_unstable"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
--- a/clients/runner/Cargo.toml
+++ b/clients/runner/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 [dependencies]
 clap = { version = "4.0.17", features = ["derive"]}
 hex = "0.4.3"
-tokio = { version = "1.8", features = ["rt-multi-thread", "macros", "time"] }
+tokio = { version = "1.37", features = ["rt-multi-thread", "macros", "time"] }
 codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "full", "bit-vec"] }
 sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
 thiserror = "1.0.0"
--- a/clients/runtime/Cargo.toml
+++ b/clients/runtime/Cargo.toml
@@ -31,7 +31,7 @@ log = "0.4.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.71"
 thiserror = "1.0"
-tokio = { version = "1.0", features = ["full"] }
+tokio = { version = "1.37", features = ["full"] }
 prometheus = { version = "0.12.0", features = ["process"] }
 url = "2"
 
--- a/clients/stellar-relay-lib/Cargo.toml
+++ b/clients/stellar-relay-lib/Cargo.toml
@@ -34,7 +34,7 @@ substrate-stellar-sdk = {git = "https://github.com/pendulum-chain/substrate-stel
 
 err-derive = "0.3.1"
 
-tokio = { version = "1.0", features = [
+tokio = { version = "1.37", features = [
     "macros",           # allows main function to be async
     "rt-multi-thread",  # for multi-thread runtime
     "sync",             # to make channels available
--- a/clients/stellar-relay-lib/Cargo.toml
+++ b/clients/stellar-relay-lib/Cargo.toml
@@ -16,7 +16,7 @@ wallet = { path = "../wallet", features = ["testing-utils"] }
 
 [dependencies]
 hex = "0.4.3"
-log = {version = "0.4.14"}
+tracing = { version = "0.1", features = ["log"] }
 
 base64 = "0.13.0"
 rand = "0.8.5"
--- a/clients/stellar-relay-lib/Cargo.toml
+++ b/clients/stellar-relay-lib/Cargo.toml
@@ -38,7 +38,8 @@ tokio = { version = "1.0", features = [
     "macros",           # allows main function to be async
     "rt-multi-thread",  # for multi-thread runtime
     "sync",             # to make channels available
-    "time"              # for timeouts and sleep, when reconnecting
+    "time",             # for timeouts and sleep, when reconnecting
+    "tracing"           # for tokio console
 ] }
 async-std = { version = "1.12.0", features = ["attributes"] }
 
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -7,6 +7,7 @@ use substrate_stellar_sdk::{
 	types::{AuthenticatedMessageV0, Curve25519Public, HmacSha256Mac, MessageType},
 	XdrCodec,
 };
+use tracing::{debug, error, trace};
 
 use crate::{
 	connection::{
@@ -86,7 +87,7 @@ impl Connector {
 		body: &[u8],
 	) -> Result<(), Error> {
 		let remote_info = self.remote_info.as_ref().ok_or(Error::NoRemoteInfo)?;
-		log::trace!(
+		trace!(
 			"verify_auth(): remote sequence: {}, auth message sequence: {}",
 			remote_info.sequence(),
 			auth_msg.sequence
@@ -96,7 +97,7 @@ impl Connector {
 		let auth_msg_xdr =
 			String::from_utf8(auth_msg_xdr.clone()).unwrap_or(format!("{:?}", auth_msg_xdr));
 
-		log::debug!("verify_auth(): received auth message from Stellar Node: {auth_msg_xdr}");
+		debug!("verify_auth(): received auth message from Stellar Node: {auth_msg_xdr}");
 
 		if remote_info.sequence() != auth_msg.sequence {
 			// must be handled on main thread because workers could mix up order of messages.
@@ -169,7 +170,7 @@ impl Connector {
 
 	pub fn stop(&mut self) {
 		if let Err(e) = self.tcp_stream.shutdown(Shutdown::Both) {
-			log::error!("stop(): failed to shutdown tcp stream: {}", e);
+			error!("stop(): failed to shutdown tcp stream: {}", e);
 		}
 	}
 }
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -7,7 +7,7 @@ use substrate_stellar_sdk::{
 	types::{AuthenticatedMessageV0, Curve25519Public, HmacSha256Mac, MessageType},
 	XdrCodec,
 };
-use tracing::{debug, error, trace};
+use tracing::{error, trace};
 
 use crate::{
 	connection::{
@@ -92,12 +92,6 @@ impl Connector {
 			remote_info.sequence(),
 			auth_msg.sequence
 		);
-
-		let auth_msg_xdr = auth_msg.to_base64_xdr();
-		let auth_msg_xdr =
-			String::from_utf8(auth_msg_xdr.clone()).unwrap_or(format!("{:?}", auth_msg_xdr));
-
-		debug!("verify_auth(): received auth message from Stellar Node: {auth_msg_xdr}");
 
 		if remote_info.sequence() != auth_msg.sequence {
 			// must be handled on main thread because workers could mix up order of messages.
--- a/clients/stellar-relay-lib/src/connection/connector/message_handler.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/message_handler.rs
@@ -2,7 +2,7 @@ use substrate_stellar_sdk::{
 	types::{ErrorCode, Hello, MessageType, StellarMessage},
 	XdrCodec,
 };
-
+use tracing::{error, warn, trace};
 use crate::connection::{
 	authentication::verify_remote_auth_cert,
 	helper::{error_to_string, time_now},
@@ -33,13 +33,13 @@ impl Connector {
 
 			MessageType::ErrorMsg => match auth_msg.message {
 				StellarMessage::ErrorMsg(e) => {
-					log::error!(
+					error!(
 						"process_raw_message(): Received ErrorMsg during authentication: {}",
 						error_to_string(e.clone())
 					);
 					return Err(Error::from(e))
 				},
-				other => log::error!(
+				other => error!(
 					"process_raw_message(): Received ErrorMsg during authentication: {:?}",
 					other
 				),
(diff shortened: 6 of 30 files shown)


clients/runner/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 [dependencies]
 clap = { version = "4.0.17", features = ["derive"]}
 hex = "0.4.3"
-tokio = { version = "1.8", features = ["rt-multi-thread", "macros", "time"] }
```
b-yap, line 10: latest tokio vresion is [1.37.](https://docs.rs/tokio/1.37.0/tokio/index.html)

clients/runtime/Cargo.toml
```diff
@@ -31,7 +31,7 @@ log = "0.4.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.71"
 thiserror = "1.0"
-tokio = { version = "1.0", features = ["full"] }
+tokio = { version = "1.37", features = ["full"] }
```
b-yap, line 34: The reason for this is the "lost its waker" issue in tokio, [fixed in the newer versions](https://github.com/tokio-rs/console/issues/345#issuecomment-1817939351).

clients/stellar-relay-lib/Cargo.toml
```diff
@@ -16,7 +16,7 @@ wallet = { path = "../wallet", features = ["testing-utils"] }
 
 [dependencies]
 hex = "0.4.3"
-log = {version = "0.4.14"}
+tracing = { version = "0.1", features = ["log"] }
```
b-yap, line 19: helps with [tokio-console and instrumentation](https://github.com/tokio-rs/console?tab=readme-ov-file#whats-all-this-then): 
> the [console-subscriber](https://github.com/tokio-rs/console/blob/main/console-subscriber) crate in this repository contains an implementation of the instrumentation-side API as a [tracing-subscriber](https://lib.rs/crates/tracing-subscriber) [Layer](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html), for projects using [Tokio](https://tokio.rs/) and [tracing](https://lib.rs/crates/tracing).

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
```
b-yap, line 2: to make [`tokio-console` work:](https://github.com/tokio-rs/console?tab=readme-ov-file#using-it):
> in order to collect task data from Tokio, the tokio_unstable cfg must be enabled.

b-yap: With the Rust stable version, there hasn't been any problems. The `tokio-console` command displays just fine, although I was not able to see the [histogram](https://github.com/tokio-rs/console/blob/main/assets/readme/task-details.png).

The worst it could do would be returning a blank view.

clients/stellar-relay-lib/src/connection/connector/connector.rs
```diff
@@ -86,18 +87,12 @@ impl Connector {
...
 	) -> Result<(), Error> {
 		let remote_info = self.remote_info.as_ref().ok_or(Error::NoRemoteInfo)?;
-		log::trace!(
+		trace!(
 			"verify_auth(): remote sequence: {}, auth message sequence: {}",
 			remote_info.sequence(),
 			auth_msg.sequence
 		);
 
-		let auth_msg_xdr = auth_msg.to_base64_xdr();
```
b-yap, line 95: I decided to remove this because the xdr was huge and may not be helpful at all.

clients/stellar-relay-lib/src/connection/connector/message_handler.rs
```diff
@@ -80,29 +78,23 @@ impl Connector {
...
+				error!("process_stellar_message(): Received ErrorMsg during authentication: {e:?}");
 				if e.code == ErrorCode::ErrConf || e.code == ErrorCode::ErrAuth {
 					return Err(Error::from(e))
 				}
 				return Ok(Some(StellarMessage::ErrorMsg(e)))
 			},
 
+			// we do not handle other messages. Return to caller
 			other => {
-				log::trace!(
```
b-yap, line 101: we are only interested on messages that will be processed by the vault.
Logging should be done over there than here.

clients/vault/src/main.rs
```diff
@@ -147,6 +147,7 @@ async fn start() -> Result<(), ServiceError<Error>> {
 
 #[tokio::main]
 async fn main() {
+	console_subscriber::init();
```
b-yap, line 150: To [use the `tokio-console`](https://github.com/tokio-rs/console?tab=readme-ov-file#instrumenting-your-program).

clients/vault/src/system.rs
```diff
@@ -679,15 +682,6 @@ impl VaultService {
 
 		tasks.append(&mut replace_tasks);
 
-		tasks.push((
```
b-yap, line 682: :scream: removing this duplicate, as it was original found in https://github.com/pendulum-chain/spacewalk/blob/e7a672ba1e21c98a70df30a6ee458317951dd597/clients/vault/src/system.rs#L623-L630
inside `fn create_initial_tasks(...`

----

Originally it runs after `Replace Cancellation Scheduler` and  before `Redeem Request Listener`:
https://github.com/pendulum-chain/spacewalk/blob/31df1832483199b978ae07f6ed4e018a11ff4dba/clients/vault/src/system.rs#L619-L639

clients/stellar-relay-lib/src/connection/connector/message_sender.rs
```diff
@@ -28,14 +29,14 @@ impl Connector {
 
 	pub async fn send_hello_message(&mut self) -> Result<(), Error> {
 		let msg = self.create_hello_message(time_now())?;
-		log::info!("send_hello_message(): Sending Hello Message: {}", to_base64_xdr_string(&msg));
+		info!("send_hello_message(): Sending Hello Message: {}", to_base64_xdr_string(&msg));
```
ebma, line 32: Let's change the log level to `debug`.

clients/stellar-relay-lib/src/connection/connector/message_sender.rs
```diff
@@ -28,14 +29,14 @@ impl Connector {
...
 
 		self.send_to_node(msg).await
 	}
 
 	pub(super) async fn send_auth_message(&mut self) -> Result<(), Error> {
 		let msg = create_auth_message();
-		log::info!("send_auth_message(): Sending Auth Message: {}", to_base64_xdr_string(&msg));
+		info!("send_auth_message(): Sending Auth Message: {}", to_base64_xdr_string(&msg));
```
ebma, line 39: `debug` here as well

clients/vault/src/system.rs
```diff
@@ -940,4 +934,12 @@ impl VaultService {
 		tracing::info!("Got new block at height {startup_height}");
 		Ok(startup_height)
 	}
+
+	/// shuts down the resubmission task running in the background
+	async fn shutdown_wallet(&self) {
```
ebma, line 939: Let's rename to `try_shutdown_wallet()` because I suppose this will only try to shut it down but not error if it's not running (yet) (?)

clients/vault/src/system.rs
```diff
@@ -940,4 +934,12 @@ impl VaultService {
...
+		tracing::info!("shutdown_wallet(): stop the resubmission scheduler");
+		let mut wallet = self.stellar_wallet.write().await;
+		wallet.stop_periodic_resubmission_of_transactions().await;
```
ebma, line 942: I'd also rename the function on the wallet to `try_stop_periodic_resubmission_of_transactions()`

clients/wallet/src/resubmissions.rs
```diff
@@ -26,10 +26,26 @@ const MAXIMUM_TX_FEE: u32 = 10_000_000; // 1 XLM
 
 #[cfg_attr(test, mockable)]
 impl StellarWallet {
+	/// sends a signal to stop the resubmission task
+	pub async fn stop_periodic_resubmission_of_transactions(&mut self) {
+		match &self.resubmission_end_signal {
+			None => {
+				tracing::warn!("stop_periodic_resubmission_of_transactions(): no schedule to stop");
```
ebma, line 33: Let's maybe change the level to `debug!` as I assume this message will always be shown on the initial startup of the vault client (ie. when not triggered by a periodic restart) and it might be confusing to some.

clients/wallet/src/resubmissions.rs
```diff
@@ -40,11 +56,18 @@ impl StellarWallet {
 		tokio::spawn(async move {
 			let me_clone = Arc::clone(&me);
 			loop {
-				pause_process_in_secs(interval_in_seconds).await;
+				// a shutdown message was sent. Stop the loop.
+				if let Some(_) = receiver.recv().await {
+					tracing::info!("start_periodic_resubmission_of_transactions_from_cache(): scheduler stopped.");
```
ebma, line 61: `debug`

.cargo/config.toml
ebma: This doesn't work for me, unfortunately I have to explicitly pass the `RUSTFLAGS="--cfg tokio_unstable"` when trying to run or build the client. 

I assume it's because I do have a `~/.cargo/config.toml` file where I already specify some rust flags
```
[target.aarch64-apple-darwin]
rustflags = ["-C", "link-arg=-fuse-ld=/opt/homebrew/opt/llvm/bin/ld64.lld"]
```
and this might override the workspace configuration. 

I confirmed this by removing my 'global' config.toml and then recompiling. This worked.

Generally I would prefer if we could enable this `tokio-console` thing with an extra feature flag and have it disabled by default. Do you think this would be possible @b-yap?

ebma (commented):
The code looks good to me overall, nice job and great feature @b-yap 🚀 
I left some comments.

.cargo/config.toml
b-yap: See https://github.com/pendulum-chain/spacewalk/pull/517/commits/e77d7acb7598b6beb300dffcae085c9667e53777

I tried to set it through `[env]` .cargo/config.toml, and it didn't work. All `RUSTFLAGS`, `CARGO_ENCODED_RUSTFLAGS` and `CARGO_BUILD_FLAGS` won't work.


I also tried to do it through a build script and didn't work:
```rust
fn main(){
    cfg_if::cfg_if! {
        if #[cfg(feature = "allow-debugger")] {
            println!("cargo:rustc-env=RUSTFLAGS=\"--cfg tokio_unstable\"");
        }
    }
}
```
and
`println!("cargo:rustc-flags=\"--cfg tokio_unstable\"");`
as explained in https://github.com/rust-lang/cargo/pull/10141.


In the end I updated the doc to tell the user to set the rustflags themselves.

ebma (approved):
Tested it locally again, can confirm it works with the new feature flag.

Great changes, thanks a lot @b-yap 👌

b-yap:
@ebma Would be great if we have the vaults running with this feature. I bet it'll look very different (and interesting).

ebma:
True, but I'm not sure if we can easily connect to the clients with `tokio-console` if they are running in a docker container. 

@zoveress in this PR we added support for a profiling tool that lets us investigate resource consumption of tasks within our process. As you can see [here](https://github.com/tokio-rs/console?tab=readme-ov-file#running-the-console), it's fairly simple to run. As far as I understand, the changes we made to the client will make it expose a new endpoint (by default) on port 6669 that is used by the `tokio-console` tool for accessing the debug information. Using this setup would be very easy to do in the EC2 instances but maybe not in our Kubernetes cluster. @zoveress do you have an idea how we could be able to access it also on Kubernetes?

This is not super important but would be nice to have.

zoveress:
Is the endpoint password secured?

ebma:
No it's not and I think it's only supposed to be used locally. Can we expose it to the GitLab runner only and access it from there somehow?

PR #523, opened 2024-05-08
Modify DecimalsLookup trait, add PINK, HDX, vDOT to decimals list

### LINKED ISSUE #522: Modify DecimalsLookup trait to use it's own CurrencyId type
### Context

`DecimalsLookup` [trait](https://github.com/pendulum-chain/spacewalk/blob/ea66a3062e6edf05a4ed7b377f107e1f95b688f1/primitives/src/lib.rs#L498) needs to reference it's own `CurrencyId` type instead of `spacewalk_primitives::CurrencyId` in the `decimals` function, so that it can be implemented by mocks which usually have the `CurrencyId` as a `u64`.

### Requirement

Modify `DecimalsLookup` trait so that it uses it's own `CurrencyId` type.

### Extra

Add PINK, HDX and vDOT tokens [here](https://github.com/pendulum-chain/spacewalk/blob/ea66a3062e6edf05a4ed7b377f107e1f95b688f1/primitives/src/lib.rs#L519-L535)

bogdanS98:
Closes #522.

### DIFF
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -440,12 +440,10 @@ impl CurrencyInfo for Asset {
 	fn name(&self) -> &str {
 		match self {
 			Asset::StellarNative => "Stellar",
-			Asset::AlphaNum4 { code, issuer: _ } => {
-				from_utf8(&remove_trailing_non_alphanum_bytes(code)).unwrap_or("unspecified")
-			},
-			Asset::AlphaNum12 { code, issuer: _ } => {
-				from_utf8(&remove_trailing_non_alphanum_bytes(code)).unwrap_or("unspecified")
-			},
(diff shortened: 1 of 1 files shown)


ebma (commented):
Can you please run `cargo +nightly-2024-02-09 fmt --all -- --check`?

I had a similar problem in [this](https://github.com/pendulum-chain/spacewalk/actions/runs/8999264317/job/24726581991) action. We should probably change our pre-commit action somehow @b-yap 

Besides that, can we please also add the decimals for the HDX token (XCM: 8, with 12 decimals) and already in anticipation of the vDOT token (probably XCM: 9, with 10 decimals) to the list?

ebma (approved):
Looks good to me, let's wait for the CI

Issue #482, opened 2024-01-26
Add Stellar testnet archive URLs to vault client config


ebma:
Apparently, the Stellar testnet validators do have their own archives. They can be found here:
- `history.stellar.org/prd/core-testnet/core_testnet_001`
- `history.stellar.org/prd/core-testnet/core_testnet_002`
- `history.stellar.org/prd/core-testnet/core_testnet_003`

## TODO 
- In the code, we so far assumed that there are no archives for Stellar testnet. We should change this assumption and remove any code that prevents restoring proof messages from archives on Stellar testnet.
- We should also change our files for the Stellar overlay configuration on testnet to include these archive URLs.

prayagd:
@ebma is this required for the upcoming pendulum launch?

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/482?planning-poker) with Zenhub @b-yap @bogdanS98 @ebma @gianfra-t @TorstenStueber

ebma:
No, it's not required.

prayagd:
Should i move this to icebox then?

ebma:
This is not required for the upcoming launch but it helps us test the vaults on Foucoco in a setting that is closer to the Amplitude/Pendulum networks because so far we couldn't test the use of Stellar archives on Foucoco. That's why I wouldn't move it to the icebox and instead give it a medium priority.

PR #524, opened 2024-05-08
482 add stellar testnet archive urls to vault client config

gianfra-t:
Regarding #482 

### About
Adds archive url's for Stellar testent, and code changes such that we don't discriminate anymore between testnet and mainnet for fetching this values.

### Notable changes
- Previously we used [this](https://github.com/pendulum-chain/spacewalk/pull/524/files#diff-620b801d9005e873daac78ea8504b5d049853aeba6712e346690c05238f5445fL109) condition to skip fetching envelopes from an archive node if the vault was running on testnet. Now the behavior is the same for both testnet and public.
- Transforming the slot index to hex to form the archive url had to be slightly [modified](https://github.com/pendulum-chain/spacewalk/pull/524/files#diff-411a309c26319ce6a3f8022c8d4d325b38530360cf7a728a0a6a0edc7f9aae00R119) such that a full 8 bytes representation of the slot index is formed, padding the most significant bytes with 0 if necessary. This was not necessary before since mainnet indexes are currently large enough that they use 8 bytes without padding, which is not the case for testnet. (See implementation [here](https://github.com/pendulum-chain/scp-messages/blob/35be736facffc17b21a18a4b32b0d034f56948a7/src/archiveFile.ts#L39)).
- Adds two tests that mirror the public counterpart to test fetching and opening transaction sets and envelopes.
- Adds the archive url's to the configuration files.


gianfra-t:
@pendulum-chain/devs beyond the simple unit tests added, does anyone know if there is a configuration by which we can trigger in the integration tests a situation in which the client will try to look for these archives?

gianfra-t:
Thanks for pointing that out @b-yap ! I replaced [here](https://github.com/pendulum-chain/spacewalk/blob/1c3ea1c163e721267e45a8f8baae3b4964df6ff3/clients/vault/tests/vault_integration_tests.rs#L610) such that the testnet would be used and it works (locally).

I didn't add a new `test_issue_execution_succeeds_from_archive` just for testnet since I think it may be redundant, but we could.

b-yap:
hhhm, something wrong with the rustfmt; it's reverted back.  I tried to do `cargo +nightly-2024-02-09` and it accepts `return`s after a `;`

b-yap:
[`trailing_semicolon` is an `unstable` option](https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#trailing_semicolon), and so are the others. Probably makes sense why it works only now and then.

[Torsten commented on this](https://github.com/pendulum-chain/spacewalk/pull/513#issuecomment-2075256268) before. I didn't want the reformatting to overlap with this PR.

gianfra-t:
Oh okay, sorry! So we have to be careful and leave the format to the pre-commit hook @b-yap? Signing the commits again no problem.

ebma:
@gianfra-t and @b-yap I created a new issue https://github.com/pendulum-chain/spacewalk/issues/525 for the formatting. I already noticed this problem last week and it's quite annoying that we now have this mismatch between the formatting of the precommit hook vs what the CI expects.

> if the archives are already available in testnet, should we revert the testchain to only have 1? It could reduce the code.

@b-yap I don't think we should. The main reason I created the tests is to be able to replicate issues that only seem to occur on Stellar mainnet. Being able to test on both Stellar test and main network is quite valuable IMO.

> I didn't add a new test_issue_execution_succeeds_from_archive just for testnet since I think it may be redundant, but we could.

True, it's redundant but I'd say we add the test case anyway. We can afford to have the CI run 3 extra minutes in order to be safe that restoring messages from archives works as expected. In the future it could happen that the SDF introduces some changes that break our logic and as they will introduce them on testnet first, we would be able to catch it earlier.

clients/stellar-relay-lib/resources/config/testnet/stellar_relay_config_sdftest1.json
```diff
@@ -10,5 +10,7 @@
     "version_str": "stellar-core 20.2.0.rc3 (34d82fc00426643e16b7ad59c9fde169b778eb4b)",
     "is_pub_net": false
   },
-  "stellar_history_archive_urls": []
+  "stellar_history_archive_urls": [
+    "http://history.stellar.org/prd/core-testnet/core_testnet_001"
```
b-yap, line 14: any reason why we don't include `core_testnet_002` and `core_testnet_003`?

gianfra-t: No particular reason, in fact I thought about adding all of them, but then looking at the analogous `stellar_relay_config_mainnet_iowa.json` I see that it only uses [one](https://github.com/pendulum-chain/spacewalk/blob/189765ca732396900fb4b49caa1ab6e82c7a9180/clients/stellar-relay-lib/resources/config/mainnet/stellar_relay_config_mainnet_iowa.json#L14) of the mainnet archives. 

You think every config file should have all the possible archive url's @b-yap, including for mainnet?

gianfra-t: I guess the only argument in favor of separating them could be to have a more fine-grained control in case one of them is not working, so that the selection is not random when trying to fetch.

clients/vault/tests/vault_integration_tests.rs
```diff
@@ -608,6 +608,18 @@ async fn test_issue_cancel_succeeds() {
...
+
+#[tokio::test(flavor = "multi_thread")]
+#[ntest::timeout(1_200_000)] // timeout at 20 minutes
+#[serial]
+async fn test_issue_execution_succeeds_from_archive_testnet() {
+	let is_public_network = false;
+	do_test_issue_execution_succeeds_from_archive(is_public_network).await;
+}
+
+async fn do_test_issue_execution_succeeds_from_archive(is_public_network: bool){
```
ebma, line 622: Nit: Adhering to the naming schema of the other functions let's call them

```suggestion
#[serial]
async fn test_issue_execution_succeeds_from_archive_on_mainnet() {
	let is_public_network = true;
	test_issue_execution_succeeds_from_archive_on_network(is_public_network).await;
}

#[tokio::test(flavor = "multi_thread")]
#[ntest::timeout(1_200_000)] // timeout at 20 minutes
#[serial]
async fn test_issue_execution_succeeds_from_archive_on_testnet() {
	let is_public_network = false;
	test_issue_execution_succeeds_from_archive_on_network(is_public_network).await;
}

async fn test_issue_execution_succeeds_from_archive_on_network(is_public_network: bool){
```

clients/stellar-relay-lib/resources/config/testnet/stellar_relay_config_sdftest1.json
ebma: I think this is just some legacy fragment, because in the very beginning we only knew about one Stellar archive URL. By now, we know 'all' of the official ones and there is no good reason not to add all of them to the array. 

It might be least confusing if we streamline this across all the configuration files and make each of them list all archives (for the respective network only, of course).

ebma (approved):
Looks good to me too 👍

clients/stellar-relay-lib/resources/config/testnet/stellar_relay_config_sdftest1.json
gianfra-t: Alright, also makes sense. Added [here](https://github.com/pendulum-chain/spacewalk/pull/524/commits/af6cb702eb0a903253f8cb0f4035ada4de767a08).
We could also add `stellar-history` as well as `core-live` but seems to me a bit confusing.

ebma: Hmm.. let's maybe still have all 6 in the mainnet configurations to signify that all of them can be used.

b-yap (approved):

PR #536, opened 2024-06-04
Update to Polkadot v1.1.0

gianfra-t:
Issue #534.

### Changes in Pallets
- [Removal](https://github.com/paritytech/polkadot-sdk/commit/5e7b27e98c0b9dacd4b6f0724fd6335106d94101 ) of `Header` and `BlockNumber` in `frame_system`.
- [Replacement](https://github.com/paritytech/substrate/commit/c446786a498dee7b413de101efc3c339d3eddf44) of type `Index` with `Nonce`.
- Change the [macro](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-945cf6fcb57d25950f6c53bcd090426caa4c1b08b3a32d1da6670febd283e5abR166) use for defining the `Default` trait of the `GenesisConfig` struct with a manual definition, to fix an error on the macro `construct_runtime` when using the previous `#[derive(default)]` macro.
- `Default` trait has to be defined for all `GenesisConfig` structs, not just when `std` [feature](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-7518d359d3767fb4dbef7f14346021b42b4716ca60a4769f9bf0b9238b06f71bL197) is enabled.
- `construct_runtime` [can](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-740639f23487a48fa0e536379f3854a530e701ebd939b28d6a8dd765e65e8435R26) be used without defining the components of the module. This is less error prone since hard to debug errors may appear when missing a component. (If we are okay with this change, I can later modify the rest).
- [Replace](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-7518d359d3767fb4dbef7f14346021b42b4716ca60a4769f9bf0b9238b06f71bL220) the use of the to-be-deprecated `GenesisBuild`.
- New [field](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-60b7dc5d741848eed1c1d09ec5f08c7fb7b1bef7236b441f8490b51a8ff5228eR173) in `pallet_aura` config. `AllowMultipleBlocksPerSlot` should be `false` if used in normal scenario (not augmented with a custom pallet).
- New [field](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-60b7dc5d741848eed1c1d09ec5f08c7fb7b1bef7236b441f8490b51a8ff5228eR183) in `pallet_grandpa` `MaxNominators`. The value is set at `1000`, previously this was set by [default](https://github.com/paritytech/polkadot-sdk/commit/50a63c30d9622db9818329e3c2f30c52f09e6e19#diff-73a7edd851fb67b6f58ecb8eaae73b70ebcb9eb96c00212218621c4227912c9cL38) as `200`. This value is used for weight calculations.

### Node changes

Changes made using as base the `substrate-node-template` from the developer hub, at a [commit](https://github.com/substrate-developer-hub/substrate-node-template/blob/e6bf90cfbf774fc284056068fe75e3e97eeb0306/node/Cargo.toml) using version `1.0.0.`

Most relevant change is how the off-chain worker is spawned. [Here](https://github.com/substrate-developer-hub/substrate-node-template/commit/e6bf90cfbf774fc284056068fe75e3e97eeb0306#diff-594440db8adfc5d1b37711309312bfa06b49e488309186ad062a896778cf9fbfR184) is the change for the template between the relevant version.

### Other code changes

Modification of [parameters](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-e90b5a1f9c1d8e825b7761df51b334f100d48612ad6b4335aed8c4adf842c78cL242) of `SubxtClientConfig` . 
- `rpc_max_response_size` is defined as the [default](https://github.com/techwizard210/Substrate_Parity/blob/367dab0d4bd7fd7b6c222dd15c753169c057dd42/client/cli/src/commands/run_cmd.rs#L111) value, in megabytes.
- `rpc_max_request_size` is [deprecated](https://github.com/techwizard210/Substrate_Parity/blob/367dab0d4bd7fd7b6c222dd15c753169c057dd42/client/cli/src/commands/run_cmd.rs#L106C6-L106C21) and has no effect.

### Subxt update
`subxt` had to be updated from `0.25.0` to `0.28.0` due to a conflict with `schnorkell` package. This is the lowest version that solves that issue. Nevertheless, this new version introduced some breaking changes that have to be handled:
- flag `substrate-compat` has to be [used](https://github.com/paritytech/subxt/commit/b316301d61f924c510c40f505995f28a83c1365f) now for many of the functions and exports regarding substrate compatibility. 
- Pallet and error information of `ModuleError` are now accessed [differently](https://github.com/paritytech/subxt/blob/6e51b683974664c876bbd72882b863a7355b4b24/CHANGELOG.md#dispatch-error-changes-878), which impacts [this](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-6469b7c5aed83b33ec6bbd2a61154190361c333995a98a3c172436491a901b9bR107) match.
- Introduction of [dynamic types.](https://github.com/paritytech/subxt/blob/6e51b683974664c876bbd72882b863a7355b4b24/CHANGELOG.md#unify-how-we-encode-and-decode-static-and-dynamic-types-842). This mostly affects the way in which we replace types define on the metadata with type defined in the code. Types that do not implement `EncodeAsType` and `DecodeAsType` now should be wrapped in a `Static` struct, which is the case with `CurrencyId` and all the corresponding uses of it on the code. 
To avoid wrapping as static the type `AccountId`, we [use](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-aa6c3e8e09bb98599091d49cb6e143f17d7d9ec031d8450ff57fa9d925bd9afaR10) the `AccountId` type from `subxt`. 
- The last point introduces the need to transform from `Keyring` to subxt's `AccountId`, which is done now as in the package [tests](https://github.com/paritytech/subxt/blob/6196d26cede282b13ea35a8d20483c0818d88175/core/src/utils/account_id.rs#L217). Here is an [example](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-a6bab8ae5f7b0a5c8cc8e61bf10f67add3b01635768e3b0e0270d1ae11fdc8baR85) on our code.
- Since the implementation of `AccountId` from `subxt` does not have some of the traits required, we need to transform to the `sp` version of it before using that functionality, a good example is [here](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-aa6c3e8e09bb98599091d49cb6e143f17d7d9ec031d8450ff57fa9d925bd9afaR280) where we use need `sp_runtime` version for hashing.


TorstenStueber:
Thanks for the detailed description.

> construct_runtime [can](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-740639f23487a48fa0e536379f3854a530e701ebd939b28d6a8dd765e65e8435R26) be used without defining the components of the module

I think this is fine as also the Polkadot relaychain runtime defines it that way.

.github/workflows/ci-main.yml
```diff
@@ -1,16 +1,25 @@
 on:
   pull_request:
     branches:
-      - 'main'
+      - 'main' # excluding main
```
TorstenStueber, line 4: Doesn't mean "only main" instead of "excluding main"?

clients/runtime/src/conn.rs
```diff
@@ -12,23 +12,24 @@ const RETRY_TIMEOUT: Duration = Duration::from_millis(1000);
...
 	url: &str,
 	max_concurrent_requests: Option<usize>,
-	max_notifs_per_subscription: Option<usize>,
+	_max_notifs_per_subscription: Option<usize>,
 ) -> Result<Client, Error> {
 	let (sender, receiver) = ws_transport(url).await?;
 	let ws_client = ClientBuilder::default()
 		.request_timeout(CONNECTION_TIMEOUT)
 		.max_concurrent_requests(max_concurrent_requests.unwrap_or(1024))
-		.max_notifs_per_subscription(max_notifs_per_subscription.unwrap_or(256))
```
TorstenStueber, line 31: Why remove this?

clients/runtime/src/lib.rs
```diff
@@ -64,41 +67,39 @@ compile_error!("You need to select at least one of the metadata features");
...
-	)
-)]
-pub mod metadata {
-	#[subxt(substitute_type = "sp_core::crypto::AccountId32")]
-	use crate::AccountId;
-	#[subxt(substitute_type = "spacewalk_primitives::CurrencyId")]
-	use crate::CurrencyId;
-	#[subxt(substitute_type = "sp_arithmetic::fixed_point::FixedU128")]
-	use crate::FixedU128;
+// #[cfg_attr(
```
TorstenStueber, line 81: Is this note required anymore? In that case can we remove the comments?

clients/runtime/src/retry.rs
```diff
@@ -34,15 +34,17 @@ where
 {
 	let mut backoff = get_exponential_backoff();
 	loop {
+		println!("attempting...");
```
TorstenStueber, line 37: Why adding this unrelated changes to the PR for the upgrade to Polkadot v1.1.0? Can we keep all changes in this PR strictly for upgrading to v1.1.0?

clients/runtime/src/tests.rs
```diff
@@ -106,22 +109,52 @@ async fn test_subxt_processing_events_after_dispatch_error() {
...
-		AccountKeyring::Alice.into(),
-		DEFAULT_TESTING_CURRENCY,
-		DEFAULT_WRAPPED_CURRENCY,
-	);
-
-	parachain_rpc.register_public_key(dummy_public_key()).await.unwrap();
-	parachain_rpc.register_vault(&vault_id, 3 * 10u128.pow(12)).await.unwrap();
-	parachain_rpc.get_vault(&vault_id).await.unwrap();
-	assert_eq!(parachain_rpc.get_public_key().await.unwrap(), Some(dummy_public_key()));
+    env_logger::init_from_env(
```
TorstenStueber, line 114: This code change (this and the following lines) seems to go beyond upgrading to v1.1.0, why is that?

primitives/src/lib.rs
```diff
@@ -139,7 +140,7 @@ pub mod issue {
 	}
 
 	// Due to a known bug in serde we need to specify how u128 is (de)serialized.
-	// See https://github.com/paritytech/substrate/issues/4641
+	// See https://github.com/paritytech/polkadot-sdk/issues/4641
```
TorstenStueber, line 143: I think that the replacement here (and in two more occurrences in this file below) was not correct as this would now point at a different issue.

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel    = "1.77.0"
+channel    = "nightly-2024-02-09"
```
TorstenStueber, line 2: Is it possible to use the same stable toolchain as for Pendulum?

testchain/docs/rust-setup.md
```diff
@@ -157,7 +157,7 @@ runtimes. You will need to configure your Rust compiler to use
 compile Substrate runtime code to the Wasm target.
 
 > There are upstream issues in Rust that need to be resolved before all of Substrate can use the stable Rust toolchain.
-> [This is our tracking issue](https://github.com/paritytech/substrate/issues/1252) if you're curious as to why and how this will be resolved.
+> [This is our tracking issue](https://github.com/paritytech/polkadot-sdk/issues/1252) if you're curious as to why and how this will be resolved.
```
TorstenStueber, line 160: This replacement is also incorrect as it points to a different issue now.

testchain/runtime/mainnet/src/lib.rs
```diff
@@ -623,15 +621,12 @@ impl pooled_rewards::Config for Runtime {
...
 
 construct_runtime! {
-	pub enum Runtime where
-		Block = Block,
-		NodeBlock = primitives::Block,
-		UncheckedExtrinsic = UncheckedExtrinsic
+	pub enum Runtime 
 	{
-		System: frame_system::{Pallet, Call, Storage, Config, Event<T>} = 0,
+		System: frame_system::{Pallet, Call, Storage, Config<T>, Event<T>} = 0,
```
TorstenStueber, line 626: We could then also remove the explicit parameters (e.g., `{Pallet, Call, Storage, Config<T>, Event<T>}`) in this `construct_runtime!`.

testchain/runtime/testnet/src/lib.rs
```diff
@@ -613,15 +612,12 @@ impl pooled_rewards::Config for Runtime {
...
 
 construct_runtime! {
-	pub enum Runtime where
-		Block = Block,
-		NodeBlock = primitives::Block,
-		UncheckedExtrinsic = UncheckedExtrinsic
+	pub enum Runtime 
 	{
-		System: frame_system::{Pallet, Call, Storage, Config, Event<T>} = 0,
+		System: frame_system::{Pallet, Call, Storage, Config<T>, Event<T>} = 0,
```
TorstenStueber, line 617: And here as well.

TorstenStueber (commented):
The changes in the alphabetically later part of this PR are quite straight forward, however, in the earlier part it looks to me that at some places there are changes that go beyond an upgrade to version v1.1.0. Is that PR here based on some other PR or is some other PR added into it? I left some comments at respective places.
 
E.g., there are a few changes in the CI files that seem to be unrelated to the upgrade to v1.1.0. Why is that?

Please also format the code.

testchain/runtime/mainnet/src/lib.rs
gianfra-t: Let's do that!

gianfra-t:
@TorstenStueber Thanks for the comments. Many of the changes you pointed out are temporary (like logs, update to the toolchain, etc) because there is an issue with one of the unit tests that is only failing on ubunutu (or linux in general as far as I could tell), but not in mac-os as I commented [here](https://github.com/pendulum-chain/spacewalk/pull/536#issuecomment-2153360127). Sorry the comment was not detailed enough.

One of the tests in questions is `test_register_vault` which I have been using for testing a solution. For now, it seems the solution is to manually finalize blocks to "unstuck" the test. It still requires a bit of exploration to define a more robust workaround.

clients/runtime/src/retry.rs
gianfra-t: This is one of the temporary logs to debug the ubuntu-only issue.

clients/runtime/src/tests.rs
gianfra-t: Also temporary.

clients/runtime/src/lib.rs
gianfra-t: Apparently with the new `subxt` version, the macro will fail when `all--features` flag is passed. 
This didn't happen before.

 I left this commented to discuss this because as the [previous comment](https://github.com/pendulum-chain/spacewalk/pull/536/files/4d27e969bbe1502c6a90e2cd90d7345602238c14#diff-19f5b8d58a7b3276f119607fd62317f9fde6fe5f4d88e8b639b7404fd661c9c2R68) above said, we are always using foucoco metadata. I imagine this was because this was the last call to [the](https://github.com/pendulum-chain/spacewalk/pull/536/files/4d27e969bbe1502c6a90e2cd90d7345602238c14#diff-19f5b8d58a7b3276f119607fd62317f9fde6fe5f4d88e8b639b7404fd661c9c2R95) macros. So with this changes we would be using only the `standalone` metadata for every case. I am not sure if this is correct, or we still need to differentiate. In the later case, we would need to tweak a bit more the conditionals.

primitives/src/lib.rs
gianfra-t: You're correct. It was incorrectly replaced.

.github/workflows/ci-main.yml
gianfra-t: Yes of course, in any case this comment is from copying the `dev` and only temporary.

clients/runtime/src/error.rs
```diff
@@ -104,9 +104,8 @@ impl Error {
 	pub fn is_module_err(&self, pallet_name: &str, error_name: &str) -> bool {
 		matches!(
 			self,
-			Error::SubxtRuntimeError(SubxtError::Runtime(DispatchError::Module(ModuleError{
-				pallet, error, ..
-			}))) if pallet == pallet_name && error == error_name,
+			Error::SubxtRuntimeError(SubxtError::Runtime(DispatchError::Module(module_error
+			))) if module_error.details().unwrap().pallet.name() == pallet_name && &module_error.details().unwrap().variant.name == error_name,
```
ebma, line 108: Let's refactor this statement so that we don't have to unrwap() twice and it becomes easier to read.

clients/runner/src/runner.rs
```diff
@@ -309,7 +309,7 @@ impl Runner {
 			.storage()
 			.at_latest()
 			.await?
-			.fetch_raw(&lookup_bytes)
+			.fetch_raw(&*lookup_bytes)
```
ebma, line 312: Should we just clone then?
```suggestion
			.fetch_raw(lookup_bytes.clone())
```

clients/runtime/src/rpc.rs
```diff
@@ -301,25 +342,24 @@ impl SpacewalkParachain {
...
 
 	/// Subscribe to new parachain blocks.
 	pub async fn on_block<F, R>(&self, on_block: F) -> Result<(), Error>
 	where
 		F: Fn(SpacewalkHeader) -> R,
 		R: Future<Output = Result<(), Error>>,
-	{
-		let mut sub = self.api.rpc().subscribe_finalized_block_headers().await?;
+	{	
+		let mut sub = self.legacy_rpc.chain_subscribe_finalized_heads().await?;
```
ebma, line 362: If functions like this are considered legacy, is there a better practice now that we should rather use instead?

clients/stellar-relay-lib/Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde_json = "1.0.93"
 serde_with = "2.2.0"
 
 
-substrate-stellar-sdk = {git = "https://github.com/pendulum-chain/substrate-stellar-sdk", branch = "polkadot-v0.9.42", default-features = false, features = ['all-types']}
+substrate-stellar-sdk = {git = "https://github.com/pendulum-chain/substrate-stellar-sdk", default-features = false, features = ['all-types']}
```
ebma, line 33: Let's also declare a `branch` here.

clients/vault/src/main.rs
```diff
@@ -131,7 +132,7 @@ async fn start() -> Result<(), ServiceError<Error>> {
 	// This file is auto-removed when `drop`ped.
 	let _pidfile = PidFile::create(
 		&String::from(DEFAULT_SPEC_NAME),
-		signer.read().await.account_id(),
+		&AccountId::new(signer.read().await.account_id().0),
```
ebma, line 135: Nit: It's not clear what the outcome of this statement is so let's maybe assign it to an extra variable first.

clients/vault/src/metrics.rs
```diff
@@ -497,7 +497,7 @@ pub async fn monitor_bridge_metrics(
 				for currency_id in updated_currencies {
 					for vault in vaults
 						.iter()
-						.filter(|vault| &vault.vault_id.collateral_currency() == currency_id)
+						.filter(|vault| vault.vault_id.collateral_currency() == **currency_id)
```
ebma, line 500: Are two asterisks needed?

clients/runtime/src/error.rs
gianfra-t: We can use an inner match maybe.

clients/runtime/src/rpc.rs
gianfra-t: It seems from the [CHANGELOG](https://github.com/paritytech/subxt/blob/master/CHANGELOG.md#0320---2023-09-27) that it is the "well maintained" one, while the other is `UnstableBackend`.

clients/vault/src/metrics.rs
gianfra-t: I think so, because it is a reference to `Static<CurrencyId>`, the second de-reference will get the inner value of the Static wrapper.

gianfra-t:
@pendulum-chain/devs Regarding the failing tests `test_register_vault`and `test_subxt_processing_events_after_dispatch_error`, this is what I know so far:
- They started to fail after the update, and failed only on linux based systems.
- The failure mode is as follows: After the transaction is sent to the local testnet, we await for it to be included in the finalized block [here](https://github.com/pendulum-chain/spacewalk/blob/9b76de91aabf40e59fa0fa2c2b74cdf8437b7512/.github/workflows/ci-main.yml). This sometime will wait forever, even when the testnet continues to work correctly (apparently, based on logs) and produced new blocks. The test naturally fails due to a `Timeout` error.
- Updating `subxt` to different versions, up to current `0.33.0` didn't solve the issue. 
- Updating the dependencies of the node from `polkadot-v1.1.0` to `1.5.0` or higher seems to make the error more uncommon for `test_register_vault` but more common for `test_subxt_processing_events_after_dispatch_error`. This was all tested locally on an ubuntu (pop-os) desktop. It continues to be random when they fail, so it is hard to make a statistic about the issue.

We know that this must be a local-only issue since integration test are working on both operating systems.

ebma:
> We know that this must be a local-only issue since integration test are working on both operating systems.

If the integration tests are working reliably we must be able to fix this issue somehow as [one of the vault integration tests](https://github.com/pendulum-chain/spacewalk/blob/53f81a2dac91d523eaf7178d01a402a0fa536d55/clients/vault/tests/vault_integration_tests.rs#L31) is about registering the vault. I [added some changes](https://github.com/pendulum-chain/spacewalk/pull/536/commits/a5ca11274d7f6618ae90f1597b1f01718387b40a) in the hope that a closer replica of the code used by the integration test would help but this doesn't seem to be the case unfortunately.

clients/runtime/src/lib.rs
TorstenStueber: @ebma what do you think?

TorstenStueber:
@gianfra-t Understood now why there are all the extra changes. Let's maybe mark this PR as DRAFT as long as this is problem is not resolved.

Regarding the bug: it boils down to not being able to detect that a transaction is finalized? Is it possible to isolate this? E.g., have minimal code that submits the transaction and waits for it to be finalized and not being able to detect on Linux? That way, we might conclude that this is subxt bug and file a bug report.

gianfra-t:
Yes it boils to that. I think it is more likely the finalized event is not received (maybe even not sent?) rather than not detected on the `subxt` side. It should be possible to isolate, we may have to simplify the logic such that the code for the transaction and the `.await` on the transaction finality is all in the test itself for readability. The problem with reproducing this bug is that you must also have the parachain running, not to mention the fact that it fails only "half" the time.

@TorstenStueber regarding the extra temporary changes, I think I removed them in the lasts commits. Maybe I missed some of them. I will later make another pass at the full changes to look for that.

TorstenStueber:
How do you want to proceed? Testing this in isolation is quite some effort but I think we need to solve this before we can continue (or is it less crucial than it looks to me?) Is there any other less effort way to understand where the bug originates?

gianfra-t:
I'm not sure how to proceed. Luckily this is not something we see on the integration tests, which is a strong hint that this must be something only related to the test environment. It also makes this less crucial.

I think the less effort way may be to post an issue on the substrate or subxt repo to see if someone has encountered this before or is aware. Alternatively we can try to track where this finalized event is originated and log to see if it really is originated, I tried this but was unable to find the polkadot crate that is responsible for this on the node.

clients/runtime/src/lib.rs
ebma: I think the comment 
```
// All of the parachain features use the same metadata (from Foucoco) for now.
// We can change this once the spacewalk pallets were added to the runtimes of the other chains as
// well.
```
is stale and needs to be removed. I added it at a time where Spacewalk was only live on Foucoco, so that was the only metadata available. 

Let's try to figure out why exactly the subxt macro fails for `--all-features`. But we definitely need to keep differentiating the .scale files per runtime otherwise the vaults will probably not be able to connect to the respective network anymore.

ebma:
> Testing this in isolation is quite some effort but I think we need to solve this before we can continue (or is it less crucial than it looks to me?)

Since the integration tests work, I'd say it's not that crucial and we can assume that the vaults still work properly. So to unblock the further changes for upgrading we could also merge this without having those particular tests fixed and follow up with a new PR. We should do that soon after though. 

> I think the less effort way may be to post an issue on the substrate or subxt repo to see if someone has encountered this before or is aware. 

Sure, we can try. @gianfra-t would you be able to raise that ticket? Maybe someone can point us in the right direction.

b-yap:
> Sure, we can try. @gianfra-t would you be able to raise that ticket?

@gianfra-t just create a ticket, I can pick it up and try to find a fix too.

gianfra-t:
Sure, I will create that issue today, sorry I didn't yesterday. Do you think it's better to raise it on subxt or polkadot repo?

gianfra-t:
That sounds good. Discussing about the issue and the update is too much here!

gianfra-t:
@pendulum-chain/devs I submitted the issue [here](https://github.com/paritytech/subxt/issues/1650).

ebma:
Thanks @gianfra-t 👍 might be worth mentioning in the ticket that this only happens on Linux but works on macOS.

gianfra-t:
I totally forgot about that "little" fact... thanks for reminding me @ebma ! I will edit it at the end.

clients/runtime/client/src/lib.rs
```diff
@@ -239,35 +240,31 @@ impl<C: ChainSpec + 'static> SubxtClientConfig<C> {
...
-			execution_strategies: Default::default(),
 			force_authoring: Default::default(),
 			offchain_worker: Default::default(),
 			prometheus_config: Default::default(),
 			rpc_cors: Default::default(),
-			rpc_http: Default::default(),
-			rpc_ipc: Default::default(),
-			rpc_ws: Default::default(),
-			rpc_ws_max_connections: Default::default(),
+			rpc_max_connections: 1000,
```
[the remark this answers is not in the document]
gianfra-t: 🤔 I think we changed this to try if it solved the testing, it didn't. I will put it back to `100` like you found here.

clients/runtime/src/rpc.rs
```diff
@@ -59,10 +60,10 @@ cfg_if::cfg_if! {
 }
 
 // timeout before retrying parachain calls (5 minutes)
```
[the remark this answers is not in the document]
gianfra-t: Yes, first thing attempted. Going back to 5 since there is no reason to change it.

pallets/stellar-relay/src/traits.rs
```diff
@@ -1,7 +1,7 @@
 use codec::{Decode, Encode, MaxEncodedLen};
-use frame_support::{traits::Get, BoundedVec, RuntimeDebug};
+use frame_support::{traits::Get, BoundedVec, pallet_prelude::RuntimeDebug};
 use scale_info::TypeInfo;
-use serde::{Deserialize, Serialize};
+use sp_runtime::serde::{Deserialize, Serialize};
```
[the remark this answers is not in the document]
gianfra-t: Although it's probably the same `serde`, we better use `serde` directly to avoid confusion like in all other pallets.

By the way, It took me a while to change this because I was getting errors by just changing this import, and turns out that a `cargo clean` had to be done.. Very strange.

clients/runtime/src/lib.rs
gianfra-t: I think the reason it fails is because when compiling with `--all-features`, it tries to apply the macro multiple times. Not sure why this was not an issue before `subxt` update. 

So I think this [changes](https://github.com/pendulum-chain/spacewalk/pull/536/commits/4409eef8a9870325ab9084a224d1232cb8508060) should solve this. I added an extra feature `all-features` that is not intended to do anything but be activated when `--all-features` is used. Then we can differentiate and select only one metadata, in this case standalone. As far as I know selecting all features in only used during testing right?

gianfra-t:
I have commented out the problematic [tests](https://github.com/pendulum-chain/spacewalk/pull/536/commits/0b4950ee6578acce4db544ac89a5397a0d4f83ac#diff-a6bab8ae5f7b0a5c8cc8e61bf10f67add3b01635768e3b0e0270d1ae11fdc8baR94) such that we can finish everything only leaving those as the last step. Let me know what you think.

clients/runtime/src/tests.rs
```diff
@@ -79,49 +88,54 @@ async fn test_too_low_priority_matching() {
...
-		DEFAULT_TESTING_CURRENCY,
-		DEFAULT_WRAPPED_CURRENCY,
-	);
-
-	parachain_rpc.register_public_key(dummy_public_key()).await.unwrap();
-	parachain_rpc.register_vault(&vault_id, 3 * 10u128.pow(12)).await.unwrap();
-	parachain_rpc.get_vault(&vault_id).await.unwrap();
-	assert_eq!(parachain_rpc.get_public_key().await.unwrap(), Some(dummy_public_key()));
-}
+// #[tokio::test(flavor = "multi_thread")]
```
ebma, line 97: Instead of commenting them out, let's just add the [ignore](https://doc.rust-lang.org/reference/attributes/testing.html#the-ignore-attribute) macro, possibly also with an explanation that these tests are currently broken on Ubuntu and will be fixed later.

.github/workflows/ci-main.yml
```diff
@@ -19,37 +19,30 @@ jobs:
...
-      - name: Remove rust-toolchain.toml for nightly
-        if: matrix.rust == 'nightly'
-        # To make sure that the nightly version will be used all throughout
-        run: |
-          rm /home/runner/work/spacewalk/spacewalk/rust-toolchain.toml
+      # - name: Remove rust-toolchain.toml for nightly
+      #   if: matrix.rust == 'nightly'
+      #   # To make sure that the nightly version will be used all throughout
+      #   run: |
+      #     rm /home/runner/work/spacewalk/spacewalk/rust-toolchain.toml
```
ebma, line 45: Are we sure that this is not needed anymore? Or did you remove it because the path in l. 45 was not found?

b-yap (approved):
I have no more comments.

Approve, as long as CI passes.

gianfra-t:
@pendulum-chain/devs I don't think the test will pass with ubuntu. I think we will have to migrate to mac-os in order to solve the issue explored on this [pr](https://github.com/pendulum-chain/spacewalk/pull/540).

@b-yap I can see that you were able to fix and make the CI work with mac-os, and that the [test](https://github.com/pendulum-chain/spacewalk/actions/runs/10006612859/job/27662807591) passed successfully. Yet at least the last attempt with ubuntu [failed](https://github.com/pendulum-chain/spacewalk/actions/runs/9661120599/job/27255531817?pr=536) due to the timeout error.

.github/workflows/ci-main.yml
gianfra-t: No, it was required. was commented out just to test but has been re-added.

clients/runner/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "runner"
-version = "1.0.10"
+version = "1.0.9"
```
ebma, line 3: We shouldn't turn back the version.

clients/runtime/src/lib.rs
```diff
@@ -111,13 +139,12 @@ pub struct WrapperKeepOpaque<T> {
 pub struct SpacewalkRuntime;
 
 impl Config for SpacewalkRuntime {
-	type Index = Index;
-	type BlockNumber = BlockNumber;
+	type AssetId = ();
```
ebma, line 142: What is the `AssetId` used for? Is it okay to leave it empty?

clients/runtime/src/types.rs
```diff
@@ -168,13 +171,11 @@ pub trait PrettyPrint {
...
 
 mod account_id {
-	use subxt::ext::sp_core::crypto::Ss58Codec;
-
 	use super::*;
 
 	impl PrettyPrint for AccountId {
 		fn pretty_print(&self) -> String {
-			self.to_ss58check_with_version(SS58_PREFIX.into())
+			"self.into()".to_string()
```
ebma, line 178: Is this a missing TODO?

clients/vault/src/oracle/storage/impls.rs
```diff
@@ -427,7 +427,7 @@ mod test {
 			.expect("should have an archive url")
 			.clone();
 		let scp_archive_storage = ScpArchiveStorage(testnet_archive_url);
-		let slot_index = 10;
+		let slot_index = 235001;
```
ebma, line 430: This also has to be a regression. We'd prefer using a lower slot

clients/vault/src/oracle/storage/traits.rs
```diff
@@ -163,11 +163,8 @@ pub trait ArchiveStorage {
 
 pub(crate) async fn download_file_and_save(url: &str, file_name: &str) -> Result<(), Error> {
 	let response = reqwest::get(url).await.map_err(|e| Error::ArchiveError(e.to_string()))?;
-	if response.status().is_server_error() | response.status().is_client_error() {
-		return Err(Error::ArchiveResponseError(format!("{response:?}")));
-	}
-
```
ebma, line 169: Do we really want to remove this?

clients/vault/src/system.rs
```diff
@@ -946,16 +942,4 @@ impl VaultService {
...
-		let opt_agent = self.agent.clone();
-		self.agent = None;
-
-		if let Some(arc_agent) = opt_agent {
-			tracing::info!("try_shutdown_agent(): shutting down agent");
-			arc_agent.shutdown().await;
-		} else {
-			tracing::debug!("try_shutdown_agent(): no agent found");
-		}
-	}
```
ebma, line 960: Same here

clients/wallet/src/mock.rs
```diff
@@ -133,16 +133,7 @@ pub fn wallet_with_secret_key_for_storage(
...
-	_public_key(dest_secret)
-}
-
-pub fn default_source() -> PublicKey {
-	let source_secret = get_source_secret_key_from_env(IS_PUBLIC_NETWORK);
-	_public_key(source_secret)
-}
-
-fn _public_key(secret: String) -> PublicKey {
-	let dest_secret_key = SecretKey::from_encoding(secret).expect("should work");
```
ebma, line 145: Do we not need a default source anymore?

clients/wallet/src/operations.rs
```diff
@@ -199,8 +197,16 @@ pub mod redeem_request_tests {
...
+		"GA5ZSEJYB37JRC5AVCIA5MOP4RHTM335X2KGX3IHOJAPP5RE34K4KZVN";
+	const DEFAULT_DEST_PUBLIC_KEY: &str =
+		"GBWYDOKJT5BEYJCCBJRHL54AMLRWDALR2NLEY7CKMPSOUKWVJTCFQFYI";
+
 	fn default_testing_stellar_pubkeys() -> (PublicKey, PublicKey) {
-		(default_source(), default_destination())
+		(
+			public_key_from_encoding(DEFAULT_SOURCE_PUBLIC_KEY),
+			public_key_from_encoding(DEFAULT_DEST_PUBLIC_KEY),
+		)
```
ebma, line 209: Also regression?

pallets/issue/src/benchmarking.rs
```diff
@@ -28,17 +29,12 @@ fn deposit_tokens<T: crate::Config>(
 	account_id: &T::AccountId,
 	amount: BalanceOf<T>,
 ) {
-	assert_ok!(<orml_currencies::Pallet<T>>::deposit(currency_id, account_id, amount));
+	assert_ok!(<orml_tokens::Pallet<T>>::deposit(currency_id, account_id, amount));
```
ebma, line 32: Regression. Using `orml_currencies` here is important.

testchain/node/src/service.rs
```diff
@@ -24,6 +25,10 @@ use spacewalk_runtime_testnet::RuntimeApi as TestnetRuntimeApi;
 // Native executor instance.
 pub struct TestnetExecutor;
 
+/// The minimum period of blocks on which justifications will be
+/// imported and generated.
+const GRANDPA_JUSTIFICATION_PERIOD: u32 = 512;
```
ebma, line 30: Have we tried fiddling with this parameter? I found a [test](https://github.com/paritytech/polkadot-sdk/blob/5dc0670a85d575480e0840204c20a5771cd8f7d3/substrate/client/consensus/grandpa/src/tests.rs#L327) in the polkadot-sdk where it's defined to just 32. I honestly don't really understand the implications of this parameter, just something that came to my mind.

ebma (commented):
We need to change the version from 1.0.9 back to 1.0.10 again. Also, I think at least one merge commit has to be faulty. At least I found a lot of changes that look like regressions and should be reverted. 

Why exactly do we need the new `all-features` feature?

clients/runner/Cargo.toml
gianfra-t: Sure! Must be as you say a faulty merge commit to main, because it is up-to-date now.

clients/vault/src/oracle/storage/impls.rs
gianfra-t: Yes, we [changed](https://github.com/pendulum-chain/spacewalk/issues/538) it after opening this PR.

clients/vault/src/oracle/storage/traits.rs
gianfra-t: I don't remember a reason to remove or touch this part. Probably part of the all same merge issue.

testchain/node/src/service.rs
gianfra-t: You mean for the test issue? No, this is the standard value [used](https://github.com/paritytech/polkadot-sdk/blob/dc4047c7d05f593757328394d1accf0eb382d709/templates/solochain/node/src/service.rs#L24) in one of the templates of the `polkadot-sdk`. Yet, for our instant seal this shouldn't matter since we are not using the consensus at all.

clients/runtime/src/lib.rs
gianfra-t: It is a new field from the `subxt` `Config` trait, but as the comment of this type [says](https://github.com/paritytech/subxt/blob/c3267ed488b0fcee5f537c60c341cc5f1c7d63e0/core/src/config/mod.rs#L61), I believe it is not relevant for our standalone testchain.

clients/runtime/src/types.rs
gianfra-t: It probably was 😅, or a very "convenient" typo. Let me see if the original implementation compiles.

gianfra-t:
@ebma the `--all-features` idea is used for filtering out the [application](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-19f5b8d58a7b3276f119607fd62317f9fde6fe5f4d88e8b639b7404fd661c9c2R83) of the metadata types multiple times when all the features are enabled. Apparently this was not a problem with the previous versions of `subxt` and it probably took the last one, but after the upgrade we got a collision since it tried to use the types from all the chains on the substitution.

Since when running the test like we generally do with the `--all-features` command it will as well enable this one. The name does not necessarily needs to match and could be anything, it will just be enabled due to the command as well.

pallets/security/src/types.rs
```diff
@@ -1,10 +1,10 @@
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sp_std::{cmp::Ord, fmt::Debug};
+use serde::{Deserialize, Serialize};
```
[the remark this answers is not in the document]
gianfra-t: Sorry for the stale comment. We are also [using](https://github.com/pendulum-chain/spacewalk/pull/536/files#diff-9218096256096db2d409218d53525ac1201febd7a2edca56c46ab982fd7794f8R4) `serde` in the pallet, you think it's okay?

gianfra-t:
@pendulum-chain/devs I wanted to comment on the test [issue](https://github.com/pendulum-chain/spacewalk/pull/540) we were facing a while ago, and summarize how we are dealing with it at the moment. 

As a summary of the attempts and experiments using extra logs in this [PR](https://github.com/pendulum-chain/spacewalk/pull/540#issue-2373594209) we identify that the issue happens when sealing a block on linux operating systems (especially in low spec CI runners), the [creation](https://github.com/pendulum-chain/polkadot-sdk/blob/522a0213cd9efbc79de073e182f25c0509987a8b/substrate/client/basic-authorship/src/basic_authorship.rs#L339) of a new block and the application of the [inherents](https://github.com/pendulum-chain/polkadot-sdk/blob/522a0213cd9efbc79de073e182f25c0509987a8b/substrate/client/basic-authorship/src/basic_authorship.rs#L341) consumes all the time specified by the whole process of applying extrinsics, therefore ending up hitting the [deadline](https://github.com/pendulum-chain/polkadot-sdk/blob/522a0213cd9efbc79de073e182f25c0509987a8b/substrate/client/basic-authorship/src/basic_authorship.rs#L453) condition before applying a single extrinsic. Please refer to the following example CI [run](https://github.com/pendulum-chain/spacewalk/actions/runs/10073204867/job/27849939845#step:11:2256), logs were set before and after each of these operations. We can see that we spend already 4 seconds on each by the timestamps.

Therefore, in order to workaround this issue, we modify this [constant](https://github.com/pendulum-chain/polkadot-sdk/blob/522a0213cd9efbc79de073e182f25c0509987a8b/substrate/client/consensus/manual-seal/src/seal_block.rs#L33) to a larger number (in this case, 60 seconds) such that the extrinsics have enough time to be applied. We can then use this forked `polkadot-sdk` repo only on the CI runs, by just [renaming](https://github.com/pendulum-chain/spacewalk/pull/536/commits/a2f19577b0032d420ed8f15fb26a0023c403db93#diff-a1a3cb9bdeb5541d148091d973cf266aa3b317e6415a86630e816cbe27cf8b9cR49) the `Cargo.toml` that contains the patch during the run.

ebma (approved):
The last CI job failed due to a shaky test so let's assume that it works generally. Thanks for all your efforts with this issue @gianfra-t. Let's merge this now.

Issue #538, opened 2024-06-12
Vault client indefinitely waiting for other tasks to shut down


ebma:
## Context

For some non-recoverable errors, the vault client tries to restart. Before restarting, it will wait for pending tasks to shut down. It seems like not all tasks are receiving the shutdown signal, or maybe they do but are still stuck. This causes the vault client to wait indefinitely (not even the periodic restart will work here).

The following incidents happened on lower-spec machines, so maybe they are more likely to occur when the clients don't have much resources available. 

## TODO

Try to find the tasks that are not successfully shut down. 

### Incident 1

```
root@ip-172-31-88-54 ~]# kubectl logs spacewalk-amplitude-tzs-05-65ccff6fdd-m5h9n  -n spacewalk -f
[2024-06-12T10:00:25Z INFO  runner] Connected to the parachain
[2024-06-12T10:00:25Z INFO  runner::runner] Downloading vault-amplitude at: "./vault-amplitude"
[2024-06-12T10:00:25Z INFO  runner::runner] Fetching executable from https://github.com/pendulum-chain/spacewalk/releases/download/v1.0.8/vault-amplitude
[2024-06-12T10:00:25Z INFO  runner::runner] Client started, with pid 14
Jun 12 10:00:25.886  INFO vault: Starting Prometheus exporter at http://0.0.0.0:9616
Jun 12 10:00:25.886  INFO Server::run{addr=0.0.0.0:9616}: warp::server: listening on http://0.0.0.0:9616
Jun 12 10:00:25.889  INFO vault::process: Creating PID file at: /tmp/amplitude_5H1NjwFazxmz1sLJU3Hv3K2HmkYPrEd6h3jZRYkPS4wUcjhN.pid
Jun 12 10:00:25.895  INFO service: Version: 1.0.8
Jun 12 10:00:25.895  INFO service: Vault uses Substrate account with ID: 6mpGcpNdmKttMkQG42VGhq1fU2gofz4uXA3XeoPuNtrnqZLM
Jun 12 10:00:25.895  INFO runtime::conn: Connecting to the spacewalk-parachain...    
Jun 12 10:00:26.079  INFO runtime::conn: Connected!    
Jun 12 10:00:26.417  INFO runtime::rpc: spec_name="amplitude"    
Jun 12 10:00:26.417  INFO runtime::rpc: spec_version=17    
Jun 12 10:00:26.417  INFO runtime::rpc: transaction_version=13    
[...]

PR #539, opened 2024-06-21
#538 shutdown the `OracleAgent` at `VaultService` level

b-yap:
Addresses #538. 

In the logs, the `try_shutdown_wallet()` logs just fine, so I plan to do the same thing for `OracleAgent`. 
To do this, the agent has to be _accessible_ at the `VaultService` level. 
So I added an`agent` field in `VaultService`. 

I removed subscription of the `ShutdownSender` because of its erratic behavior in the `OracleAgent`; sometimes it works, sometimes it doesn't.


clients/vault/src/oracle/agent.rs
```diff
@@ -123,24 +123,12 @@ pub async fn start_oracle_agent(
 			}
 		}
 
+		tracing::info!("start_oracle_agent(): shutting down overlay connection");
 		// shutdown the overlay connection
 		overlay_conn.stop();
 	});
 
-	tokio::spawn(on_shutdown(shutdown_sender.clone(), async move {
```
b-yap, line 130: This sometimes work; and sometimes it doesn't. The logs don't show up. 
I decided to remove this and let the `VaultService` perform the `disconnect_signal_sender` .

clients/vault/src/oracle/agent.rs
```diff
@@ -77,14 +80,11 @@ pub async fn start_oracle_agent(
...
 
 	let shutdown_sender_clone = shutdown_sender.clone();
-	// a clone used to forcefully call a shutdown, when StellarOverlay disconnects.
-	let shutdown_sender_clone2 = shutdown_sender.clone();
-
 	// disconnect signal sender tells the StellarOverlayConnection to close its TcpStream to Stellar
 	// Node
 	let (disconnect_signal_sender, mut disconnect_signal_receiver) = mpsc::channel::<()>(2);
 
-	service::spawn_cancelable(shutdown_sender_clone.subscribe(), async move {
```
b-yap, line 87: One less subscription, and it's either here or line 130 that's causing the 
`Waiting for 1 tasks to shut down`.

b-yap: This will also look better on _`tokio-console`_, as it won't point to the `spawn_cancelable` line of code

clients/vault/src/oracle/agent.rs
```diff
@@ -123,24 +123,12 @@ pub async fn start_oracle_agent(
...
-		if let Err(e) = disconnect_signal_sender.send(()).await {
-			tracing::warn!("start_oracle_agent(): failed to send disconnect signal: {e:?}");
-		}
-	}));
-
-	Ok(OracleAgent { collector, is_public_network, message_sender: Some(sender), shutdown_sender })
-}
-
-impl Drop for OracleAgent {
-	fn drop(&mut self) {
```
b-yap, line 141: Cannot use the Drop trait anymore, since `mpsc::channel::Sender` only works in async.

clients/vault/src/oracle/agent.rs
```diff
@@ -194,10 +182,10 @@ impl OracleAgent {
 	}
 
 	/// Stops listening for new SCP messages.
-	pub fn stop(&self) {
-		tracing::debug!("stop(): Shutting down OracleAgent...");
-		if let Err(e) = self.shutdown_sender.send(()) {
-			tracing::error!("stop(): Failed to send shutdown signal in OracleAgent: {:?}", e);
+	pub async fn shutdown(&self) {
```
b-yap, line 185: rename method to `shutdown`, as we are _truly_ shutting down the OracleAgent.

clients/vault/src/system.rs
```diff
@@ -273,6 +273,7 @@ pub struct VaultService {
 	shutdown: ShutdownSender,
 	vault_id_manager: VaultIdManager,
 	secret_key: String,
+	agent: Option<Arc<OracleAgent>>
```
b-yap, line 276: This way, the agent is accessible at the `VaultService` level.

clients/stellar-relay-lib/src/overlay.rs
```diff
@@ -59,6 +59,7 @@ impl StellarOverlayConnection {
 	pub fn listen(&mut self) -> Result<Option<StellarMessage>, Error> {
 		loop {
 			if !self.is_alive() {
+				info!("listen(): sender half of overlay has closed.");
```
ebma, line 62:
```suggestion
				debug!("listen(): sender half of overlay has closed.");
```

ebma (commented):
Cool, looks great 👍 

Looking at the failing CI tests, seems like we have a new problem with our archive restoration logic? Probably unrelated to this PR but a new general problem.

clients/vault/src/oracle/storage/impls.rs
```diff
@@ -427,7 +427,7 @@ mod test {
 			.expect("should have an archive url")
 			.clone();
 		let scp_archive_storage = ScpArchiveStorage(testnet_archive_url);
-		let slot_index = 235001;
```
ebma, line 430: How about we use an even smaller value? Maybe even slot 0? Then we are 100% resilient for future testnet resets.

clients/vault/src/oracle/storage/impls.rs
```diff
@@ -487,7 +487,7 @@ mod test {
 		let tx_archive_storage = TransactionsArchiveStorage(testnet_archive_url);
 
 		//arrange
-		let slot_index = 235001;
+		let slot_index = 219411;
```
ebma, line 490: Same here

ebma:
You tested that change locally @b-yap? How annoying.

b-yap:
@ebma Yes I tested it locally, but analyzed and found faults in the code, so I updated it. :+1:

clients/wallet/src/operations.rs
```diff
@@ -197,16 +199,8 @@ pub mod redeem_request_tests {
 
 	const IS_PUBLIC_NETWORK: bool = false;
 
-	const DEFAULT_SOURCE_PUBLIC_KEY: &str =
```
b-yap, line 200: @ebma I removed the hard-coded accounts

ebma: And it works with the other one? That's nice

ebma (approved):
Looks good to me 👍

clients/wallet/src/operations.rs
b-yap: Yes it works. All tests in this file will use the keys in the env.

clients/wallet/src/mock.rs
```diff
@@ -133,7 +133,16 @@ pub fn wallet_with_secret_key_for_storage(
 
 pub fn default_destination() -> PublicKey {
 	let dest_secret = get_dest_secret_key_from_env(IS_PUBLIC_NETWORK);
```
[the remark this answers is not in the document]
ebma: Cool 👍

PR #543, opened 2024-07-22
Add rustflags in cargo config

b-yap:
Defines the rustflags directly in `.cargo/config.toml` file. 
This is required for `tokio-console`. https://github.com/pendulum-chain/tasks/issues/343

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
 [build]
-CARGO_BUILD_RUSTFLAGS="--cfg tokio_unstable"
+rustflags = ["--cfg", "tokio_unstable"]


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+CARGO_BUILD_RUSTFLAGS="--cfg tokio_unstable"
```
ebma, line 2: Why are we not doing it like [this](https://github.com/pendulum-chain/spacewalk/pull/517/files/87ad4d58da5994c94e6a268bbe1dcf1e6661e9ca#diff-9a4f3e4537ebb7474452d131b0d969d89a51286f4269aac5ef268e712be17268)? What's the difference?

b-yap: Oh my god, good catch.

b-yap: I think it'll behave just the same.
 But yeah, I'll let it look similar to https://github.com/tokio-rs/console?tab=readme-ov-file#instrumenting-your-program

ebma (approved):
LGTM, let's merge.

PR #545, opened 2024-08-01
Rewrite Task of listening Stellar Messages

b-yap:
## Summary
We need to fix the vaults from getting unknowingly stuck, potentially caused by running unhandled zombie tasks.

 One idea is to include the polling of the stellar messages:
https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/vault/src/oracle/agent.rs#L100-L104
in the monitoring:
https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/vault/src/system.rs#L806-L817

This means updating the `OracleAgent`'s message_sender is delayed; passing the `OracleAgent` to tasks must be mutable;  hence using **_`Arc<RwLock<>>`_** instead of `Arc<>` alone.

But we **_cannot_** have these current tasks STARTING TOGETHER WITH the polling task. The Stellar-overlay _has to run already_, and all open requests MUST finish first.
https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/vault/src/system.rs#L787-L790

 An idea is to introduce another variant of the `ServiceTask`, where it waits for something to finish before a task starts. Prechecking will be required.
```
enum ServiceTask {
	...
	// Runs a task after a prequisite check has passed.
	PrecheckRequired(Task),
}
```

And to make sure the stellar-overlay and the client are communicating well, stellar-overlay will also send to client the:
* `hello` message  - useful to signal the client to prepare itself
* lib-related error  - this way the client will not wait forever, if stellar-overlay has issues.

------
#### `stellar-relay-lib`
  * Reduce clones in any way possible
      * use pass-by-value instead, to avoid cloning references
          * from `(&param)` to `(param)` 
      * reuse variables that were passed by reference
          * made obvious in `message_reader.rs` file 
      * clone only the needed field, instead of the whole structure
  * Return a lib-specific error and _hello_ message
      *  To signal the user (the agent) if the relay has started; and if the **_lib itself_** encountered a problem
  * Reposition timeouts and yields wisely
  * Let the user do the polling/looping
  
#### `vault`
* Removed `err-derive` dependency, and just use the existing `thiserror` dependency.
* Removed unused errors, like the RpcError
* `tokio_spawn` for naming tasks. This is only viable for `tokio_unstable`.
* Increase the event buffer capacity of the console-subscriber
* `OracleAgent` adjustment 
    *  Instead of `Arc<OracleAgent>`, it will be `Arc<RwLock<OracleAgent>>`
        * initializing the field `message_sender` is moved a bit later in the code; hence we need `OracleAgent` to be mutable.
* Listening for stellar messages will be monitored, alongside other tasks.
     * Every minute, a health check will be logged, to ensure messages coming from Overlay
     * When receiving hello message, flag the proof building.
        *  Continue to log messages, until the first slot was saved. Afterwards, this flag is not needed anymore.
* proof_builder.rs
     * Build proof ONLY if it's ready (if a slot was saved, meaning the overlay is running)
* system.rs
     * Added another variant of `ServiceTask`; the `PrecheckRequired`
         *  Executing open requests is required before running some tasks; once finished, these tasks will start.
             The following tasks need this precheck:
                * `issue::listen_for_issue_requests`
                * `issue::listen_for_issue_cancels`
                * `issue::listen_for_executed_issues`
                * `CancellationScheduler` for Issue
                * `CancellationScheduler` for Replace
                * `listen_for_replace_requests`
                * `listen_for_accept_replace`
                * `active_block_listener`
                     
I will add relevant comments when necessary.


clients/service/src/lib.rs
```diff
@@ -169,10 +169,22 @@ impl<Config: Clone + Send + 'static, F: Fn()> ConnectionManager<Config, F> {
...
+	shutdown_tx: ShutdownSender,
+	future2: F,
+	// a consumer that receives a precheck signal to start a task.
+	precheck_signal: Option<tokio::sync::broadcast::Receiver<()>>
+) -> Result<(), E>
 where
 	F: Future<Output = Result<(), E>>,
 {
+	if let Some(mut precheck_signal) = precheck_signal {
+		if let Err(e) = precheck_signal.recv().await {
```
b-yap, line 182: start the thread ONLY if precheck is finished.

clients/stellar-relay-lib/src/config.rs
```diff
@@ -35,13 +35,13 @@ impl StellarOverlayConfig {
 
 	#[allow(dead_code)]
 	pub(crate) fn node_info(&self) -> NodeInfo {
-		self.node_info.clone().into()
+		NodeInfo::new(&self.node_info)
```
b-yap, line 38: reduce clones

clients/stellar-relay-lib/src/connection/connector/connector.rs
```diff
@@ -340,7 +340,7 @@ mod test {
 			cert: new_auth_cert,
 			nonce: [0; 32],
 		};
-		connector.set_remote(RemoteInfo::new(&hello));
+		connector.set_remote(RemoteInfo::new(hello));
```
b-yap, line 343: Consume `hello`

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -110,16 +113,16 @@ async fn read_message_from_stellar(connector: &mut Connector) -> Result<Xdr, Err
...
 					&mut lack_bytes_from_prev,
 					&mut readbuf,
 					expect_msg_len,
 				)
 				.await
 				{
-					Ok(None) => continue,
-					Ok(Some(xdr)) => return Ok(xdr),
+					Ok(false) => continue,
+					Ok(true) => return Ok(readbuf),
```
b-yap, line 125: instead of returning a "cloned" `readbuf`, reuse it.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -180,7 +183,7 @@ async fn read_message(
 
 	// only when the message has the exact expected size bytes, should we send to user.
 	if actual_msg_len == xpect_msg_len {
-		return Ok(Some(readbuf.clone()))
```
b-yap, line 183: instead of sending back a cloned value, reuse the `readbuf`.

clients/stellar-relay-lib/src/node/mod.rs
```diff
@@ -19,6 +19,19 @@ pub struct NodeInfo {
 	pub network_id: NetworkId,
 }
 
+impl NodeInfo {
+	pub(crate) fn new(cfg:&NodeInfoCfg) -> Self {
```
b-yap, line 23: Eliminate an all out clone of `NodeInfoCfg`. We only need to clone 1 field, the `version_str`

clients/stellar-relay-lib/src/node/remote.rs
```diff
@@ -36,11 +36,11 @@ impl Debug for RemoteInfo {
 }
 
 impl RemoteInfo {
-	pub fn new(hello: &Hello) -> Self {
+	pub fn new(hello: Hello) -> Self {
```
b-yap, line 39: reduce clone by consuming `Hello`

clients/stellar-relay-lib/src/overlay.rs
```diff
@@ -65,8 +73,7 @@ impl StellarOverlayConnection {
 			return Err(Error::Disconnected)
 		}
 
-		timeout(Duration::from_secs(1), self.receiver.recv()).await
-			.map_err(|_| Error::Timeout)
+		Ok(self.receiver.recv().await)
```
b-yap, line 76: Remove extra timeout. Timeouts can be triggered from inside the polling.

clients/vault/src/lib.rs
```diff
@@ -59,3 +59,17 @@ cfg_if::cfg_if! {
...
 	}
 }
+
+pub fn tokio_spawn<F>(task_name: &str, future: F) -> tokio::task::JoinHandle<F::Output>
+	where
+		F: std::future::Future + Send + 'static,
+		F::Output: Send + 'static,
+{
+	cfg_if::cfg_if!{
+		if #[cfg(all(tokio_unstable, feature = "allow-debugger"))] {
```
b-yap, line 69: allow naming tasks, if `allow-debugger` feature is on.

clients/vault/src/oracle/agent.rs
```diff
@@ -15,16 +18,40 @@ use crate::oracle::{
...
 pub struct OracleAgent {
-	collector: Arc<RwLock<ScpMessageCollector>>,
+	pub collector: Arc<RwLock<ScpMessageCollector>>,
 	pub is_public_network: bool,
 	/// sends message directly to Stellar Node
 	message_sender: Option<StellarMessageSender>,
 	/// sends an entire Vault shutdown
-	shutdown_sender: ShutdownSender,
-	/// sends a 'stop' signal to `StellarOverlayConnection` poll
-	overlay_conn_end_signal: mpsc::Sender<()>,
```
b-yap, line 27: We do not need this signal anymore; OverlayConnection already implements `Drop` trait.

clients/vault/src/oracle/agent.rs
```diff
@@ -57,82 +84,76 @@ async fn handle_message(
 	Ok(())
 }
 
-/// Start the connection to the Stellar Node.
-/// Returns an `OracleAgent` that will handle incoming messages from Stellar Node,
-/// and to send messages to Stellar Node
-pub async fn start_oracle_agent(
```
b-yap, line 63: Renamed to `listen_for_stellar_messages(...)` to return nothing. This is an added task to be monitored in the tokio_metrics

clients/vault/src/oracle/agent.rs
```diff
@@ -57,82 +84,76 @@ async fn handle_message(
...
-						tracing::error!("start_oracle_agent(): encounter error in overlay: {e:?}");
-
-						if let Err(e) = shutdown_sender_clone.send(()) {
-							tracing::error!("start_oracle_agent(): Failed to send shutdown signal in thread: {e:?}");
-						}
-						break
-					},
-				}},
+	// log a new message received, every 1 minute.
+	let interval = Duration::from_secs(60);
```
b-yap, line 105: This is intentional, to keep logging every minute to show whether the vault is stuck or not.

clients/vault/src/oracle/agent.rs
```diff
@@ -57,82 +84,76 @@ async fn handle_message(
...
-
-	let shutdown_sender_clone = shutdown_sender.clone();
-	// disconnect signal sender tells the StellarOverlayConnection to close its TcpStream to Stellar
-	// Node
-	let (disconnect_signal_sender, mut disconnect_signal_receiver) = mpsc::channel::<()>(2);
-
-	let sender_clone = overlay_conn.sender();
-	tokio::spawn(async move {
-		loop {
-			tokio::select! {
```
b-yap, line 90: Removed tokio::select, since this task will close based on the `run_and_monitor_task()`.

clients/vault/src/oracle/agent.rs
```diff
@@ -177,25 +198,6 @@ impl OracleAgent {
...
-	pub async fn last_slot_index(&self) -> Slot {
-		self.collector.read().await.last_slot_index()
-	}
-
-	pub async fn remove_data(&self, slot: &Slot) {
-		self.collector.read().await.remove_data(slot);
-	}
-
-	/// Stops listening for new SCP messages.
-	pub async fn shutdown(&self) {
```
b-yap, line 190: This is unnecessary already; the `Drop` trait will handle dropping connections

clients/vault/src/lib.rs
ebma: Do we need to add this check? Or can we just always use the builder pattern and in case the `allow-debugger` feature is not enabled, the name just wouldn't show?

clients/vault/src/system.rs
```diff
@@ -609,6 +608,7 @@ impl VaultService {
 			(
 				"Restart Timer",
 				run(async move {
+					tracing::info!("Periodic restart in 3 hours.");
```
ebma, line 611: Let's insert the actual value of `RESTART_INTERVAL` in the log, otherwise this leads to confusion once we change the constant.

clients/vault/src/system.rs
```diff
@@ -26,26 +27,15 @@ use service::{wait_or_shutdown, Error as ServiceError, MonitoringConfig, Service
...
+use crate::{cancellation::ReplaceCanceller, error::Error, issue, issue::IssueFilter, metrics::{monitor_bridge_metrics, poll_metrics, publish_tokio_metrics, PerCurrencyMetrics}, oracle::OracleAgent, redeem::listen_for_redeem_requests, replace::{listen_for_accept_replace, listen_for_execute_replace, listen_for_replace_requests}, requests::execution::execute_open_requests, service::{CancellationScheduler, IssueCanceller}, ArcRwLock, Event, CHAIN_HEIGHT_POLLING_INTERVAL, tokio_spawn};
+use crate::oracle::listen_for_stellar_messages;
 
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
 pub const AUTHORS: &str = env!("CARGO_PKG_AUTHORS");
 pub const NAME: &str = env!("CARGO_PKG_NAME");
 pub const ABOUT: &str = env!("CARGO_PKG_DESCRIPTION");
 
-const RESTART_INTERVAL: Duration = Duration::from_secs(10800); // restart every 3 hours
+const RESTART_INTERVAL: Duration = Duration::from_secs(7200); // restart every 3 hours
```
ebma, line 38: This is now only 2 hours so let's adjust the comment.

clients/vault/src/lib.rs
b-yap: the [builder only works for tokio unstable](https://docs.rs/tokio/latest/tokio/task/join_set/struct.Builder.html), unfortunately.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -43,24 +43,33 @@ pub(crate) async fn poll_messages_from_stellar(
...
+				if let Err(e) = send_to_user_sender.send(stellar_msg).await {
 					warn!("poll_messages_from_stellar(): Error occurred during sending message {} to user: {e:?}",
-						String::from_utf8(stellar_msg.to_base64_xdr())
-						.unwrap_or_else(|_| format!("{:?}", stellar_msg.to_base64_xdr()))
-					);
+					String::from_utf8(stellar_msg_as_base64_xdr.clone())
+					.unwrap_or_else(|_| format!("{stellar_msg_as_base64_xdr:?}"))
+				);
 				}
+				tokio::task::yield_now().await;
```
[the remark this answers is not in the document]
b-yap: For the `break`s, I wanted a message to be sent back to the user that the stellar-relay will stop functioning.
Breaking out of the loop will stop the thread entirely. 

I chose to yield _only_ when `message_reader()` is doing fine. This will hit _pause_ so that other tasks will run. Then if its turn has come, the loop will go on.

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -173,44 +180,24 @@ impl ScpMessageCollector {
...
-
 	/// Insert envelopes fetched from the archive to the map
 	///
 	/// # Arguments
 	///
 	/// * `envelopes_map_lock` - the map to insert the envelopes to.
 	/// * `slot` - the slot where the envelopes belong to
 	fn get_envelopes_from_horizon_archive(&self, slot: Slot) -> impl Future<Output = ()> {
-		tracing::debug!("get_envelopes_from_horizon_archive(): Fetching SCP envelopes from horizon archive for slot {slot}...");
+		tracing::info!("get_envelopes_from_horizon_archive(): Fetching SCP envelopes from horizon archive for slot {slot}...");
```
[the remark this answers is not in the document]
b-yap: I printed the archive url  (see Line 215)

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -243,61 +237,66 @@ impl ScpMessageCollector {
...
 						);
-							// remove the file since it's invalid.
-							scp_archive_storage.remove_file(slot);
-							continue;
-						}
-
-						let mut envelopes_map = envelopes_map_arc.write();
-						let mut from_archive_map = env_from_archive_map.write();
-
-						if envelopes_map.get(&slot).is_none() {
```
b-yap, line 283: This caused the [timeout issue](https://github.com/pendulum-chain/spacewalk/actions/runs/11176052005/job/31083875194#step:4:5645).

The map had 1 envelope  for the given slot. Since it did not return `none`, this insert logic was always skipped.

I appended the existing envelope wit the envelopes found in the archive. See lines 287-292.

ebma: Wow, very nice catch!

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -243,61 +237,66 @@ impl ScpMessageCollector {
...
+							relevant_envelopes.len(),
+							externalized_envelopes_count
+						);
+				let mut envs = envelopes_map.get(&slot).map(|envs| envs.clone()).unwrap_or(vec![]);
+
+				if envs.len() > 0 {
+					tracing::info!("get_envelopes_from_horizon_archive(): {} envelopes already exist for slot {slot}", envs.len());
+				}
+
+				relevant_envelopes.append(&mut envs);
```
ebma, line 291: The stellar relay pallet automatically filters duplicate envelopes right? So it should be fine if we add duplicates here.
However, if we send duplicates to the oracle, this also increases the size of the already large extrinsic call. So how about we still try to filter and duplicate the envelopes here?

b-yap: Ooor, can we just remove pre-existing envelopes? Literally just replace it?

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -43,39 +46,58 @@ pub(crate) async fn poll_messages_from_stellar(
 		}
 
 		// check for messages from Stellar Node.
-		let xdr = match read_message_from_stellar(&mut connector).await {
-			Err(e) => {
+		// if reading took too much time, flag it as "disconnected"
+		let xdr = match timeout(
+			Duration::from_secs(READ_TIMEOUT_IN_SECS),
+			read_message_from_stellar(&mut connector),
```
b-yap, line 52: Instead of having the timeout inside `read_message_from_stellar(...)`, I have it handled outside of it. This here is more understandable, than having in the `message_reader.rs`

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
```diff
@@ -43,39 +46,58 @@ pub(crate) async fn poll_messages_from_stellar(
...
+			Ok(None) => tokio::task::yield_now().await,
 			Err(e) => {
 				error!("poll_messages_from_stellar(): Error occurred during processing xdr message: {e:?}");
 				break;
 			},
 		}
 	}
 
+	// push error to user
+	if let Err(e) = send_to_user_sender.send(crate_specific_error()).await {
```
b-yap, line 88: Something happened inside here, and it must be communicated back to the caller/user.

clients/stellar-relay-lib/src/node/mod.rs
```diff
@@ -19,6 +19,19 @@ pub struct NodeInfo {
 	pub network_id: NetworkId,
 }
 
+impl NodeInfo {
+	pub(crate) fn new(cfg: &NodeInfoCfg) -> Self {
```
b-yap, line 23: Let us create a new one and clone only the `version_str`, to avoid cloning the whole struct.

clients/stellar-relay-lib/src/overlay.rs
```diff
@@ -44,6 +40,17 @@ impl StellarOverlayConnection {
 
 		let connector = Connector::start(local_node_info, conn_info).await?;
 
+		#[cfg(tokio_unstable)]
+		tokio::task::Builder::new()
```
b-yap, line 44: the `task::builder` is only available in `tokio_unstable`.

clients/stellar-relay-lib/src/tests/mod.rs
```diff
@@ -43,31 +40,45 @@ async fn stellar_overlay_should_receive_scp_messages() {
...
 	let scps_vec_clone = scps_vec.clone();
-
-	timeout(Duration::from_secs(300), async move {
+	tokio::spawn(async move {
 		let mut ov_conn_locked = ov_conn.lock().await;
-		if let Ok(Some(msg)) = ov_conn_locked.listen() {
-			scps_vec_clone.lock().await.push(msg);
-
-			ov_conn_locked.stop();
+		loop {
```
b-yap, line 56: We have to loop here because `fn listen()` does not have a loop inside it, anymore.

clients/vault/src/oracle/agent.rs
```diff
@@ -57,84 +87,104 @@ async fn handle_message(
...
+		.await.map_err(|e|{
+			tracing::error!("listen_for_stellar_messages(): Failed to connect to Stellar overlay network: {e:?}");
+			service::Error::StartOracleAgentError
+		})?;
 
-	let mut overlay_conn = connect_to_stellar_overlay_network(config.clone(), secret_key).await?;
 	// use StellarOverlayConnection's sender to send message to Stellar
 	let sender = overlay_conn.sender();
+	{
+		oracle_agent.write().await.message_sender = Some(sender.clone());
```
b-yap, line 109: The whole reason why `Arc<OracleAgent>` has become `Arc<RwLock<OracleAgent>>`. We save the `message_sender` once `overlay_conn` has been initialized.

clients/vault/src/oracle/agent.rs
```diff
@@ -57,84 +87,104 @@ async fn handle_message(
...
+	loop {
+		let collector = oracle_agent.read().await.collector.clone();
+
+		match overlay_conn.listen().await {
+			Ok(None) => {},
+			Ok(Some(StellarMessage::Hello(_))) => {
+				tracing::info!(
+					"listen_for_stellar_messages(): received hello message from Stellar"
+				);
+				is_proof_building_ready = Some(true);
```
b-yap, line 125: If _hello_ message is received, it signals the agent that the overlay is running. 
However, it might not be ready for proof_building just yet, hence why it is a `Some()` (or why I chose `Option`  as a representation of this).

clients/vault/src/oracle/agent.rs
```diff
@@ -57,84 +87,104 @@ async fn handle_message(
...
-
-					if let Err(e) = shutdown_sender_clone.send(()) {
-						tracing::error!(
-							"start_oracle_agent(): Failed to send shutdown signal in thread: {e:?}"
-						);
-					}
-					break;
-				},
-			}
+				if let Some(true) = is_proof_building_ready {
```
b-yap, line 140: Keep logging, until the `proof_building_ready` variable is rendered useless.

clients/vault/src/oracle/collector/proof_builder.rs
ebma: Right, that should of course also work! Let's do that then

clients/vault/src/oracle/agent.rs
```diff
@@ -57,84 +87,104 @@ async fn handle_message(
...
-		shutdown_sender,
-		overlay_conn_end_signal: disconnect_signal_sender,
-	})
+	tracing::info!("listen_for_stellar_messages(): shutting down overlay connection");
+	overlay_conn.stop();
+
+	Ok(())
+}
+#[cfg(any(test, feature = "integration"))]
+pub async fn start_oracle_agent(
```
b-yap, line 171: This is only for testing purposes

clients/vault/src/oracle/agent.rs
```diff
@@ -33,14 +52,25 @@ pub struct OracleAgent {
...
-			collector.write().await.handle_envelope(env, message_sender).await?;
+			// if the first slot was saved, it means proof building is ready.
+			if let Some(slot) = collector.write().await.handle_envelope(env, message_sender).await?
+			{
+				if let Some(true) = is_proof_building_ready {
+					tracing::info!(
+						"handle_message(): First slot saved: {slot}. Ready to build proofs "
+					);
+				}
+				*is_proof_building_ready = None;
```
b-yap, line 72: proof building is ready; we do not need this flag/signal anymore.

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -16,16 +16,6 @@ use crate::oracle::{
 	ScpArchiveStorage, ScpMessageCollector, TransactionsArchiveStorage,
 };
 
-/// Returns true if the SCP messages for a given slot are still recoverable from the overlay
-/// because the slot is not too far back.
-fn check_slot_still_recoverable_from_overlay(last_slot_index: Slot, slot: Slot) -> bool {
```
b-yap, line 21: Moved this at the bottom of the file

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -173,44 +187,24 @@ impl ScpMessageCollector {
...
 		}
 	}
 
-	/// Returns the Proof
-	///
-	/// # Arguments
-	///
-	/// * `slot` - the slot where the txset is  to get.
-	/// * `sender` - used to send messages to Stellar Node
-	pub async fn build_proof(&self, slot: Slot, sender: &StellarMessageSender) -> Option<Proof> {
```
b-yap, line 192: moved this at the top of the file

.github/actions/prerequisite-nightly/action.yml
```diff
@@ -18,6 +18,8 @@ runs:
         token: ${{ inputs.token }}
 
     - name: Install Rust Nightly
-      uses: ./.github/actions/install-rust-nightly
```
b-yap, line 21: This github actions is using a deprecated `set-output`:
```
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.
```

Replaced with dtolnay

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
gianfra-t: This is very nice improvement because we were reading from the stream [here](https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/stellar-relay-lib/src/connection/connector/message_reader.rs#L178) without a timeout.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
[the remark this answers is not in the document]
gianfra-t: Makes sense, you want to make sure the `StellarOverlayConnection` process the message before continuing. This is quite complex to model in my mind since if we don't yield, then we probably continue with the next iteration, and at some point something when we hit `.await()` it will yield. But then perhaps `send_to_user_sender.is_closed()` is `true` and we break without the chance for the receiver to do something with the message.

So it "may" be redundant, but it doesn't hurt at all. Did you notice that without this yield something was failing?

clients/stellar-relay-lib/src/tests/mod.rs
gianfra-t: We could also `.await().expect()` the spawn handle, maybe then we don't need the other `timeout` thread. I read that maybe a panic inside a thread may not make the test fail.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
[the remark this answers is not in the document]
ebma: I agree it makes sense, thanks for explaining @b-yap.

ebma (commented):
This is a great improvement, thanks a lot @b-yap! We need to adjust some of the log levels because some messages are a bit spammy for 'regular' vault operators. 

## Some experiments

I ran the amplitude USDC #1 vault locally to check the resource consumption with tokio-
console. 

tokio-console after ~1.5 h (run #1)
image

CPU usage after ~1.5h:
image

--- 

After a while it seemed like tokio-console was not updating properly anymore so I was curious if tokio-console stopped working andI tried restarting it but it is not able to connect again. I tested if the vault is still operational by creating fresh issue and redeem requests against it and indeed, it was able to pick them up and complete them as expected. So maybe it's just some bug with tokio-console that makes it unable to connect after a while.

Unable to connect:
image

Proof that it's still able to process issue and redeem requests even though tokio-console shows blank:
image

---

To check the CPU usage, I then ran the vault a second time


CPU usage after 5 min:
image

We can see that in the beginning (before having received/processed any issue/redeem request) the % CPU is at ~2%- 3%.

I then ran the spacewalk-testing-service so created first 1 issue request and then 1 redeem request.

This increased the % CPU to ~5%
image

I kept monitoring this for about 10 minutes and from then on, the % CPU fluctuated between 4%-10%. It's hard to say this for sure but it might suggest that the resource consumption increases after an issue or redeem request is processed but stays more or less constant before. 

After 20 minutes it stayed almost always above 10%.. 
tokio-console after 20 minutes:
image

I'll try to experiment with this again tomorrow.

clients/vault/src/oracle/agent.rs
gianfra-t: So this is for the case when we receive a stellar message (`Some(msg)`) before to the `Hello` variant, then `handle_message` will not `trace`. Why do we need to receive first the `Hello` message to consider the proof building ready?

clients/vault/src/oracle/collector/proof_builder.rs
```diff
@@ -158,11 +171,12 @@ impl ScpMessageCollector {
...
 			Some(res) => Some(res),
 			None => {
+				tracing::info!("get_txset(): FOR SLOT {slot} check_slot_still_recoverable_from_overlay: LAST SLOT INDEX: {}",self.last_slot_index());
 				// If the current slot is still in the range of 'remembered' slots
 				if check_slot_still_recoverable_from_overlay(self.last_slot_index(), slot) {
-					self.fetch_missing_txset_from_overlay(slot, sender).await;
+					self.ask_overlay_for_txset(slot, sender).await;
 				} else {
-					tokio::spawn(self.get_txset_from_horizon_archive(slot));
+					self.get_txset_from_horizon_archive(slot).await;
```
gianfra-t, line 179: So we are now doing `await` on this vs spawning in a new thread as before. This means if I understand correctly that `build_proof` will now wait [here](https://github.com/pendulum-chain/spacewalk/pull/545/files#diff-732d3f237cd7cd9cf03fc64a9ae387ce0e80ad0928e995a3b9e61df4a4c24429R212) and the loop won't continue and call `build_proof` again. Why don't we need this behavior anymore now?

clients/vault/src/oracle/collector/proof_builder.rs
gianfra-t: That must have been hard to catch 😅. What was the reason we were checking for an empty map in the first place? And why we don't care now to append.

clients/vault/src/oracle/agent.rs
gianfra-t: So the reason we cannot call `connect_to_stellar_overlay_network` to get the `overlay_conn` over [here](https://github.com/pendulum-chain/spacewalk/blob/9cad8e4893182ab90cd2483d121df98f048eb402/clients/vault/src/system.rs#L815) and pass the sender there to the `oracle_agent` is because the overlay process will start to receive messages while the `oracle_agent` is not be ready ? Ready I mean not yet started the `read` loop in this function bellow.

clients/vault/src/oracle/collector/proof_builder.rs
b-yap: Before tokio-console: when the vaults hang, the last logs were in `build_proof` section getting data from the archive. 

I wanted to eliminate unnecessary tasks as possible, especially since these are not monitored at all. Should the original `tokio::spawn(self.get_txset_from_horizon_archive(slot));` become a zombie, there is nothing to stop that task. 

**The time spent looping `build_proof` continuously while the map is empty** 
_VS_
**The time spent waiting for the archive to fill the map** 

is just the same.

clients/service/src/lib.rs
b-yap: I forgot to reply to this.. 
I did not use a timeout, but the `try_recv()`; if it's not ready, then the loop will continue. 
The loop ends until it is  a success OR an error happens.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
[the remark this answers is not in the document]
b-yap: >  Did you notice that without this yield something was failing?

@gianfra-t So I yielded because in the tokio-console, this thread almost always was **BUSY**, giving no turn for other tasks to run. It was always trying to read messages from Stellar Relay. A yield for a few seconds wouldn't hurt.

> So it "may" be redundant,

True, it could be. But it is a loop, no one is awaiting `fn poll_messages_from_stellar()`.

clients/vault/src/oracle/collector/proof_builder.rs
b-yap: > What was the reason we were checking for an empty map in the first place? 

@gianfra-t  good qustion, and I don't have an answer. :see_no_evil:


> And why we don't care now to append.

We don't append anymore because of possible dupes. See [Marcel's comment](https://github.com/pendulum-chain/spacewalk/pull/545#discussion_r1791591783):
   _...
   if we send duplicates to the oracle, this also increases the size of the already large extrinsic call. 
   ..._

clients/vault/src/oracle/agent.rs
b-yap: Yes.

`OracleAgent` is our middleman to use the `stellar-relay-lib`'s connection.  Right now the `overlay_conn` lives in the logic of `OracleAgent`. I do not want to expose `overlay_conn` anywhere else, aside from the `oracle` directory. Separation of concerns.


But I do see your point. If we declare the `overlay_conn` separately, then we can keep `Arc<OracleAgent>`.
I am going to explore your idea; see how cleanly I can do this. 

Thanks Gianni!

clients/vault/src/oracle/agent.rs
b-yap: This was an attempt to cleanly understand the logs. 

Looking at your pov, the hello does look like an overkill, for proof building ready. 
I will edit this portion and see what I can do. 

Thank you Gianni!

ebma: I'm not sure I understand why we need this variable? Also, with Option<bool> it can now be three things, `None`, `Some(true)` and `Some(false)`. As far as I can tell, we only check for `Some(true)` and never for `None` so it should be fine if we don't make it an `Option` type but a simple boolean as we apparently only need the two cases.

clients/vault/src/requests/execution.rs
```diff
@@ -288,9 +288,11 @@ pub async fn execute_open_requests(
 	parachain_rpc: SpacewalkParachain,
 	vault_id_manager: VaultIdManager,
 	wallet: Arc<RwLock<StellarWallet>>,
-	oracle_agent: Arc<OracleAgent>,
+	oracle_agent: ArcRwLock<OracleAgent>,
 	payment_margin: Duration,
+	precheck_signal: tokio::sync::broadcast::Sender<()>,
```
ebma, line 293: Nit: this is missing in the doc comment above.

clients/vault/src/issue.rs
```diff
@@ -26,6 +26,7 @@ pub(crate) async fn initialize_issue_set(
 	issue_set: &ArcRwLock<IssueRequestsMap>,
 	memos_to_issue_ids: &ArcRwLock<IssueIdLookup>,
 ) -> Result<(), Error> {
+	tracing::info!("initialize_issue_set(): started");
```
ebma, line 29: Nit: make this and the other similar logs `debug`

clients/vault/src/replace.rs
```diff
@@ -96,6 +96,8 @@ pub async fn listen_for_replace_requests(
 	event_channel: Sender<Event>,
 	accept_replace_requests: bool,
 ) -> Result<(), ServiceError<Error>> {
+	tracing::info!("listen_for_replace_requests(): started");
```
ebma, line 99:
```suggestion
	tracing::debug!("listen_for_replace_requests(): started");
```

clients/vault/tests/vault_integration_tests.rs
```diff
@@ -1163,17 +1158,23 @@ async fn test_execute_open_requests_succeeds() {
 			// add it to the set
 			sleep(Duration::from_secs(5)).await;
 
+			let (precheck_signal, mut rceiver) = tokio::sync::broadcast::channel(1);
```
ebma, line 1161:
```suggestion
			let (precheck_signal, mut receiver) = tokio::sync::broadcast::channel(1);
```

clients/service/src/lib.rs
```diff
@@ -170,17 +173,38 @@ impl<Config: Clone + Send + 'static, F: Fn()> ConnectionManager<Config, F> {
...
 where
 	F: Future<Output = Result<(), E>>,
 {
+	if let Some(mut precheck_signal) = precheck_signal {
+		loop {
+			match precheck_signal.try_recv() {
+				// Received a signal to start the task
+				Ok(_) => break,
+				Err(TryRecvError::Empty) =>
+					tracing::trace!("wait_or_shutdown precheck signal: waiting..."),
```
ebma, line 191: I think this is problematic because we are again creating tasks that never yield, potentially causing issues for the scheduling of other tasks. 
```suggestion
				Err(TryRecvError::Empty) => {
					tracing::trace!("wait_or_shutdown precheck signal: waiting..."),
					sleep(Duration::from_millis(100)).await; // Yield to the runtime
				}
```

ebma (commented):
Sorry for leaving more comments yet again. I pointed out more logs that I think we should turn to the debug log level instead. 

Besides that I encountered an error when running the vault locally. It happened on start
```
thread 'tokio-runtime-worker' panicked at clients/runtime/src/rpc.rs:296:49:
called `Result::unwrap()` on an `Err` value: Rpc(ClientError(MaxSlotsExceeded))
```
It's unrelated to the changes of this PR but @b-yap if it's not too complex, maybe you could try to replace all the `unwrap()` calls in `clients/runtime/src/rpc.rs` as part of this as well? 😬 

---

I also want to share good news though. Further experiments I did locally showed that the CPU usage only piles up when running the vault with `--features allow-debugger` ie. support for debugging with tokio-console. Without that flag, the client's resource consumption remains quite low at ~1-4% when idling. Seems like the refactoring indeed fixed our bugs with tokio and we can be optimistic that these changes greatly improve the performance once live on production (without the debugging enabled).

clients/service/src/lib.rs
gianfra-t: Or we can use `precheck_signal.recv().await`? this one is [async](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html#method.recv).

clients/vault/src/oracle/collector/proof_builder.rs
gianfra-t: I see! makes sense, in fact makes much more sense since `get_txset_from_horizon_archive` will return when the `tx_set` was found. But before we were potentially actually spawning many instances of `get_txset_from_horizon_archive` for the same slot, I thought maybe we needed several of these tasks, or that we needed to continue and return a proof with `None` set.

clients/vault/src/oracle/agent.rs
gianfra-t: Okay! Yeah I also thought that a boolean works as well since we use two states but if it's for logging only, of course leave it as long as you need.

clients/vault/src/oracle/agent.rs
gianfra-t: Okay I see clearly now. No need to change it!

You need to create the oracle agent on the high level [execution](https://github.com/pendulum-chain/spacewalk/blob/9cad8e4893182ab90cd2483d121df98f048eb402/clients/vault/src/system.rs#L782) to pass it to all the other tasks. 
Before, the `overlay_conn` task was created along with the `oracle_agent` [here](https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/vault/src/oracle/agent.rs#L72) and now that second process is spawned inside `listen_for_stellar_messages`. 

I think the separation of concerns is very valid, I just wanted to understand. And I don't see any pro-con regarding where we spawn the processes so far, since we are not handling them (poll_messages_from_stellar) anyway and live independently.

clients/stellar-relay-lib/src/connection/connector/message_reader.rs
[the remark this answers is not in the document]
gianfra-t: > no one is awaiting fn poll_messages_from_stellar()

True! I meant that inside the loop itself, on the other operations like `read_message_from_stellar` we await so unless that one is also never yielding, the overall process will (I think?). But it is more explicit this way that we want to yield at this point so it's fine by me 👍.

clients/vault/src/oracle/collector/proof_builder.rs
gianfra-t: Great 👍. I miss that other comment. Thanks!

gianfra-t (approved):
Super nice (and involved😅) refactor 💪! And I'm glad it solved the hanging and apparently also the high CPU usage issue. Thanks @b-yap for going back to my questions and comments.

clients/vault/src/oracle/agent.rs
b-yap: `is_proof_building_ready` completed **_removed_**.

clients/service/src/lib.rs
gianfra-t: I think if the receiver `precheck_signal` has already consumed the message then next time calling `recv().await`, it will yield until a new one arrives. Or do you think there will always be a message in the broadcast channel? 

This implementation is equivalent though. But with `try_recv` we do need to yield that's true.

clients/runtime/src/rpc.rs
```diff
@@ -313,13 +310,19 @@ impl SpacewalkParachain {
...
 	}
 
 	pub async fn get_finalized_block_hash(&self) -> Result<Option<H256>, Error> {
 		Ok(Some(self.api.backend().latest_finalized_block_ref().await?.hash()))
 	}
 
+	async fn get_latest_storage(
+		&self,
+	) -> Storage<SpacewalkRuntime, OnlineClient<SpacewalkRuntime>> {
+		self.api.storage().at_latest().await.expect("failed to get latest storage")
```
ebma, line 323: Can we map this to an error instead of `expect`ing or would that require a lot of refactoring?

ebma: I think we unfortunately cannot expect this to never fail, so it would be very good if we can handle this error gracefully.

ebma (approved):
Great that you found a way to get rid of the `ArcRwLOck` on the oracle agent, improves the readability a lot 🙏

clients/service/src/lib.rs
b-yap: >  do you think there will always be a message in the broadcast channel?

It will only send 1 `()` message. See lines 335-337 in clients/vault/src/requests/execution.rs.

[The first try was to use `recv().await`](https://github.com/pendulum-chain/spacewalk/pull/545#discussion_r1714911315) but the possibility of it waiting forever is there. 

However.. we also cannot move forward with the task anyway, if the precheck (or the `execute_open_requests`) is not finished. 
I also would not like to wrap the `recv()` in a `timeout()` because open requests might really take longer than the timeout we set. Also, a `timeout` is another thread; That's 8 tasks requiring prechecks, meaning 8 additional threads. (See [Marcel's comment](https://github.com/pendulum-chain/spacewalk/pull/545#pullrequestreview-2360773756) for the image of these tasks: Issue Cancel Listener, Issue Request Listener, etc.)

b-yap: Readability, reverting to `recv().await` would be better. I would add that back, but we have to risk it awaiting. 
I think `execute_open_requests()` is reliable enough to be able to send the message fast.

clients/stellar-relay-lib/src/tests/mod.rs
b-yap: Updated it. :+1:

clients/runtime/src/rpc.rs
b-yap: wrapped it inside `Error::SubxtRuntimeError()`

b-yap:
@ebma @gianfra-t CI passed. Merging this.

Issue #549, opened 2024-08-13
 Stellar overlay: Mandatory Flow control in bytes


pieterjan84:
FYI: https://github.com/stellar/stellar-core/pull/4353

Flow control in bytes is now required. Also listed as breaking change in latest pre-release of stellar-core. 

Implementation made for Stellarbeat: https://github.com/stellarbeat/js-stellar-node-connector/blob/master/src/connection/flow-controller.ts

If I'm mistaking that this is an issue for your project, please ignore :-)

prayagd:
Thank you @pieterjan84 for raising the issue.

cc @pendulum-chain/devs is this relevant for us?

b-yap:
Yes it is an issue:
```
Error occurred during processing xdr message: ConfigError("Error{ code:ErrConf message:flow control bytes disabled }")
```

Thank you once again @pieterjan84 !

ebma:
This was fixed in #550 but we forgot to close it.

Issue #520, opened 2024-05-07
Test Case `stellar_overlay_should_receive_scp_messages` fails 


b-yap:
This test case:
https://github.com/pendulum-chain/spacewalk/blob/ea66a3062e6edf05a4ed7b377f107e1f95b688f1/clients/stellar-relay-lib/src/tests/mod.rs#L42-L67

failed twice:
```
failures:

---- tests::stellar_overlay_should_receive_scp_messages stdout ----
thread 'tests::stellar_overlay_should_receive_scp_messages' panicked at clients/stellar-relay-lib/src/tests/mod.rs:66:5:
assertion failed: !scps_vec.lock().await.is_empty()
```
See
https://github.com/pendulum-chain/spacewalk/actions/runs/8980428758/job/24674454969?pr=513#step:12:510
https://github.com/pendulum-chain/spacewalk/actions/runs/8980428758/job/24672556732?pr=513#step:12:509

Investigate whether the test case only needs an update, or [an update of the config file](https://github.com/pendulum-chain/spacewalk/tree/main/clients/stellar-relay-lib/resources/config/testnet) is required.

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/520?planning-poker) with Zenhub @b-yap @bogdanS98 @ebma @gianfra-t @TorstenStueber

gianfra-t:
I see that this test [works](https://github.com/pendulum-chain/spacewalk/actions/runs/10287207285/job/28523428666?pr=546#step:13:507) at least in the latest runs. Maybe it was solved by changing the config file?

bogdanS98:
@gianfra-t There was at least one config file change in the last 2 months (e.g. [this](https://github.com/pendulum-chain/spacewalk/commit/524b15e10866dee855a9b2c4d0a930702124b59b)) and this ticket's description was last updated on May 7th which is way before that. I also think this is solved already by the new config file changes.

ebma:
Would be great if this got fixed automatically. @gianfra-t did you notice the test case failing in any of your comments of the https://github.com/pendulum-chain/spacewalk/pull/550 PR? If that's not the case I would be confident enough to close this ticket.

gianfra-t:
[...]

PR #550, opened 2024-08-23
Solve mandatory flow control

gianfra-t:
Issue: #549.

Modifying `FlowControl` and `sendMore` methods to be compliant with changes specified in [here](https://github.com/stellarbeat/js-stellar-node-connector/commit/51e2b308ac7520973eccbc748509a01ccb0e4519).

### Changes

Modify the [flow controller](https://github.com/pendulum-chain/spacewalk/pull/550/files#diff-9efd3eebef967795a862697c1d3cdb59b88a7ac8435c286f4750735823edecc1) logic as in the latest [implementation](https://github.com/stellarbeat/js-stellar-node-connector/blob/master/src/connection/flow-controller.ts) of the `js-stellar-node` package.

Notably, the modification of the `sendMore` method.

- `FlowControl` in bytes is [enabled](https://github.com/pendulum-chain/spacewalk/pull/550/files#diff-9efd3eebef967795a862697c1d3cdb59b88a7ac8435c286f4750735823edecc1R22) (or not) on [start](https://github.com/pendulum-chain/spacewalk/pull/550/files#diff-a415e544a5cbc568c7c5b061585fcbe1650f2edc4dabaf6548e3e7eeb67cb13aR116) analogous to it's javascript [counterpart](https://github.com/stellarbeat/js-stellar-node-connector/blob/master/src/connection/connection.ts#L583).
- Adjust constants based on [this](https://github.com/stellarbeat/js-stellar-node-connector/commit/51e2b308ac7520973eccbc748509a01ccb0e4519#diff-6925ed62473229a20eaac1f707a831b3848ed70edc00284981868de19d64893dR25).

### DIFF
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -7,7 +7,7 @@ use substrate_stellar_sdk::{
 	types::{AuthenticatedMessageV0, Curve25519Public, HmacSha256Mac, MessageType},
 	XdrCodec,
 };
-use tracing::{error, trace};
+use tracing::{debug, error, trace};
 
 use crate::{
 	connection::{
@@ -34,7 +34,7 @@ pub struct Connector {
 	receive_scp_messages: bool,
 
 	handshake_state: HandshakeState,
-	flow_controller: FlowController,
+	pub(crate) flow_controller: FlowController,
 
 	/// for writing/reading xdr messages to/from Stellar Node.
 	pub(crate) tcp_stream: TcpStream,
@@ -231,17 +231,13 @@ impl Connector {
 	pub fn handshake_completed(&mut self) {
 		self.handshake_state = HandshakeState::Completed;
 	}
-
-	pub fn inner_check_to_send_more(&mut self, msg_type: MessageType) -> bool {
-		self.flow_controller.send_more(msg_type)
-	}
-	pub fn enable_flow_controller(
-		&mut self,
-		local_overlay_version: u32,
-		remote_overlay_version: u32,
-	) {
-		self.flow_controller.enable(local_overlay_version, remote_overlay_version)
-	}
+	// pub fn enable_flow_controller(
+	// 	&mut self,
+	// 	local_overlay_version: u32,
+	// 	remote_overlay_version: u32,
+	// ) {
+	// 	self.flow_controller.enable(local_overlay_version, remote_overlay_version, )
+	// }
 }
 
 #[cfg(test)]
@@ -425,12 +421,12 @@ mod test {
 		assert!(connector.is_handshake_created());
 	}
 
-	#[tokio::test]
-	#[serial]
-	async fn enable_flow_controller_works() {
-		let (node_info, _, mut connector) = create_connector().await;
-
-		assert!(!connector.inner_check_to_send_more(MessageType::ScpMessage));
-		connector.enable_flow_controller(node_info.overlay_version, node_info.overlay_version);
-	}
+	// #[tokio::test]
+	// #[serial]
+	// async fn enable_flow_controller_works() {
+	// 	let (node_info, _, mut connector) = create_connector().await;
+	//
+	// 	assert!(!connector.inner_check_to_send_more(MessageType::ScpMessage));
+	// 	connector.enable_flow_controller(node_info.overlay_version, node_info.overlay_version);
+	// }
 }
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -236,7 +236,7 @@ impl Connector {
 	) -> StellarMessage {
 		return self
 			.flow_controller
-			.start_control(local_overlay_version, remote_overlay_version);
+			.start(local_overlay_version, remote_overlay_version);
 	}
 
 	pub fn handshake_completed(&mut self) {
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -229,15 +229,19 @@ impl Connector {
 	pub fn got_hello(&mut self) {
 		self.handshake_state = HandshakeState::GotHello;
 	}
-	pub fn maybe_start_flow_control_bytes(&mut self, local_overlay_version: u32, remote_overlay_version: u32) -> StellarMessage {
-		return self.flow_controller.start_control(local_overlay_version, remote_overlay_version)
+	pub fn maybe_start_flow_control_bytes(
+		&mut self,
+		local_overlay_version: u32,
+		remote_overlay_version: u32,
+	) -> StellarMessage {
+		return self
+			.flow_controller
+			.start_control(local_overlay_version, remote_overlay_version);
 	}
-
 
 	pub fn handshake_completed(&mut self) {
 		self.handshake_state = HandshakeState::Completed;
 	}
-
 }
 
 #[cfg(test)]
@@ -245,11 +249,6 @@ mod test {
 	use crate::{connection::hmac::HMacKeys, node::RemoteInfo, StellarOverlayConfig};
 	use serial_test::serial;
 
-	use substrate_stellar_sdk::{
-		compound_types::LimitedString,
-		types::Hello,
-		PublicKey,
-	};
 	use crate::{
 		connection::{
 			authentication::{create_auth_cert, ConnectionAuth},
@@ -259,6 +258,7 @@ mod test {
 		node::NodeInfo,
 		ConnectionInfo,
 	};
+	use substrate_stellar_sdk::{compound_types::LimitedString, types::Hello, PublicKey};
 
 	use wallet::keys::get_source_secret_key_from_env;
 
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -7,6 +7,8 @@ use substrate_stellar_sdk::{
 	types::{AuthenticatedMessageV0, Curve25519Public, HmacSha256Mac},
 	XdrCodec,
 };
+
+use substrate_stellar_sdk::types::StellarMessage;
 use tracing::{error, trace};
 
 use crate::{
@@ -227,17 +229,15 @@ impl Connector {
 	pub fn got_hello(&mut self) {
 		self.handshake_state = HandshakeState::GotHello;
 	}
+	pub fn maybe_start_flow_control_bytes(&mut self, local_overlay_version: u32, remote_overlay_version: u32) -> StellarMessage {
+		return self.flow_controller.start_control(local_overlay_version, remote_overlay_version)
+	}
+
 
 	pub fn handshake_completed(&mut self) {
 		self.handshake_state = HandshakeState::Completed;
 	}
-	// pub fn enable_flow_controller(
-	// 	&mut self,
-	// 	local_overlay_version: u32,
-	// 	remote_overlay_version: u32,
-	// ) {
-	// 	self.flow_controller.enable(local_overlay_version, remote_overlay_version, )
-	// }
+
 }
 
 #[cfg(test)]
@@ -250,7 +250,6 @@ mod test {
 		types::Hello,
 		PublicKey,
 	};
-
 	use crate::{
 		connection::{
 			authentication::{create_auth_cert, ConnectionAuth},
@@ -420,13 +419,4 @@ mod test {
 		connector.handshake_completed();
 		assert!(connector.is_handshake_created());
 	}
-
-	// #[tokio::test]
-	// #[serial]
-	// async fn enable_flow_controller_works() {
-	// 	let (node_info, _, mut connector) = create_connector().await;
-	//
-	// 	assert!(!connector.inner_check_to_send_more(MessageType::ScpMessage));
-	// 	connector.enable_flow_controller(node_info.overlay_version, node_info.overlay_version);
-	// }
 }
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -247,7 +247,7 @@ mod test {
 
 	use substrate_stellar_sdk::{
 		compound_types::LimitedString,
-		types::{Hello, MessageType},
+		types::Hello,
 		PublicKey,
 	};
 
--- a/clients/stellar-relay-lib/src/connection/connector/connector.rs
+++ b/clients/stellar-relay-lib/src/connection/connector/connector.rs
@@ -4,10 +4,10 @@ use std::{
 	net::Shutdown,
 };
 use substrate_stellar_sdk::{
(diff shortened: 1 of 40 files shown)


clients/stellar-relay-lib/src/connection/connector/message_handler.rs
```diff
@@ -24,7 +23,7 @@ impl Connector {
 		match msg_type {
 			MessageType::Transaction | MessageType::FloodAdvert if !self.receive_tx_messages() => {
 				self.increment_remote_sequence()?;
-				self.check_to_send_more(MessageType::Transaction).await?;
+				self.check_to_send_more(MessageType::Transaction, data.len()).await?;
```
gianfra-t, line 26: Unlike how the data length of the message is calculated [here](https://github.com/stellarbeat/js-stellar-node-connector/blob/master/src/connection/connection.ts#L316), I believe in our implementation we already have the "cleaned" `XDR` representation at this point, and does not need to be stripped like it is done [here](https://github.com/stellarbeat/js-stellar-node-connector/blob/master/src/connection/connection.ts#L380).

clients/stellar-relay-lib/src/connection/connector/connector.rs
```diff
@@ -424,13 +419,4 @@ mod test {
 		connector.handshake_completed();
 		assert!(connector.is_handshake_created());
 	}
-
-	#[tokio::test]
-	#[serial]
-	async fn enable_flow_controller_works() {
```
gianfra-t, line 430: Not quite sure how to re-introduce this test, or equivalent. Since the `enabled` state will be set when on response of the `Auth` message, so we would have to start the message polling process.

ebma:
Did you run cargo fmt with the latest nightly or which toolchain version? How come there are so many changes 🤔

gianfra-t:
Ohh, I may have run it from a later nightly version. In any case I had to because even after `cargo fmt` I got the [following](https://github.com/pendulum-chain/spacewalk/actions/runs/10558944696/job/29249380931#step:14:15) on the CI.
Let me try by rolling back and running it with our `+nightly-2024-02-09` to be sure.

gianfra-t:
@ebma still many changes, must be because we have not run `fmt` in a while. Especially since we ignore the CI on the update and we skipped this step. 
Please @pendulum-chain/devs compare from [this](https://github.com/pendulum-chain/spacewalk/compare/d74009bf1d8c27fbd1f314c4fe486cd230e8b73f..fbab2026d8796a27e80fb62c7c8e790a58022965) commit to avoid looking at the `fmt` changes, or let me know and we can rollback temporarily the format commit.

clients/stellar-relay-lib/src/connection/flow_controller.rs
```diff
@@ -1,33 +1,83 @@
...
+	pub(crate) flow_control_bytes_enabled: bool,
+	messages_received_in_current_batch: u32,
+	bytes_received_in_current_batch: u32,
 }
 
 impl FlowController {
-	pub fn enable(&mut self, local_overlay_version: u32, remote_overlay_version: u32) {
-		self.enabled = remote_overlay_version >= 20 && local_overlay_version >= 20;
+	pub fn enable_bytes(&mut self, local_overlay_version: u32, remote_overlay_version: u32) {
+		self.flow_control_bytes_enabled = remote_overlay_version >= 28 && local_overlay_version >= 28 && AUTH_FLAG == 200;
```
ebma: The last check `AUTH_FLAG == 200` is useless if we always define this as a constant. I think we should pass it into this function instead or remove the condition completely

clients/stellar-relay-lib/src/connection/handshake.rs
```diff
@@ -14,7 +17,8 @@ pub enum HandshakeState {
 }
 
 pub fn create_auth_message() -> StellarMessage {
-	let auth = Auth { flags: 100 };
+	// 200 for flowControlInBytes enabled.
+	let auth = Auth { flags: AUTH_FLAG };
```
ebma, line 21: Should we always enable it? I can see [here](https://github.com/stellarbeat/js-stellar-node-connector/blob/9120f24b75867700ba0ece76369166959205c6b9/src/connection/connection.ts#L634C9-L634C22),  [here](https://github.com/stellarbeat/js-stellar-node-connector/blob/9120f24b75867700ba0ece76369166959205c6b9/src/connection/handshake-message-creator.ts#L12) that it should be enabled if the overlay version of the node is `>= 28` so we could simplify it to the statement here, but maybe it's better if we instead use a condition as they do in the JS code. If that's too much refactoring, we can also adjust the comment to something like
```
// AUTH_FLAG is set to `200` to enable flow control in bytes when connecting to nodes with overlay version >= 28. 
// We assume that the overlay version is always >= 28
```

clients/stellar-relay-lib/src/connection/connector/connector.rs
ebma: I guess we don't need it anymore.

ebma (commented):
Nice changes @gianfra-t, thanks for diving into this 👌 I added some thoughts

clients/stellar-relay-lib/src/connection/connector/message_handler.rs
gianfra-t: Let's see. What I can tell is that I tried before to substract the "padding" (as is calculated in the js implementation) and we get a panic since the data length is smaller. So if the message itself is padded, it is not exactly by those amount of bytes.

clients/stellar-relay-lib/src/connection/flow_controller.rs
```diff
@@ -1,33 +1,83 @@
...
+		let msg = StellarMessage::SendMore(SendMore { num_messages: MAX_FLOOD_MSG_CAP});
+		return msg
+	}
+
+	fn reset_batch_counters(&mut self) {
+		self.messages_received_in_current_batch = 0;
+		self.bytes_received_in_current_batch = 0;
+	}
 
+	pub fn send_more(&mut self, message_type: MessageType, data_len: usize) -> Option<StellarMessage> {
```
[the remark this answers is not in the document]
gianfra-t: Sure, in fact it should be called `reclaim_capacity` if I correctly understand the process, but that's how it is called on the counterpart. What about `create_reclaim_capacity_message`?

clients/stellar-relay-lib/src/connection/flow_controller.rs
gianfra-t: True, the reason I left it is because we may want to change the flag back to `100`, which we can do by changing the constant, and will not enable this bytes control.
We can do the fully parametrized implementation like [here](https://github.com/stellarbeat/js-stellar-node-connector/blob/9120f24b75867700ba0ece76369166959205c6b9/src/connection/connection.ts#L634) and it will make sense to compare it against the flag.

clients/stellar-relay-lib/src/connection/handshake.rs
gianfra-t: Sure, I'll try since it doesn't look like too much effort.

clients/stellar-relay-lib/src/connection/flow_controller.rs
gianfra-t: I removed the check, since we enable anyway flag `200` when local overlay is >= 28, therefore it is redundant.

clients/stellar-relay-lib/src/connection/connector/message_handler.rs
gianfra-t: I think I was wrong, please check the last committed [change](https://github.com/pendulum-chain/spacewalk/pull/550/commits/7a4a105c8f3922984aca6a711de77dbfd49cdafc), now it should be good. 

Two main things that hinted this:
- We do the same [slicing](https://github.com/pendulum-chain/spacewalk/blob/d74009bf1d8c27fbd1f314c4fe486cd230e8b73f/clients/stellar-relay-lib/src/connection/connector/message_handler.rs#L51) to verify [authentication](https://github.com/stellarbeat/js-stellar-node-connector/blob/9120f24b75867700ba0ece76369166959205c6b9/src/connection/connection.ts#L372)
- We slice also before parsing the message ([ours](https://github.com/pendulum-chain/spacewalk/blob/main/clients/stellar-relay-lib/src/connection/xdr_converter.rs#L64), [theirs](https://github.com/stellarbeat/js-stellar-node-connector/blob/9120f24b75867700ba0ece76369166959205c6b9/src/connection/connection.ts#L380)).

The reason it was failing, is because the call to `check_send_more` [here](https://github.com/pendulum-chain/spacewalk/pull/550/commits/7a4a105c8f3922984aca6a711de77dbfd49cdafc#diff-a415e544a5cbc568c7c5b061585fcbe1650f2edc4dabaf6548e3e7eeb67cb13aL98) was incorrect. Also, I added these two [cases](https://github.com/pendulum-chain/spacewalk/pull/550/commits/7a4a105c8f3922984aca6a711de77dbfd49cdafc#diff-a415e544a5cbc568c7c5b061585fcbe1650f2edc4dabaf6548e3e7eeb67cb13aR96) on the match (as the [original](https://github.com/stellarbeat/js-stellar-node-connector/blob/9120f24b75867700ba0ece76369166959205c6b9/src/connection/connection.ts#L509)), even though we do nothing with them, it would not be correct to call the control on them.

clients/stellar-relay-lib/src/connection/flow_controller.rs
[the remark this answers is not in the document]
ebma: `create_reclaim_capacity_message()` is a great name 👍

clients/stellar-relay-lib/src/connection/connector/message_handler.rs
ebma: True, nice catches 👌 makes sense to me

ebma:
@gianfra-t did you try if running the [example](https://github.com/pendulum-chain/spacewalk/blob/flow-control-issue/clients/stellar-relay-lib/examples/connect.rs) still works?

gianfra-t:
Yes, still works. By work you mean an endless stream of the incoming messages right?

ebma:
Yes exactly, just running it for about 10 minutes and successfully receiving and decoding messages from the overlay network.

gianfra-t:
I think we still have some test issues, this case was `test_off_chain_liquidation`. Is this relevant for #520 ?

clients/stellar-relay-lib/src/connection/flow_controller.rs
```diff
@@ -1,42 +1,101 @@
...
+pub const MAX_FLOOD_MSG_CAP: u32 = 200;
+pub const FLOW_CONTROL_SEND_MORE_BATCH_SIZE: u32 = 40;
+pub const PER_FLOOD_READING_CAPACITY_BYTES: u32 = 300000;
+pub const FLOW_CONTROL_SEND_MORE_BATCH_SIZE_BYTES: u32 = 100000;
 
 #[derive(Debug, Default)]
 pub struct FlowController {
-	enabled: bool,
-	flood_msg_cap: u32,
+	pub(crate) flow_control_bytes_enabled: bool,
```
b-yap, line 10: Does this have to be accessible anywhere else, inside the crate?

clients/stellar-relay-lib/src/connection/connector/message_handler.rs
```diff
@@ -65,8 +65,9 @@ impl Connector {
 		&mut self,
 		msg: StellarMessage,
 		msg_type: MessageType,
+		data_len: usize,
 	) -> Result<Option<StellarMessage>, Error> {
-		match msg {
+		match msg.clone() {
```
b-yap, line 70: why clone?

clients/stellar-relay-lib/src/connection/flow_controller.rs
```diff
@@ -1,42 +1,101 @@
...
-		self.enabled = remote_overlay_version >= 20 && local_overlay_version >= 20;
+	pub fn enable_bytes(&mut self, local_overlay_version: u32, remote_overlay_version: u32) {
+		self.flow_control_bytes_enabled =
+			remote_overlay_version >= 28 && local_overlay_version >= 28;
 	}
 
-	pub fn send_more(&mut self, message_type: MessageType) -> bool {
-		if !self.enabled {
-			return false;
+	pub fn start_control(
```
b-yap, line 21: maybe `fn start()` is understandable itself?

clients/stellar-relay-lib/src/connection/flow_controller.rs
```diff
@@ -1,42 +1,101 @@
...
+		&mut self,
+		local_overlay_version: u32,
+		remote_overlay_version: u32,
+	) -> StellarMessage {
+		self.enable_bytes(local_overlay_version, remote_overlay_version);
+
+		if self.flow_control_bytes_enabled {
+			let msg = StellarMessage::SendMoreExtended(SendMoreExtended {
+				num_messages: MAX_FLOOD_MSG_CAP,
```
b-yap, line 30: Please rename `MAX_FLOOD_MSG_CAP` to `PER_FLOOD_READING_CAPACITY`.

clients/stellar-relay-lib/src/connection/handshake.rs
```diff
@@ -6,15 +6,22 @@ use substrate_stellar_sdk::{
 	PublicKey,
 };
 
+// We enable flowControlWithBytes conditionally. This should be enabled if the node version is
+// higher than 28. https://github.com/stellarbeat/js-stellar-node-connector/blob/9120f24b75867700ba0ece76369166959205c6b9/src/connection/connection.ts#L634
+// Will not be needed anymore after the minimum version is set to 28.
+pub const AUTH_FLAG_BYTES_CONTROL: i32 = 200;
```
b-yap, line 12: Maybe this doesn't have to be `pub`? 
Same with `AUTH_FLAG_CONTROL`

b-yap (changes requested):
Sorry I couldn't do this myself.

I added a few comments just for readability sake. But overall, very nice, @gianfra-t !!

b-yap:
I also reran the `test_off_chain_liquidation` locally by itself, and it ran fine. I'll trigger the re-run and let's see what happens.

clients/stellar-relay-lib/src/connection/flow_controller.rs
gianfra-t: Actually now that you say, it doesn't. I did it because I was [trying](https://github.com/pendulum-chain/spacewalk/pull/550#discussion_r1731079342) to re-add a test for it but I found some roadblocks for that.

clients/stellar-relay-lib/src/connection/connector/message_handler.rs
gianfra-t: Another remnant of previous changes. Removing. Good catch!

gianfra-t:
The tests seemed to be stuck back in in the build process, but the [latest successful](https://github.com/pendulum-chain/spacewalk/actions/runs/10579331573/job/29351126684) CI and @b-yap local testing seem to indicate that it is okay. Should we merge this?

ebma (approved):
I'm fine with merging.

b-yap (approved):
yep, I'm fine with merging it too

Issue #548, opened 2024-08-12
Make vault clients export metrics with default values if no data is available


ebma:
## Context

When a vault client is running for a liquidated vault, some data is not exported anymore. This makes it difficult for the monitoring services to work properly as 'no data' would generally be considered a bug or faulty. To facilitate the monitoring and alerting of our vault clients we should change this to rather return a default value of instead. 

## TODO 

We want to change the vault clients to keep exporting data for the following metrics, even if the vault is liquidated:

- Locked collateral
- Required collateral
- Issue count
- Stellar Balance
- Redeem count
- Collateralization

For the metrics that are not applicable for liquidated vaults, namely locked and required collateral as well as collateralization, we should return `0` instead. For the others, we can keep exporting the correct values.

ebma:
@pendulum-chain/product this ticket helps improve our automated alerts as it allows to monitor vaults even when they are liquidated.

prayagd:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Pendulum-6408a0365cda59c57d39bee5/issues/pendulum-chain/spacewalk/548?planning-poker) with Zenhub @b-yap @bogdanS98 @ebma @gianfra-t @TorstenStueber

gianfra-t:
@ebma all the listed fields show no data right now? For instance, even the redeem count for a liquidated vault.

PR #559, opened 2024-10-16
Support liquidated vaults in metrics

gianfra-t:
Issue: #548.

### Description of the issue.

Right now, when we start the client pointing to a liquidated vault, the vault will never be saved into the `VaultIdManager` created on initiation (see [here](https://github.com/pendulum-chain/spacewalk/blob/main/clients/vault/src/system.rs#L123)).

The issue is that many of the metrics code relies on iterating this map to obtain the metrics from on-chain data (eg.: [here](https://github.com/pendulum-chain/spacewalk/blob/main/clients/vault/src/metrics.rs#L418) or [here](https://github.com/pendulum-chain/spacewalk/blob/main/clients/vault/src/metrics.rs#L496)).

### Proposed workaround

Add a new field to  `VaultData` to keep track of a liquidated vault, and [add the vault](https://github.com/pendulum-chain/spacewalk/compare/548-make-vault-clients-export-metrics-with-default-values-if-no-data-is-available?expand=1#diff-0f76594421d4493b3bba1c9ed7864cdab7148f19c5d3a18a9bcc9a174bb85aa8R149) even if the status is `Liquidated` when the systems starts.

Modifying the `get` methods [like](https://github.com/pendulum-chain/spacewalk/compare/548-make-vault-clients-export-metrics-with-default-values-if-no-data-is-available?expand=1#diff-0f76594421d4493b3bba1c9ed7864cdab7148f19c5d3a18a9bcc9a174bb85aa8R175) `get_entries` to filter by default non-liquidated vaults allows to maintain compatibility, while in metrics we fetch all vaults on the `VaultIdManager`.

### Testing

To verify that the missing exports, I ran a new vault against the `testnet`, but momentarily patching the code such that it was not added to `VaultIdManager`. When exploring the metrics, I could see that most of them are missing. 

### Extra metric

Added a `Liquidated` metric that displays if the vault is liquidated.

```
# HELP liquidated Boolean reporting if the vault is currently liquidated
# TYPE liquidated gauge
liquidated{currency="XCM(0)_Stellar(USDC:GAKNDFRRWA3RPWNLTI3G4EBSD3RGNZZOY5WKWYMQ6CQTG3KIEKPYWAYC)"} 0
```

This takes the `liquidated` status from `VaultData`, this is currently NOT updated and is only updated upon restart. I believe the vault will not be set back to regular status since the `recover` extrinsic does not emit. Also, we are not polling to check the status of it from the state which means that a liquidated vault will stay in this state (in the client) until restart.

Furthermore, we will not detect if the vault is liquidated while the client is running. Although it was observed that this leads to a restart of the client in production.


clients/vault/src/system.rs
```diff
@@ -150,11 +156,32 @@ impl VaultIdManager {
...
+			return None;
+		}
+		return Some(vault);
+	}
+
 	pub async fn get_vault(&self, vault_id: &VaultId) -> Option<VaultData> {
 		self.vault_data.read().await.get(vault_id).cloned()
 	}
 
+	// Get all ACTIVE vaults
```
gianfra-t, line 172: I didn't want to change the name of this method to `get_active_entries` since it is used in other places and wanted to minimize the modifications to the rest of the code, but we could of course.

clients/vault/src/system.rs
```diff
@@ -150,11 +156,32 @@ impl VaultIdManager {
 		self.vault_data.read().await.get(vault_id).map(|x| x.stellar_wallet.clone())
 	}
 
+	pub async fn get_active_vault(&self, vault_id: &VaultId) -> Option<VaultData> {
```
gianfra-t, line 159: The reason I added this method is to have it into account for the future. Now, `get_vault` will bring the data even if the vault is liquidated. This impacts a few processes, namely:
- [Redeem](https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/vault/src/redeem.rs#L26).
- [Pay and Execute](https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/vault/src/requests/execution.rs#L242)
- [Accept and replace](https://github.com/pendulum-chain/spacewalk/blob/7c7989875e95e1cfe3b0aeba9bb6af01d9e33e58/clients/vault/src/replace.rs#L38)

I think these are all processes that the liquidated vault should be able to perform anyway. This is the most critical difference about this change, simply replacing in those instances with `get_active_vault` will not modify the functionality, in case this assumptions are wrong.

ebma (approved):
I tested this with the liquidated XLM vault on Amplitude. I recalled that we actually still do have a liquidated vault available, namely the [old XLM vault](https://www.notion.so/satoshipay/Spacewalk-Vault-Client-Information-0f9e02b91c54422dbc42a5e4172781ff?pvs=4#20a77cc113bd4fd0b11215edc94f507e). This is because we were not able to call the recover() extrinsic because of some very old non-resolvable redeem request so we created a separate XLM vault instead.

I can confirm that it works as expected. The metrics are properly exported also for the liquidated vault. And the 'liquidated' gauge also works. Great job @gianfra-t 🙏

gianfra-t:
That's great @ebma ! Thanks for testing, I kinda ignore Amplitude since I couldn't connect to polkadot.js and then forgot to check there. I'll merge this then!

PR #569, opened 2024-11-25
Upgrade to Polkadot v1.6.0

### LINKED ISSUE #567: Update to Polkadot-sdk 1.6.0
Part of master ticket [tasks/#419](https://github.com/pendulum-chain/tasks/issues/419).

To perform the upgrade of Pendulum to Polkadot-sdk release `1.6.0` we need to update all the Spacewalk dependencies to [polkadot-sdk release v1.6.0](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.6.0).

### Note on extra dependencies
`subxt` and `jsonrpc` are two critical dependencies that we may want (or need) to update, depending on the common dependencies used also with `polkadot-sdk`.

`polkadot-sdk` released Jan 16, an idea would be to update `subxt` to `v0.34.0` which [released](https://github.com/paritytech/subxt/releases/tag/v0.34.0) right after that. Consequently, `jsonrpc` could be bumped to `0.21` which is what `subxt` [uses](https://github.com/paritytech/subxt/blob/v0.34.0/Cargo.toml#L74C1-L74C10) for that version. This may not be needed since we are already at `v0.33.0` with `subxt` but is good to keep in mind, since hard-to-debug compilation errors may arise from these version definitions (see what happened [here](https://github.com/pendulum-chain/spacewalk/pull/536) as described on PR description).

There are other `jsonrpc` imports which use a [previous version](https://github.com/pendulum-chain/spacewalk/blob/0407091aa1c15a1c8a946bf1cd7007e24789978b/testchain/node/Cargo.toml#L76) and may require more refactor to upgrade, with low or no benefit.

gianfra-t:
Closes #567.

### Main changes
- Bump all `polkadot-sdk` dependencies to `1.6.0`.
- Move dependencies to workspace.
- Update `chain_spec` due to deprecated `ChainSpec::from_genesis(.....)` method.
- Update node.


clients/runner/Cargo.toml
```diff
@@ -3,29 +3,28 @@ name = "runner"
...
-subxt = { version = "0.33.0", default-features = false, features = ["jsonrpsee", "native"]}
-sha2 = "0.8.2"
+url = { workspace = true }
+nix = { workspace = true }
+async-trait = { workspace = true }
+bytes = { workspace = true }
+signal-hook = { workspace = true }
+signal-hook-tokio = { workspace = true , features = ["futures-v0_3"] }
+futures = { workspace = true }
+subxt = {version = "0.33.0", default-features = false, features = ["jsonrpsee", "native"]}
```
ebma, line 24: Is it intentional that it's not using a workspace version?

clients/runtime/Cargo.toml
```diff
@@ -22,33 +22,32 @@ testing-utils = [
...
+sp-std = { workspace = true }
+sp-core = { workspace = true }
+sp-version = { workspace = true }
 
-# Subxt dependencies
-subxt = { version = "0.33.0", features = [ "substrate-compat"] }
-jsonrpsee = { version = "0.20.0", features = ["macros", "jsonrpsee-types", "client", "jsonrpsee-ws-client", "jsonrpsee-client-transport"] }
-# Used for performing the integration tests
+
+subxt = { version = "0.33.0", features = ["substrate-compat"] }
```
ebma, line 49: Same here.

clients/vault/Cargo.toml
```diff
@@ -23,60 +23,59 @@ allow-debugger = ["console-subscriber"]
...
-tracing = { version = "0.1", features = ["log"] }
+tracing = { workspace = true, features = ["log"] }
 
-jsonrpc-core-client = { version = "18.0.0", features = ["http", "tls"] }
+jsonrpc-core-client = { workspace = true, features = ["http", "tls"] }
 
 subxt = { version = "0.33.0", features = [ "substrate-compat"] }
 
+# Substrate dependencies
+# was with default-features = true??
```
ebma, line 57: This shouldn't matter as we are in the client here. The client does not need to avoid `default-features` or `std` for that matter.

pallets/currency/Cargo.toml
```diff
@@ -7,29 +7,29 @@ name = "currency"
...
+frame-system = { workspace = true }
 
 # Parachain dependencies
 primitives = { package = "spacewalk-primitives", path = "../../primitives", default-features = false }
 
 # Orml dependencies
-orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v1.1.0", default-features = false }
-orml-currencies = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v1.1.0", default-features = false }
-orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v1.1.0", default-features = false }
+orml-tokens = { workspace = true, default-features = true }
```
ebma, line 21: You are assigning `default-features = true` everywhere, is this intentional or do you want to do `default-features = false`?

pallets/oracle/Cargo.toml
```diff
@@ -5,46 +5,45 @@ authors = ["Pendulum"]
...
 [dependencies]
-codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive", "max-encoded-len"] }
-scale-info = { version = "2.0.0", default-features = false, features = ["derive"] }
+codec = { workspace = true, features = ["derive", "max-encoded-len"] }
+scale-info = {  workspace = true, features = ["derive"] }
 
 # Substrate dependencies
-sp-std = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.1.0", default-features = false }
-sp-arithmetic = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.1.0", default-features = false, optional = true }
+sp-std = { workspace = true }
```
ebma, line 12: Shouldn't we put `default-features = false` again here?

ebma (commented):
I think for some dependencies there is now a mismatch of how the default-features were configured.

clients/runner/Cargo.toml
gianfra-t: Intentional for now, was complaining if moved to workspace and I wanted to finish updating first before checking out why.

clients/vault/Cargo.toml
gianfra-t: Good find, it's actually wrong now.

pallets/currency/Cargo.toml
gianfra-t: Intentional, assigning `default_features = false` on the module but not on the workspace is overridden.

pallets/oracle/Cargo.toml
gianfra-t: Connected to this [one](https://github.com/pendulum-chain/spacewalk/pull/569#discussion_r1858226124), so basically since that is overridden by the workspace definition, we go the other way around, set `default_features = false` on the workspace, and then if needed, we override with `default_features = true`, which works.

pallets/currency/Cargo.toml
ebma: But in this case here, you are using the `default-features = true` on a dependency and not a dev-dependency which seems wrong to me, considering that it will be included in the wasm runtime like this?

gianfra-t: Oh!! In this case is absolutely wrong, of course. The general case holds but this is doing the opposite.

clients/vault/resources/config/mainnet/stellar_relay_config_frankfurt.json
```diff
@@ -4,18 +4,13 @@
...
     "is_pub_net": true
   },
   "stellar_history_archive_urls": [
-    "http://history.stellar.org/prd/core-live/core_live_001",
-    "http://history.stellar.org/prd/core-live/core_live_002",
-    "http://history.stellar.org/prd/core-live/core_live_003",
-    "https://stellar-history-de-fra.satoshipay.io",
-    "https://stellar-history-sg-sin.satoshipay.io",
-    "https://stellar-history-us-iowa.satoshipay.io"
+    "https://hercules-history.publicnode.org"
```
ebma, line 14: Did you change this due to the 'overloaded' issue or for a different reason? If so, you'd rather need to change the `address` field in `connection_info` to the `host` you can also find in Stellarbeat. If not, I'll leave this comment as a reminder that we should change it back eventually.

clients/wallet/src/horizon/responses.rs
```diff
@@ -202,8 +202,6 @@ pub struct TransactionResponse {
 	#[serde(deserialize_with = "de_str_to_bytes")]
 	pub result_xdr: Vec<u8>,
 	#[serde(deserialize_with = "de_str_to_bytes")]
-	pub result_meta_xdr: Vec<u8>,
```
gianfra-t, line 205: Deprecated field (see [here](https://developers.stellar.org/docs/data/horizon/api-reference/resources/transactions/object)).

primitives/src/lib.rs
```diff
@@ -529,10 +529,15 @@ impl Asset {
 #[allow(clippy::unnecessary_cast)]
 pub enum CurrencyId {
 	#[default]
+	#[serde(rename = "Native")]
```
gianfra-t, line 532: This changes may be significant. The particular reason I added these (and removed the `camelCase` renaming above, is now due to the following panic:
```
Client(Storage("Invalid JSON blob: unknown variant `xCM`, expected one of `Native`, `XCM`, `Stellar`, `ZenlinkLPToken`, `Token` at line 1 column 2105"))
```

This comes from the changes in how the chain spec is defined now (first serialized in a json format).
I believe this error was also an issue for some extrinsics.

testchain/node/Cargo.toml
```diff
@@ -12,17 +12,17 @@ targets = ["x86_64-unknown-linux-gnu"]
...
 [dependencies]
-clap = { version = "4.0.26", features = ["derive"] }
-codec = { package = "parity-scale-codec", version = "3.0.0" }
-hex-literal = "0.2.1"
-log = "0.4.8"
-serde = { version = "1.0.130", features = ["derive"] }
-serde_json = "1.0.85"
-futures = "0.3.15"
-tokio = { version = "1.37", features = ["full", "tracing"] }
+clap = { version = "4.5.3", features = [ "derive" ] }
```
gianfra-t, line 18: Node requires a higher version of `clap` compared to that used in the workspace.

clients/vault/resources/config/mainnet/stellar_relay_config_frankfurt.json
gianfra-t: Back to original configs now.

gianfra-t:
@pendulum-chain/devs the CI may get stuck on the integration tests some times (with integration tests never ending, unknown reason. [Example](https://github.com/pendulum-chain/spacewalk/actions/runs/12179775520/job/33976388803)), yet locally I was able to run successfully all of them. 

I believe this does not have to do with the update itself.

primitives/src/lib.rs
ebma: I also remember that error you are referring to. I think changing it the way you did makes sense. 

This could break some polkadot.js types as at least the XCM type seems to be sometimes referred to as `Xcm` see [here](https://github.com/pendulum-chain/pendulum.js/blob/46ca590c43ee74e044b9eab209379ed4b7570c75/packages/types/src/interfaces/vaultRegistry/types.ts#L35). Elsewhere it's used as `XCM` though, see [here](https://github.com/pendulum-chain/pendulum.js/blob/46ca590c43ee74e044b9eab209379ed4b7570c75/packages/types/src/lookup/pendulum.ts#L1699),  so maybe it's not a big deal. We'll see but can easily fix the problem by upgrading the types in pendulum.js.

testchain/node/Cargo.toml
```diff
@@ -12,17 +12,17 @@ targets = ["x86_64-unknown-linux-gnu"]
...
+codec = { workspace = true, default-features = true, version = "3.0.0" }
```
ebma, line 19: How come we define workspace = true but still define a custom version?

zepter.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - ['format', 'features', '--offline', '--locked', '--quiet']
+  # Same as `check`, but with the `--fix` flag.
+  default:
+    - [ $check.0, '--fix' ]
+    - [ $check.1, '--fix' ]
+
+# Will be displayed when any workflow fails:
+help:
+  text: |
+    Polkadot-SDK uses the Zepter CLI to detect abnormalities in the feature configuration.
```
ebma, line 38: Nit: we could change this message. Doesn't matter much.

pallets/issue/rpc/Cargo.toml
```diff
@@ -5,9 +5,9 @@ authors = ["Interlay <contact@interlay.io>"]
 edition = "2021"
 
 [dependencies]
-codec = { package = "parity-scale-codec", version = "3.0.0" }
+codec = { workspace = true, default-features = true }
 jsonrpsee = { version = "0.16.0", features = ["client", "server", "macros"] }
```
ebma, line 9: Do we need this custom version or would it also work with the one in workspace?

testchain/node/Cargo.toml
```diff
@@ -31,45 +31,46 @@ spacewalk-runtime-testnet = { package = "spacewalk-runtime-standalone-testnet",
...
+frame-system = { workspace = true, default-features = true }
 
-pallet-transaction-payment = { default-features = false, git = "https://github.com/paritytech/polkadot-sdk.git", branch = "release-polkadot-v1.1.0" }
+pallet-transaction-payment = {  workspace = true, default-features = true }
 
-substrate-stellar-sdk = { git = "https://github.com/pendulum-chain/substrate-stellar-sdk", branch = "polkadot-v1.1.0" }
+substrate-stellar-sdk = { workspace = true, default-features = true }
 
 # RPC dependencies
 jsonrpc-core = "18.0.0"
```
ebma, line 76: Would the workspace dependencies not work?

ebma (approved):
Looks good to me. I left only very minor comments.

primitives/src/lib.rs
gianfra-t: I see. Why did the pendulum types worked as `Xcm` and not `xCM` is a mystery then. But this serialization is only relevant when compiled with `std`. In any case, let's hope is not a big deal.

testchain/node/Cargo.toml
gianfra-t: Let me try. It could be that I or zepter missed them for some reason.

testchain/node/Cargo.toml
gianfra-t: Was not correct. Seems that it was ignored anyway.

pallets/issue/rpc/Cargo.toml
gianfra-t: We probably don't "need" that version, but upgrading it involves making changes to each module's rpc code that we didn't have much time to go through.

PR #570, opened 2024-12-03
[TEST- DO NOT MERGE] Baseline 1.6.0 tests

gianfra-t:
Objective of this PR is to trigger the CI, and have a baseline for which we can isolate the update-only issues that lead to failed tests on 1.6.0 PR (#569 )

### DIFF
--- a/clients/wallet/src/mock.rs
+++ b/clients/wallet/src/mock.rs
@@ -124,7 +124,7 @@ pub fn wallet_with_secret_key_for_storage(
 	storage: &str,
 	secret_key: &str,
 ) -> Result<Arc<RwLock<StellarWallet>>, Error> {
-	Ok(Arc::new(RwLock::new(StellagitrWallet::from_secret_encoded_with_cache(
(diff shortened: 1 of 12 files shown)


clients/wallet/src/mock.rs
```diff
@@ -124,7 +124,7 @@ pub fn wallet_with_secret_key_for_storage(
 	storage: &str,
 	secret_key: &str,
 ) -> Result<Arc<RwLock<StellarWallet>>, Error> {
-	Ok(Arc::new(RwLock::new(StellarWallet::from_secret_encoded_with_cache(
+	Ok(Arc::new(RwLock::new(StellagitrWallet::from_secret_encoded_with_cache(
```
ebma, line 127: This might be problematic 🤡

ebma (commented):
I restarted the test for the vault client. It failed before reaching the integration tests so we can't know if they work. Or did you try it locally as well @gianfra-t?

clients/wallet/src/mock.rs
gianfra-t: Upss. Thanks. My IDE was incredibly laggy with 4 projects open yesterday.

PR #573, opened 2024-12-23
Update to latest rev of substrate-stellar-sdk

ebma:
This is required because the SCP types changed again and the vault is otherwise not able to decode archived entries for transaction sets. This was fixed in https://github.com/pendulum-chain/substrate-stellar-sdk/pull/38 and is available on the latest version of the `polkadot-v1.1.0` branch.


ebma:
I'll merge this now as the compilation seems fine and most tests pass (expect for the usual suspects).

